PR #211, opened 2023-02-02
Exporting categorical

Joaoloula:
Exporting the scipy categorical distribution


femtomc:
@Joaoloula I'll merge this -- but first I'm going to add some tests that test `TFPCategorical` against `Categorical` to ensure the implementations align across the backends.

PR #214, opened 2023-02-02
fix map builtin use

Joaoloula:
map builtin is used in map_combinator.py, which also defines a "map"

making the use of the builtin explicit here, but not sure if it'd be best to avoid redefining "map" altogether


femtomc:
@Joaoloula I've removed the shorthand.

PR #429, opened 2023-07-24
Fix intro notebook links

sritchie:
This PR fixes references to the genjax intro notebook


femtomc (approved):
Thank you for spotting and fixing!

PR #454, opened 2023-08-03
The great key migration.

femtomc:


femtomc:
@derifatives / @nishadgothoskar / @mirkoklukas this is a massive breaking PR to change the usage of keys to conform with JAX's recommended idioms.

Now, instead of having the key piped through:

```python
key, tr = gen_fn.simulate(key, args)
```

The key is not returned:

```python
tr = gen_fn.simulate(sub_key, args)
```

PR #757, opened 2023-12-11
port all unit tests from Static to Interpreted language

littleredcomputer:
Some current differences in Interpreted vs. Static languages revealed by porting one of the unit tests:

In calling `trace`, arguments to function are curried in Static, in-line in Interpreted:

```Python
    y1 = genjax.trace("y1", genjax.normal)(0.0, 1.0)   # vs.
    y1 = trace("y1", genjax.normal, 0.0, 1.0)
```

Importance wants a ChoiceValue, but in interpreted `com.get_submap` returns a `DistributionTrace`. It works if I change it as follows:

```Python
    (_, score1) = genjax.normal.importance(key, chm.get_submap("y1"), (0.0, 1.0))    
    (_, score1) = genjax.normal.importance(key, chm["y1"].get_choices(), (0.0, 1.0))
```

Let me know if you would like me to bring the Interpreted side into compliance here or if these differences are WAI.


femtomc:
Ah, those two pieces of behavior are not WAI, and we should make them consistent.

But it's not too hard to make them consistent! Is the choice map for `InterpretedTrace.get_choices()` a `StaticChoiceMap`?

littleredcomputer:
It's a `HierarchicalChoiceMap`.

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -444,6 +444,10 @@ def __getitem__(self, addrs: Tuple):
     def __getitem__(self, addr: Any):
         return self.__getitem__((addr,))
 
+    @abc.abstractmethod
+    def __setitem__(self, key, value):
+        pass
+
```
femtomc: `ChoiceMap` instances are treated as immutable. Only special choice maps supports `__setitem__`, we shouldn't enforce that they should all have this method.

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -72,7 +72,7 @@ def get_score(self):
     def get_choices(self):
         return ChoiceValue(self.value)
 
-    def project(self, selection: Selection) -> FloatArray:
+    def project(self, selection: Selection) -> ArrayLike:
         if isinstance(selection, AllSelection):
```
femtomc, line 76: What is the type difference between `ArrayLike` and `FloatArray`? And why switch these signatures?

src/genjax/_src/generative_functions/interpreted/__init__.py
```diff
@@ -155,25 +164,15 @@ def process_message(self, msg):
         return retval
 
 
-@dataclass
+@beartype
```
femtomc, line 162: The `@dataclass` decorator automatically defines an `__init__` -- and my personal opinion is that it makes class declarations easier to read.

In general, I'm encouraging usage of `@dataclass` for `Pytree` inheritors -- as pretty printing also handles `@dataclass` declared classes in a convenient way.

femtomc: Instead of changing how we declare classes, why don't we determine if the equality check in `self in ...` for the handlers can be changed. That pattern is quite simple, and I think we likely have a lot of flexibility with it.

src/genjax/core/exceptions.py
femtomc, line 1: Big fan of this pattern, thanks for starting it.

notebooks/concepts/introduction/intro_interpreted.ipynb
femtomc, line 1: Let's iterate on this notebook before making it part of the public notebook repository -- but for now, this is great!

femtomc (approved):

src/genjax/_src/generative_functions/distributions/distribution.py
littleredcomputer: According to JAX documentation, ArrayLike means anything that can implicitly serve as a JAX tensor, which includes bare floats. Many of the default implementations say things like "return 0.0", which is a type violation under FloatArray but checks OK under ArrayLike.  See: https://jax.readthedocs.io/en/latest/jax.typing.html#jax-typing-best-practices

src/genjax/_src/generative_functions/interpreted/__init__.py
littleredcomputer: It can be, we can just write a for loop that checks if `self is stack[i]`. I don't disagree about PyTree instances, but the handler objects that get pushed on the stack? I don't think they need a custom implementation of __eq__, and these types will never be seen by the user

notebooks/concepts/introduction/intro_interpreted.ipynb
littleredcomputer: Agreed. I checked it into the branch only because that's the easiest way to "publish the results" w/o asking people to fire up Jupyter

src/genjax/_src/core/datatypes/generative.py
littleredcomputer: OK. I must have put this here because someone said `map[k] = v` somewhere and it wasn't type-checking. I can go back and find out where. I'm glad to hear that such maps are immutable, but we have to be honest about type IMO and not just mutate one because we can

src/genjax/_src/generative_functions/interpreted/__init__.py
littleredcomputer: Another way we could do it is with `@dataclass(eq=False)`. I did this, and you can even specify defaults for the fields so you don't have to write your own `new` methods, but then you need to specify factories for the mutable initializers. If we aren't going to use the `__init__` method provided by `@dataclass`, then I think we're paying too much for them. Since these are throwaway objects just meant to bind parameters for a future function call, I don't think it's painful to use a plain ol' Python object here

littleredcomputer: just for the heck of it I'm going to push an update with the `@dataclass` idea for you to look at

src/genjax/_src/core/datatypes/generative.py
littleredcomputer: I took this out, but we will need to reconsider some things to get the typing right. Is it better, I wonder, for choice maps to throw from `__setitem__` explaining the immutability, or not to have the method?

femtomc: It's better to not have the method I think.

If there's code somewhere _in interface implementations_ that has `map[k] = v`, we should correctly type the type of `map`. I'm pretty confident that the only place this could occur is in the static language currently.

You don't need to handle that in this PR.

src/genjax/_src/generative_functions/distributions/distribution.py
femtomc: Great, I've been looking for something like this. Thanks for finding it.

src/genjax/_src/generative_functions/interpreted/__init__.py
femtomc: I agree -- for the handlers, they don't need to be `PyTree`. I think perhaps the convention (as you mentioned), should be: if the class is `genjax.Pytree`, make it a `dataclass` too. But otherwise, no convention.

femtomc: I'll check out your code nonetheless!

femtomc: This looks good! 

I like the usage of `default_factory` as well. Separately, we should discuss the `Class.new` convention I have in the codebase. I don't think it applies here (you've done the right thing obviously, because we don't expect users to construct the handlers by themselves).

femtomc (approved):

PR #797, opened 2023-12-31
ChoiceMap type unification

littleredcomputer:
The types `Choice` and `EmptyChoice` are deleted. The root of the type hierarchy is now `ChoiceMap`, which can be instantiated to produce an empty choice object.  Choice objects support the `len()` operation.  ChoiceValue is now a subtype of ChoiceMap.


femtomc:
The unification described in this PR is implemented in #1072.

PR #803, opened 2024-01-04
remove LanguageConstructor and lang(). Decorators act directly

littleredcomputer:
Mandatory keyword arguments are enforced.


femtomc (approved):
Thank you for painstakingly undertaking this.

PR #820, opened 2024-01-05
Replace `.new` methods with `@dataclass` init methods.

littleredcomputer:
Many shortcuts have been deleted, as they weren't used and were kind of scattered. If we still want them, we might consider making an easy way to import them wholesale like `from genjax.shortcuts import *` for use with Jupyter etc.


src/genjax/_src/generative_functions/combinators/vector/vector_datatypes.py
```diff
@@ -119,6 +98,16 @@ class IndexedChoiceMap(ChoiceMap):
...
+    # TODO(colin): Ask McCoy: Maybe a way to unify all these interfaces would
+    # be to take a dict argument with integer keys, as in:
+    #
+    # IndexedChoiceMap.from_dict({
+    #   1: x,
+    #   2: y
+    # })
+    # The indices and values would then look "zipped together" in the source code
+    # pyright-wise, we could enforce the int type constraint on the dict keys
+
```
femtomc, line 110: I like this idea.

src/genjax/_src/generative_functions/combinators/vector/vector_datatypes.py
```diff
@@ -253,6 +242,9 @@ class VectorChoiceMap(ChoiceMap):
     def flatten(self):
         return (self.inner,), ()
 
+    # TODO(colin): Ask McCoy: I'm uncomfortable having the constructor for one
+    # class return an object of another type, so maybe we want another
+    # helper function here. Can we enforce that the type of inner is an array?
```
femtomc, line 247: I think we should just have a `vector_choice_map` which does the check and returns if the thing we're trying to pass is an `EmptyChoice`.

PR #826, opened 2024-01-08
The remainder of the `new` rewrites.

littleredcomputer:
- the only one left is DataSharedSumTree.new; that constructor name could be changed, but it's not apparent what a better name would be.

- There are some news in the runtime debugger, but that's not shipping with the next version so I left it alone

I created a new module, genjax.shortcuts, to hold things like choice_map(). Doing so allows a single function to create any kind of choice map without circular import problems and allows a single point of documentation for the shortcut policy.

The SMC things had an unusual parameter order dicatated by the JAX/static requirement, so I made n_particles a required keyword parameter of the shortcut to keep things looking nice.

### DIFF
--- a/notebooks/intermediate/smc/smc_with_translators.ipynb
+++ b/notebooks/intermediate/smc/smc_with_translators.ipynb
@@ -80,7 +80,7 @@
     "    index_sel = genjax.indexed_select(0)\n",
     "    obs_init = obs.slice(0)\n",
     "    key, sub_key = jax.random.split(key)\n",
-    "    smc_state = smc.smc_initialize(chain, 5).apply(sub_key, obs_init, (0, init_state))\n",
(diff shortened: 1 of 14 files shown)


femtomc:
@littleredcomputer `DataSharedSumTree` is entirely internal, and I don't feel strongly about enforcing our philosophy of `.new` there

femtomc (approved):
@littleredcomputer This looks good!

Note that, instead of using `typing` -- you should look at `genjax._src.core.typing` which uses `beartype` as `btyping`.

I've also attempted to put the types that we use in the code base in that one file, and just control the types from that module.

littleredcomputer:
`genjax._src.core.typing` which uses `beartype` as `btyping`.

Oops: I'll fix that first. Thanks!

PR #836, opened 2024-01-08
Build and deploy docs via GitHub Action

sritchie:
This PR

- adds a GitHub Action that will trigger docs generation and publication on any push / merge to the `nightly` branch
- moves the DEVELOPING.md guide into `docs` where a duplicate previously lived

Once we get multiple versions of the docs going, we can change the rebuild trigger to include `main` as well.


.github/workflows/ci.yml
```diff
@@ -10,14 +10,11 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
-env:
```
sritchie, line 13: this isn't currently used, and I think we can do a better job here, copying the new style that Emmy uses and that codecov prefers...

.github/workflows/ci.yml
```diff
@@ -10,14 +10,11 @@ on:
...
-  CODECOV_TOKEN: 1abb2fd7-0cb5-433c-b2cc-1596466aeb56
-
 jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
```
sritchie, line 20: gotta be up to date!!

littleredcomputer (approved):
this is really instructive, thank you

PR #847, opened 2024-01-11
Run Jupyter notebooks in CI

sritchie:
This PR:

- adjusts our `ci.yml` job to run our tests and our notebooks in parallel
- updates all notebooks to compatibility with `nightly` so they build and pass tests
- deletes failing notebooks, so we can add them via a new PR

PR #850, opened 2024-01-11
Re-add broken notebooks

sritchie:
These notebooks were removed for #847 ; this PR stages them back, so we can fix the test errors and get them re-integrated.


femtomc:
Killed by #1072.

PR #854, opened 2024-01-11
Run notebooks on docs deploy, strip cell output in pre-commit

sritchie:
This PR:

- Strips all cell outputs from the git-committed version of the notebooks
- re-generates all outputs in the `docs-build` task


littleredcomputer (approved):
AMAZING Sam

PR #856, opened 2024-01-11
Test docs in CI, all docs in strict mode

sritchie:
This PR checks that the docs (via mkdocs) don't contain any errors by running in strict mode.

The new action only runs on PRs, since the docs build happens anyway

### DIFF
--- a/.github/workflows/test_docs.yml
+++ b/.github/workflows/test_docs.yml
@@ -1,4 +1,4 @@
-name: Test Docs
+name: Test mkdocs build
 
 on:
   pull_request:
@@ -26,9 +26,6 @@ jobs:
         with:
           ssh-private-key: ${{ secrets.INSIDER_SSH_KEY }}
 
-      - name: Set up Quarto
-        uses: quarto-dev/quarto-actions/setup@v2
-
       - uses: actions/setup-python@v4
         with:
           python-version: 3.11.5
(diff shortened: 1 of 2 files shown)


noxfile.py
```diff
@@ -161,6 +161,18 @@ def build(session):
     session.run("poetry", "build")
 
 
+@session(name="mkdocs", python=python_version)
```
sritchie, line 164: I don't see how to stack these sessions... ideally the docs-serve task would extend this task, but I think this isn't supported, unfortunately :(

.github/workflows/test_docs.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Setup Nox
+        uses: daisylb/setup-nox@v2.1.0
+
+      - name: Install and configure Poetry
+        uses: snok/install-poetry@v1
+
+      - name: Install nox-poetry
+        run: pip install nox-poetry
+
+      - name: Run Docs Build
```
sritchie, line 42: I broke out the mkdocs task because we already build our notebooks via nbmake, which I think is more appropriate than letting quarto fail on docs build due to its integration with pytest.

Or maybe I'm wrong, and we can consolidate later and let this test take over instead of nbmake? I think the nbmake errors are reported better. But that might be wrong.

noxfile.py
littleredcomputer: I might argue that quarto should be a separate task, since it's so heavy

littleredcomputer (approved):

PR #857, opened 2024-01-12
Split sugar callee methods into `__call__` and `apply`.

littleredcomputer:
One usability issue that remains in GenJAX is that `@dispatch` is used for two implementations of `__call__` in SugaredGenerativeFunctionCall. IDEs will type-check against one or the other, and typically pick the `(key, *args)` form, since it's defined later, and appears to override the `(*args, **kwargs)` form.

Having a `@dispatch` on __call__ thus "redlines" all the users' function invoications.

I therefore propose that we take the less-common use, where you manually supply the key and pass the args as a tuple, and rename it `apply`, by analogy with LISP, where apply is used when you already have your arguments in a list and don't want to splat them out into a function call.

Using the technique described [here](https://stackoverflow.com/questions/51930339/how-do-i-correctly-add-type-hints-to-mixin-classes), added type information for `SupportsCalleeSugar` to describe the constrains it imposes on the class into which it is mixed

### DIFF
--- a/src/genjax/_src/generative_functions/supports_callees.py
+++ b/src/genjax/_src/generative_functions/supports_callees.py
@@ -14,16 +14,15 @@
 
 from dataclasses import dataclass
 
-from genjax._src.core.datatypes.generative import GenerativeFunction
 from genjax._src.core.pytree.pytree import Pytree
-from genjax._src.core.typing import Any, Dict, PRNGKey, Tuple
+from genjax._src.core.typing import Any, Callable, Dict, List, PRNGKey, Protocol, Tuple
 
 
 # This class is used to allow syntactic sugar (e.g. the `@` operator)
 # in languages which support callees for generative functions via a `trace` intrinsic.
 @dataclass
 class SugaredGenerativeFunctionCall(Pytree):
-    gen_fn: GenerativeFunction
+    gen_fn: Callable
     kwargs: Dict
     args: Tuple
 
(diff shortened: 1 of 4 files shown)


src/genjax/_src/generative_functions/supports_callees.py
```diff
@@ -56,15 +55,21 @@ def wrapped(*args):
...
+        self: CanSimulate, *args: Any, **kwargs
+    ) -> SugaredGenerativeFunctionCall:
         return SugaredGenerativeFunctionCall(self, kwargs, args)
 
-    @dispatch
-    def __call__(self, key: PRNGKey, args: Tuple) -> Any:
-        tr = self.simulate(key, args)
-        retval = tr.get_retval()
-        return retval
+    def apply(self: CanSimulate, key: PRNGKey, args: Tuple) -> Any:
```
femtomc, line 74: I think `apply` is a weird name for this capability, and is not descriptive.

I think, to be precise, we should name it `sample_retval` -- because that's what is occurring.

femtomc:
@littleredcomputer I think what we should do for this is the following:

For any generative function, if it supports `simulate`, it also supports a method called `sample_retval` which has the exact implementation of `apply`. This method is automatically implemented in the `GenerativeFunction` base class.

We don't need a `Protocol` to distinguish usage -- or to check `CanSimulate` -- instead, we just have `SupportsCalleeSugar` overload `__call__`.

I wish there was a way to check for overload clashes with `__call__` (e.g. if a user defines their own generative function, accidentally defines a `__call__`, and expects it to work in the static/interpreted language).

littleredcomputer:
The protocol is only for the type checker, it has no runtime component...we need a type for `self` that promises that `.simulate` is one of its methods. The only other way to do that is to have GenerativeFunction be a base class of SupportsCalleeSugar, which is not what we want. What we do want is a type constraint on what kinds of class SupportsCalleeSugar can be mixed _into_, and that's what I've attempted to provide

PR #865, opened 2024-01-12
Prepare for 0.1.0 release

sritchie:
This PR:

- removes the duplicate DEVELOPING.md (the real one is in `docs`)
- Modifies dependabot to only apply patches against `nightly`
- Deletes the unused codecov.yml for tidying purposes (we'll add this back when we go public)
- Adds a nascent CHANGELOG.md


littleredcomputer (approved):
Perfect. I didn't know the Sunday version upgrade thing was that easy.

PR #875, opened 2024-01-16
link interpreted Jupyter demo to doc tree. Some interpreted code golf.

littleredcomputer:
We also relax type constraint on TFPDistribution. Formerly, `TFPDistribution` had both `ExactDensity` and `JAXGenerativeFunction` as type constraints. But:
- `ExactDensity` is a already subclass of `GenerativeFunction`, creating a diamond inheritance
- nothing in the implementation of `TFPDistribution` is JAX-specific and furthermore the interpreted dialect shares these implementations of probability distributions

### DIFF
--- a/notebooks/_quarto.yml
+++ b/notebooks/_quarto.yml
@@ -38,6 +38,9 @@ website:
           - href: concepts/introduction/intro_to_genjax.ipynb
             text: Introduction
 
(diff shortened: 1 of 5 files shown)


sritchie:
LGTM!

femtomc:
I don’t know why the tests on this went through without the JAX compatibility methods for JAXGenerativeFunction.

PR #876, opened 2024-01-17
bug hunt branch

littleredcomputer:
- strip down readme 
- add a unit test with one successful and one unsuccessful test
- turn off other tests that would fail in this case

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,151 +9,47 @@
 </div>
 <br>
 
-<div align="center">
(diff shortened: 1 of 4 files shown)


sritchie (commented):
I think this is solid. Let's make a separate "python-bughunt" project, private, in the probcomp org before we actually do this, and commit this branch with erased git history.

sritchie:
Closing now that we have a proper repo.

PR #879, opened 2024-01-18
Restore __abstract_call__ to distributions

littleredcomputer:
PR #875 removed diamond inheritance from ExactDensities but it turns out that to play nice with JAX the `__abstract_call__` method is needed. It has no JAX content of its own, so we move it up to GenerativeFunction, which shares the (innocuous) method with the interpreted dialect.

Or maybe it's necessary for the `ExactDensity` implementors to be `JAXGenerativeFunctions` and not merely `GenerativeFunctions`. I think it's cleaner if we don't inherit `GenerativeFunction` in more than one way, but there may be a constraint here that I can't see

### DIFF
--- a/src/genjax/_src/core/datatypes/generative.py
+++ b/src/genjax/_src/core/datatypes/generative.py
@@ -1197,6 +1197,13 @@ class GenerativeFunction(Pytree):
     ) -> Trace:
         raise NotImplementedError
 
+    def __abstract_call__(self, *args) -> Any:
+        """Used to support JAX tracing, although this default implementation
+        involves no JAX operations (it takes a fixed-key sample from the
+        return value). Generative functions may customize this to improve
+        compilation time."""
+        return self.simulate(jax.random.PRNGKey(0), args).get_retval()
+
 
 @dataclass
 class JAXGenerativeFunction(GenerativeFunction, Pytree):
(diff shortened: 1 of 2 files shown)


femtomc:
You should restore ExactDensity and JAXGenerativeFunction to their previous state.

Why is there a worry about diamond inheritance? We’re using base classes as mixins for methods. Is there an IDE reason for abolishing this pattern?

littleredcomputer:
I can do that but it is interesting that the tests pass.

I should have stated my thought more clearly: how can we be sure that the
distributions will stay interpreted-friendly if they are subclasses of
JAXGenerativeFunction?

femtomc:
@littleredcomputer The interpreted language should place no restrictions on what type of generative function its programs invoke -- in other words, it shouldn't care if its callees are JAX compatible, or not.

On the other hand, its programs cannot be invoked as `JAXGenerativeFunction` -- and that's a big part of why there's a separation between the abstract class `GenerativeFunction` and the class `JAXGenerativeFunction`.

femtomc:
E.g. you'll notice that the generative functions which are passed as parameters to combinator constructors are typed as `JAXGenerativeFunction` -- which is performing this explicit check.

I'm open to a refactoring, as I do feel it's a bit heavy handed to have this type separation when really we're trying to communicate "not interpreted" with the type -- but it was the easiest thing I could think of.

littleredcomputer:
I got on this tack by seeing that people who write their own densities can't just extend `ExactDensity`, since they must mix in JAXGenerativeFunction too. 

```
@dataclass
class UniformPose(ExactDensity, JAXGenerativeFunction):
    def sample(self, key, low, high, **kwargs):
        position = jax.random.uniform(key, shape=(3,)) * (high - low) + low
```

But that's not a problem. I'll put it back the way it was.

femtomc:
Right — I’d prefer to be explicit about the JAX compatibility assumption — versus mixing it in implicitly in the core data types.

femtomc (approved):
LGTM!

PR #881, opened 2024-01-20
Implement Pytree as equinox.Module.

femtomc:


femtomc:
For this one, high-level is that `genjax.Pytree` is now `genjax.Pytree(equinox.Module)` — and every single dataclass which inherits is now frozen, which means that all our structures can only be functionally updated — and I’ve added methods for these where required.

In addition, we can’t use `functools.update_wrapper` anymore because of the frozen property, so we use `equinox.module_update_wrapper` which gives the same functionality.

I did a bit of gardening on the documentation — and gardening on module placement.

Finally, as part of the renaming PR to lower case — I’ve applied the map `static -> static_gen_fn` and `interpreted -> interpreted_gen_fn` to make the naming cohere with our other names, and updated the README, documentation, notebooks as required.

notebooks/introduction/intro_interpreted.ipynb
```diff
@@ -586,7 +586,7 @@
    "source": [
     "observations = genjax.choice_map()\n",
     "for i, y in enumerate(ys):\n",
-    "    observations[\"ys\", \"y\", i, \"value\"] = y\n",
+    "    observations = observations.insert((\"ys\", \"y\", i, \"value\"), y)\n",
```
littleredcomputer, line 589: note for myself: with the functional changes this might be better done via `from_dict` ... another possibility is a builder pattern

src/genjax/_src/generative_functions/interpreted/interpreted_gen_fn.py
```diff
@@ -204,17 +200,16 @@ def handle(self, gen_fn: GenerativeFunction, args: Tuple, addr: Any):
         (tr, w, rd, d) = gen_fn.update(sub_key, sub_trace, sub_map, argdiffs)
         retval = tr.get_retval()
         self.weight += w
-        self.choice_state[addr] = tr
-        self.discard[addr] = d
+        self.choice_state = self.choice_state.trie_insert(addr, tr)
```
littleredcomputer, line 203: trie_insert doesn't really advertise the functional meaning...I forget if there's a way in Python to mark a return value as not safely discardable (hm, upon googling, there's a [proposal](https://discuss.python.org/t/proposal-typing-no-discard-a-decorator-to-indicate-that-the-return-value-should-not-be-discarded/33138) for that). 

A type-safe way of doing this would be to overload `+` to add a `(k, v)` tuple: "trie + (k, v)" would make it clear that the return value must be captured

src/genjax/_src/shortcuts.py
```diff
@@ -0,0 +1,134 @@
...
+
+def choice_map(*vs: ChoiceMappable) -> ChoiceMap:
+    """Shortcut constructor for GenJAX ChoiceMap objects.
+
+    When called with no arguments, returns an empty (mutable) choice map which
+    you can populate using the subscript operator as in
+
+        ```python
+        chm = genjax.choice_map()
+        chm["x"] = 3.0
```
littleredcomputer, line 65: documentation is now out of date?

amdee:
Hi @femtomc and @littleredcomputer — I can't find this feature(Implement Pytree as equinox.Module. #881) on the `main` branch, and the `mrb/eqx` branch appears to have been deleted; was this feature ever implemented, and if so, where can I find it? 
Thank you.

PR #895, opened 2024-01-28
Fix diff bug related to lifting constants into `Diff(..., NoChange)`

femtomc:


littleredcomputer (approved):
as noted in chat: open concern: unit test needed

PR #897, opened 2024-01-29
Tweaks for the next release

sritchie:
This PR:

- reconfigures the GitHub Action to release to our internal artifact registry
- adds instructions on how to publish to and pull from the artifact registry
- resets the hardcoded version to the suggested `0.0.0`  (the real version will be overridden by the poetry plugin)
- changes the package name to lowercase `genjax` as requested by @femtomc

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -152,9 +152,47 @@ Before cutting a new release:
   PRs merged since the last release
 - Click "Publish Release"
 
-This will build and publish the new version to PyPI.
+This will build and publish the new version to Artifact Registry.
+
+### Manually publishing to Google Artifact Registry
+
+- Ask @sritchie to add you to the `probcomp-caliban` project on Google Cloud.
+- [Install the Google Cloud command line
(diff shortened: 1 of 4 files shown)


docs/developing.md
```diff
@@ -152,10 +152,48 @@ Before cutting a new release:
...
+  deploy to the `probcomp-caliban` artifact registry:
+
+```shell
+poetry self add poetry-dynamic-versioning[plugin]
+poetry self add keyrings.google-artifactregistry-auth
+poetry config repositories.gcp https://us-west1-python.pkg.dev/probcomp-caliban/probcomp/
+```
+
+- create a new version tag on the `main` branch of the form
+  `v<MAJOR>.<MINOR>.<INCREMENTAL>`, like `v.0.1.0`, and push the tag to the
```
littleredcomputer, line 177: I think we don't want the dot between v and the initial zero as shown in the example (or else put it in the model to the left)

littleredcomputer (changes requested):
one doc tweak then ship it

PR #919, opened 2024-02-07
fix bugs in mask combinator

georgematheos:
This PR
1. Fixes a few small bugs in MaskingCombinator.importance and MaskingCombinator.update, relating to incorrectly piping arguments for trace construction
2. (Temporarily) removes type checks on MaskingCombinator.importance and MaskingCombinator.update.  These type checks cause problems because in nightly, Mask is not a subtype of Choice.  (This subtyping _is_ implemented in mrb/gensp; once we merge this we should add back in these type checks.  This TODO is noted in a comment in the code.)

### DIFF
--- a/src/genjax/_src/generative_functions/combinators/masking_combinator.py
+++ b/src/genjax/_src/generative_functions/combinators/masking_combinator.py
@@ -77,7 +77,8 @@ class MaskingCombinator(JAXGenerativeFunction, SupportsCalleeSugar):
         tr = self.inner.simulate(key, inner_args)
         return MaskingTrace(self, tr, check)
 
-    @typecheck
+    # TODO: add back in typechecks once `Mask` is made a subtype of `Choice`
+    # @typecheck
     def importance(
         self,
         key: PRNGKey,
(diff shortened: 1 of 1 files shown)


src/genjax/_src/generative_functions/combinators/masking_combinator.py
```diff
@@ -77,19 +77,20 @@ def simulate(
         tr = self.inner.simulate(key, inner_args)
         return MaskingTrace(self, tr, check)
 
-    @typecheck
+    # TODO: add back in typechecks once `Mask` is made a subtype of `Choice`
```
sritchie, line 80: I'm fine with this,unless this is easy to pull out as a PR from @femtomc 's branch that we can merge before this one. wdyt @femtomc ?

littleredcomputer: In the meantime can you make it a Union[Choice, Mask]? (I don't know if that will work and if it causes a cascade of other type errors then forget it)

femtomc:
Optimistically, this is fixed on #1072, closing.

PR #920, opened 2024-02-08
Fix issue where scores/weights could be floats (instead of ArrayLikes).

jburnim:
Without this change, `assess` fails with a type error for a static generative function with no choices (because it returns a float instead of a JAX array).

### DIFF
--- a/src/genjax/_src/generative_functions/interpreted/fn.py
+++ b/src/genjax/_src/generative_functions/interpreted/fn.py
@@ -23,6 +23,7 @@ import itertools
 from dataclasses import dataclass, field
 
 import jax
+import jax.numpy as jnp
 import jaxtyping
 from beartype import beartype
 
@@ -153,7 +154,7 @@ class AddressVisitor:
 @beartype
 class SimulateHandler(Handler):
     key: PRNGKey
-    score: ArrayLike = 0.0
+    score: ArrayLike = field(default_factory=lambda: jnp.zeros(()))
     choice_state: Trie = field(default_factory=Trie)
     trace_visitor: AddressVisitor = field(default_factory=AddressVisitor)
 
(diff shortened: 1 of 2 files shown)


femtomc:
Hi @jburnim thanks for making this! We’ll get to it soon

littleredcomputer:
ArrayLike includes scalars as well as arrays. Could you paste an example of the type error you're seeing? Thanks!

jburnim:
I'll try to put together a minimal example.

The type errors are from calling `JAXGenerativeFunction.assess`, which is annotated to return a score that is a `FloatArray`.

jburnim:
When I run the test `test_assess_with_no_choices ` in https://github.com/probcomp/genjax/compare/jburnim_type_error_repro?expand=1 , I get the error:
```
E   beartype.roar.BeartypeCallHintReturnViolation: Method genjax._src.generative_functions.static.static_gen_fn.StaticGenerativeFunction.assess() return (0.0, Traced<ShapedArray(float32[4])>with<DynamicJaxprTrace(level=1/0)>) violates type hint tuple[jaxtyping.Float[Array, '...'], typing.Any] under non-default configuration BeartypeConf(claw_is_pep526=True, is_color=False, is_debug=False, is_pep484_tower=False, strategy=<BeartypeStrategy.O1: 2>, warning_cls_on_decorator_exception=None), as tuple index 0 item float 0.0 not instance of <protocol "jaxtyping.Float[Array, '...']">.
```

ursk:
Summarizing from chat here:
Replacing `FloatArray` annotations with `ArrayLike` also solves the issue. We seem to be using both throughout the codebase, if it's ok to change it locally like on https://github.com/probcomp/genjax/commit/614ba69e55e08d0e64fc65b73b948630ccd2bc6e does, do you want to merge that solution instead?

femtomc:
It is indeed okay to change locally to `ArrayLike`. In general, I'd like to change the type of `FloatArray` to actually accept `float` literals, because I think `ArrayLike` is a little opaque.

femtomc:
@littleredcomputer this error occurs because we likely missed some of the annotation changes (or clobbered them) in some parts of the code base.

femtomc (approved):
This LGTM! Thanks for adding the defaults to the static language transformations.

We'll push this into `main`, and then rebase `nightly` before our next big release PR.

jburnim:
Thanks!

PR #921, opened 2024-02-09
Adds test case for a Repeat inside of an Unfold

jburnim:
And attempts to fix some small issues in Repeat so that the test passes.

(Please feel free to just treat this PR as a bug report.  I find it helpful for my learning to try to put together a fix as part of reporting the issue -- but I am still not very familiar with GenJAX's codebase and design.)


src/genjax/_src/generative_functions/combinators/vector/repeat_combinator.py
```diff
@@ -60,7 +60,8 @@ def get_retval(self):
 
     @typecheck
     def project(self, selection: Selection):
-        return self.inner_trace.project(selection)
+        return jnp.sum(
+            jax.vmap(lambda it: it.project(selection))(self.inner_trace))
```
jburnim: I'm not sure if this makes sense or fits in with GenJAX's design, but the approach I took here is:

 1. To treat the contract of `Trace.score` and `Trace.project` as that they always operate on a "single instance" of a trace.  So `trace.get_score` returns a single, scalar score.  And `trace.project` also returns a single, scalar score?

 2. But `self.inner_trace` has an extra batch dimension (relative to whatever `class(self.inner_trace)` expects), because, e.g., we vmapped the inner generative function in `RepeatCombinator.simulate`.  To handle that extra batch dimension, we need to also `vmap` the inner `project` function.

 3. And it is safe to do `jnp.sum` here or in `get_score` to return a scalar score.  (Because, e.g., if we're inside an `UnfoldTrace`, then `UnfoldTrace.get_score` or `UnfoldTrace.project` is responsible for vmapping us.)

femtomc:
@jburnim Thanks for the PR! Do you mind rebasing on `nightly`? We're doing PRs into `nightly`, and then release PRs to `main`.

Sorry, we should make this clear somewhere. I can help you sort the differences on `nightly` on Monday, if you'd like.

jburnim:
The PR has been rebased on `nightly`.

sritchie:
@jburnim @littleredcomputer merge conflict fixed, this is ready for a look now.

sritchie:
@jburnim , I believe that this issue should be fixed on `main`; `repeat_combinator` now uses `vmap_combinator` internally: https://github.com/probcomp/genjax/blob/main/src/genjax/_src/generative_functions/combinators/repeat_combinator.py#L37-L67

Let us know if you still have a problem here, apologies for all of the delay.

PR #922, opened 2024-02-09
Selection : Address -> Bool

littleredcomputer:
this gets to the point of everything working except the combinators, which is not exactly a small thing: at this point, we're not JAX trace compatible.

That said, it collapses a whole type hierarchy into one native Python type.


femtomc:
Functionality merged into #1072, closing.

PR #934, opened 2024-02-13
Fix issue where scores/weights could be floats (instead of ArrayLikes)

ursk:
Alternative take on https://github.com/probcomp/genjax/pull/920, replacing FloatArray annotations with ArrayLike. This also includes a new test case `test_assess_with_no_choices` that fails without these changes.


sritchie:
Thanks @ursk , I'm merging this once tests pass!

PR #935, opened 2024-02-13
Remove accidental test secret printout

sritchie:
This works and I didn't mean to commit this step!


horizon-blue (approved):
Thankfully GitHub obfuscates the secret so we haven't leaked anything yet XD

PR #963, opened 2024-02-28
Colin/sel2

littleredcomputer:
Discussed [here](https://www.notion.so/chi-mit/Selections-Idea-2-Trace-Slices-501b9f7f51e149e2af93d6f53ff3a599?pvs=4)

all instances of `Trace.project` and `Choice.filter` in the unit tests now check both syntax forms


femtomc:
@littleredcomputer Is there a way to implement this PR in such a way that the new syntax _only relies on `filter_selection`_?

In other words, I'd prefer _if we relied on the guarantee_ that any `TraceSlice` syntax is directly mappable to a `Selection`, instead of introducing new functionality for new syntax, that is separate from the `Selection` interface.

@littleredcomputer I think, paired with https://github.com/probcomp/genjax/pull/922 -- we'll have the same level of flexibility (instead of trying to parse a `TraceSlice` into one of the existing types of `Selection`, we can create the `has_addr` (Selection as functions) directly, even when we want to support `re` as a means of specifying the `Selection`.

PR #974, opened 2024-03-04
Add basic msgpack serialization

limarta:
This provides serialization using the MsgPack backend. The implementation largely stems from[ flax's implementation](https://flax.readthedocs.io/en/latest/_modules/flax/serialization.html#msgpack_serialize).

- Supports static and combinator traces and Jax tensors.

Usage:

```python
@genjax.static_gen_fn
def model(a, b):
    p = genjax.beta(a,b) @ "p"
    v = genjax.flip(p) @ "v"
    b = genjax.bernoulli(0.25) @ "b"
    return v
tr = model.simulate(key, (jnp.array([0.1, 0.1, 0.1]), 0.5))

payload = msgpack_serialize.serialize(tr)
retrieved_tr = msgpack_serialize.deserialize(payload, model)
```


src/genjax/_src/core/serialization/msgpack.py
```diff
@@ -0,0 +1,64 @@
+import msgpack
+import jax
+import jax.numpy as jnp
+import numpy as np
+from genjax._src.core.serialization.backend import SerializationBackend
+from genjax._src.generative_functions.combinators.staging_utils import get_trace_data_shape
+
```
sritchie, line 26: Please add class docs, method docs, also this needs tests.

pyproject.toml
```diff
@@ -49,6 +49,7 @@ equinox = "^0.11.3"
 typing_extensions = ">=4.6.0,<5"
 oryx = "^0.2.6"
 deprecated = "^1.2.14"
+msgpack = "^1.0.7"
```
sritchie: is there some way @limarta to make this an optional dependency? These are called "provided dependencies" in other systems... @femtomc do you have thoughts on a dep like this? should we have an extras group, so someone can do `genjax[serializable]`?

sritchie (changes requested):
@limarta :

- needs tests
- handle the TODO
- class and method docs, please
- add a tutorial notebook showing how to perform serialization

@femtomc where else should we add docs for this new feature?

pyproject.toml
limarta: The same can be said for `dill` which is also in the toml. Also would the import statements look a bit different if `msgpack` was an extra?

femtomc: I think we should also remove `dill`, as Sam suggested.

femtomc:
Re -- this is awesome @limarta! Thanks for taking the first steps on serialization for GenJAX.

femtomc:
> @femtomc where else should we add docs for this new feature?

We should also add a new documentation page under `core/serialization.md` that documents (a) the serialization interfaces and (b) the supported backends.

pyproject.toml
femtomc: @sritchie Have we figured out how to do extras groups yet?

sritchie: This is the way to do it: https://python-poetry.org/docs/pyproject/#extras

It's not possible with poetry to do what jax does for its `cuda11` and `cuda12` extras because poetry won't let you include different clashing versions of the same-name-different-version dependency in two extras...

but we don't have that problem here!

sritchie:
@limarta can you add the new tutorial to the quarto.yml file? After that this looks good!

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "genjax"
-version = "0.0.1"
```
sritchie: @limarta can you revert this? we want this version to stay inconsequential (it could even be 0.0.0) and let the dynamic versioning assign it when requested. 

And once we merge this the version will change and render this inaccurate, since it's git-sha-based)

pyproject.toml
```diff
@@ -49,6 +49,7 @@ equinox = "^0.11.3"
 typing_extensions = ">=4.6.0,<5"
 oryx = "^0.2.6"
 deprecated = "^1.2.14"
+msgpack = {version = "^1.0.8", optional = true}
```
sritchie: remove this since you already have it below in dev right?

noxfile.py
```diff
@@ -179,6 +179,7 @@ def docs_build(session: Session) -> None:
     session.run_always(
         "poetry", "install", "--with", "docs", "--with", "dev", external=True
     )
```
sritchie: let's revert this file, no need for the no-op change

notebooks/introduction/intro_to_genjax.ipynb
```diff
@@ -1334,7 +1334,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.11.4"
+   "version": "3.11.8"
```
sritchie: can you revert all of these changes? I wish this version wasn't baked in but let's keep the PR smaller

docs/library/index.md
```diff
@@ -2,7 +2,7 @@
...
 
-The `genjax` package consists of several modules, many of which rely on functionality from the `genjax.core` module, and build upon the datatypes, transformation interpreters, and generative datatypes which are defined there. [Generative function languages](./generative_functions/) use the core datatypes and infrastructure to implement the generative function interface. [Inference algorithms](./inference/) are then implemented using the interface (and inference algorithms which utilize _learning_ (like variational inference) also use a [new extension to forward mode AD](./adev.md)).
+The `genjax` package consists of several modules, many of which rely on functionality from the `genjax.core` module, and build upon the datatypes, transformation interpreters, and generative datatypes which are defined there. [Generative function languages](./generative_functions/index.md) use the core datatypes and infrastructure to implement the generative function interface. [Inference algorithms](./inference/index.md) are then implemented using the interface (and inference algorithms which utilize _learning_ (like variational inference) also use a [new extension to forward mode AD](./adev.md)).
```
sritchie: what happened without these changes?

pyproject.toml
```diff
@@ -165,7 +168,7 @@ show_error_codes = true
 show_error_context = true
 
 [tool.poetry-dynamic-versioning]
-enable = true
+enable = false
```
sritchie: re-enable this, please! All of our release tooling depends on this. Is there some reason you had to turn it off?

src/genjax/_src/core/serialization/backend.py
```diff
@@ -15,15 +15,36 @@
 import abc
 from dataclasses import dataclass
 
-from genjax._src.core.datatypes.generative import Trace
+from genjax._src.core.datatypes.generative import GenerativeFunction, Trace
+
+"""This module contains a trace serialization interface that interacts with different backend implementations. Pickle or MsgPack may be used as a backend."""
```
sritchie: I don't see how to switch between the backends. Can you show how to switch between these modes in the doc?

sritchie (changes requested):
Few notes!

src/genjax/_src/core/serialization/backend.py
limarta: Right now the backends are exposed as two separate classes. I will update the docs to show this. Later we can let users set a flag.

PR #978, opened 2024-03-05
Add pyright pre-commit hook

sritchie:
This PR:

- Adds a pre-commit hook for poetry, running both `poetry-check` to make sure that the `pyproject.toml` is well-formed AND running `poetry install` for the user after checkout and after a merge to make sure deps stay up to date
- Adds a pre-commit hook for the pyright type checker
- deletes the `mypy` type checker entry in `noxfile.py`
- adds a new `prepare` step for installing dependencies.to `noxfile.py`
- all noxfile.py tasks now run `poetry install` with the `--all-extras` flag
- `tinygp` and `blackjax` are now properly optional dependencies. This is required for extras to work as pointed out by `poetry-check`

Obviously we have many type errors to fix. I vote to merge this and then start chipping away at the problems ASAP so we can get to a good state. @femtomc , wdyt?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,3 +34,10 @@ repos:
 
       - id: ruff-format
         types_or: [ python, pyi, jupyter ]
+
+
+repos:
+  - repo: https://github.com/RobertCraigie/pyright-python
+    rev: v1.1.352
+    hooks:
+    - id: pyright
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,6 +26,12 @@ repos:
     hooks:
       - id: nbstripout
 
+  - repo: https://github.com/python-poetry/poetry
+    rev: 1.7.1
+    hooks:
+      - id: poetry-check
+      - id: poetry-install
+
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.3.0
     hooks:
@@ -35,9 +41,8 @@ repos:
       - id: ruff-format
         types_or: [ python, pyi, jupyter ]
 
-
 repos:
   - repo: https://github.com/RobertCraigie/pyright-python
-    rev: v1.1.352
+    rev: v1.1.358
     hooks:
     - id: pyright
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+default_install_hook_types:
+  - pre-commit
+  - post-checkout
+  - post-merge
+
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v2.3.0
@@ -27,7 +32,7 @@ repos:
       - id: nbstripout
 
   - repo: https://github.com/python-poetry/poetry
-    rev: 1.7.1
+    rev: 1.8.2
     hooks:
       - id: poetry-check
       - id: poetry-install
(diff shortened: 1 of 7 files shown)


femtomc:
Not sure if we want this yet, as there's many things that `pyright` doesn't understand

sritchie:
Agree, this is not for merging yet, I wanted to see what the type lords had to say.

femtomc:
The type lords are displeased with the insolence shown in this project.

.pre-commit-config.yaml
```diff
@@ -12,6 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+default_install_hook_types:
+  - pre-commit
+  - post-checkout
```
sritchie, line 17: we can remove these if you like, @femtomc ... these force a `poetry install` on checkout so that pyright works, but maybe this is annoying.

.pre-commit-config.yaml
```diff
@@ -26,6 +31,12 @@ repos:
     hooks:
       - id: nbstripout
 
+  - repo: https://github.com/python-poetry/poetry
+    rev: 1.8.2
+    hooks:
+      - id: poetry-check
+      - id: poetry-install
```
sritchie, line 38: If I remove those from above, I can remove this poetry-install hook.

noxfile.py
```diff
@@ -37,9 +37,16 @@
 nox.options.sessions = ("tests", "lint", "build")
 
 
+@session(python=python_version)
+def prepare(session):
+    session.run_always(
+        "poetry", "install", "--with", "dev", "--all-extras", external=True
```
sritchie, line 43: I pulled this out rather than modify all of the subtasks to install extras.

noxfile.py
```diff
@@ -53,12 +60,13 @@ def tests(session):
         "benchmarks",
         "-n",
         "auto",
+        external=True,
```
sritchie, line 63: this killed a warning on tests that poetry wasn't installed in the virtualenv. Probably doesn't matter, since I didn't do it anywhere else!

noxfile.py
```diff
@@ -132,40 +140,31 @@ def safety(session) -> None:
     session.run("safety", "check", "--full-report", f"--file={requirements}")
 
 
-@session(python=python_version)
-def mypy(session) -> None:
```
sritchie, line 136: we have moved on to pyright, RIP mypy...

poetry.toml
```diff
@@ -0,0 +1,2 @@
+[virtualenvs]
+in-project = true
```
sritchie, line 2: this forces the virtualenv creation to live at `.venv`, which is needed by mypy to run in the right env.

pyproject.toml
```diff
@@ -34,6 +34,7 @@ classifiers = [
 Changelog = "https://github.com/probcomp/genjax/releases"
 
 [tool.poetry.dependencies]
+blackjax = { version = "^0.9.6", optional = true }
```
sritchie, line 37: this is the right way to declare extras, taught to me by the `poetry-check` hook.

pyproject.toml
```diff
@@ -88,7 +88,7 @@ mkdocstrings = { extras=["python"], git = "https://github.com/mkdocstrings/mkdoc
 tinygp = ["tinygp"]
 blackjax = ["blackjax"]
 msgpack = ["msgpack"]
-all = ["blackjax", "tinygp", "msgpack"]
+all = ["blackjax", "msgpack", "tinygp"]
```
sritchie, line 91: sort!

PR #981, opened 2024-03-08
Update MaskingTrace.get_choices() to just return the inner choices

jburnim:
I don't think it makes sense to Mask the choices -- the choices for the inner generative function still need to be generated or specified, whether the mask is True or False.

And, in the previous code, MaskingCombinator's assess, importance, and update methods don't look at `choice.flag` for their `choice` argument. (Because the mask must be specified by the input to the generative function, not by its choice map.)

This change also makes it easier for users to construct/update choice maps for masked generative functions (because they don't have to specify the mask in the choice map).


femtomc:
Hi @jburnim -- sorry, we've internally been busy with some conference deadlines, but I'll get to this as soon as early next week.

Related:
> the choices for the inner generative function still need to be generated or specified

I'm not sure if this reasoning is valid for "should we mask or not"? 

Yes, it is true that those choices are still generated, but it is important to correctly mask, even for `get_choices` -- for instance, when a `Trace` or choice map from a `MaskingCombinator` is used as a proposal for importance sampling, it's possible that this method is called.

If we don't mask, it's possibly we induce an incorrect weight computation (because the values of the densities of random choices are not accumulated on the score of the proposal) but the constraints would applied in the model.

femtomc:
Functionality into #1072.

PR #986, opened 2024-03-12
(ADEV) Remove `reap_key`, support swapping of keys in continuation calls in gradient strategies.

femtomc:


femtomc:
Merged into #1072, closing.

PR #987, opened 2024-03-14
Colin/console

littleredcomputer:
attempt to simplify by settling on one interface (`__rich__`), in an effort to get the best of both worlds (rich and Jax).


src/genjax/_src/console.py
```diff
@@ -117,7 +117,7 @@ def console(
 
     finally:
         return GenJAXConsole(
-            Console(soft_wrap=True),
+            Console(soft_wrap=True, **pretty_kwargs),
```
sritchie: the pyright error below is odd. I guess they want `traceback_kwargs = {}` outside the try?

sritchie:
Closing in favor of penzai.

PR #988, opened 2024-03-15
`Selection` as a functional DSL

femtomc:
This PR supercedes https://github.com/probcomp/genjax/pull/922.

This removes specialized `Selection` types (like `HierarchicalSelection`, `IndexedSelection`, etc) and clarifies _what exactly an address is_ in GenJAX, and how addresses are constructed from static and dynamic components.

`Selection` is still used in the same way (e.g. `choice_map.filter(...)` and `trace.project(...)`). 

The key functionality is extracted into a single function of signature `has_addr :: AddressComponent -> Tuple[Bool, Selection]` (roughly, a selection should be able to answer: do I have this address component, and, given this component, do I contain anything after it?)

The functional DSL for constructing selections is inspired by query languages, and has overloads which make expressing complicated selections concise. The DSL is here: https://github.com/probcomp/genjax/blob/dce96f0905efd90b97ca0f2a7c3d3ec529f975f1/src/genjax/_src/core/datatypes/generative.py#L47-L261

There's new functionality, which includes the ability to create "masked" selections (convenient: the only thing which is affected by masking is the `Bool` returned in the `has_addr` signature).

There's also new functionality, due to the fact that the DSL is JAX compatible by default -- meaning that `jax.vmap(...)` on selections which select over _dynamic_ addresses will implicitly create "or" selections across those addresses. By virtue of the JAX compatibility, `vmap` can be useful in other ways: 
* `ChoiceMap` authors can also `vmap(...)` the `has_addr` function of dynamic selections in their `ChoiceMap.filter` or `Trace.project` implementations.
* Users can `vmap(choice_map.filter, in_axes=...)(selection)` to create multiple slices of a choice map, across dynamic indices.


pyproject.toml
```diff
@@ -41,7 +41,7 @@ tensorflow-probability = "^0.23.0"
 rich = "^13.7.1"
 jaxtyping = "^0.2.25"
 optax = "^0.1.9"
-beartype = "^0.16.4"
```
littleredcomputer, line 45: do you need this for this PR? otherwise, it's showing as a merge conflict probably because of the autoupdater, so I'd kick out this and the poetry.lock change if possible

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -13,199 +13,323 @@
...
-    def complement(self) -> "MapSelection":
-        return ComplementMapSelection(self)
+    @classproperty
+    def n(cls) -> "Selection":
+        return select_none
 
-    @abstractmethod
-    def get_subselection(self, addr) -> "Selection":
-        raise NotImplementedError
+    @classproperty
```
littleredcomputer, line 160: why not just put `a = select_all`  just under the class definition

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -13,199 +13,323 @@
...
+    def a(cls) -> "Selection":
+        return select_all
 
-    @abstractmethod
-    def has_addr(self, addr) -> BoolArray:
-        raise NotImplementedError
+    @classmethod
+    @typecheck
```
littleredcomputer: simpler to `@typecheck` the class?

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -13,199 +13,416 @@
...
+            addr_comps = (addr_comps,)
+
+        def _comp_to_sel(comp: AddressComponent):
+            if isinstance(comp, EllipsisType):
+                return Selection.a
+            elif isinstance(comp, DynamicAddressComponent):
+                return Selection.idx(comp)
+            # TODO: make this work, for slices where we can make it work.
+            elif isinstance(comp, slice):
+                idxs = jnp.arange(comp)
```
littleredcomputer: this breaks, so maybe throw if we get to this point, or convert to a real index set using the `slice.indices` technique and add a test

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -13,199 +13,416 @@
...
+    def __invert__(self) -> "Selection":
+        return select_complement(self)
+
     def complement(self) -> "Selection":
-        """Return a `Selection` which filters addresses to the complement set of the
-        provided `Selection`.
+        return select_complement(self)
+
+    @classmethod
+    def get_address_head(cls, addr: Address) -> Address:
```
littleredcomputer, line 118: One thing I found in my PR is that it makes sense to insist that Address is a tuple beyond the outermost frontier of the interface, otherwise, you have to check for singleton vs. tuple literally everywhere. By the frontier I mean places like `__getitem__`, where it is natural to have either a singleton or tuple. Consider having such places promote singletons to `(s,)` and then have Address be a tuple everywhere inside, enabling the r[0], r[1:] tuple idiom where you need it

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -13,199 +13,416 @@
...
+            remaining = self.step(head)
+            return remaining.do(tail)
+        else:
+            return self.has(())
+
+    def __getitem__(self, addr: Address) -> BoolArray:
+        return self.do(addr)
+
+    def __contains__(self, addr: Address) -> BoolArray:
+        return self.has(addr)
```
littleredcomputer, line 154: easy enough to add tests for all the citations coverage exposures in this file

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -13,199 +13,416 @@
...
-    def has_addr(self, addr):
-        assert isinstance(self.selection, MapSelection)
-        return jnp.logical_not(self.selection.has_addr(addr))
+            print(console.render(check))
+            ```
 
-    def get_subselection(self, addr):
-        assert isinstance(self.selection, MapSelection)
-        return self.selection.get_subselection(addr).complement()
+           These objects are JAX compatible, meaning you can create selections over multiple indices using `jax.vmap`, and pass these objects over `jax.jit` boundaries:
```
littleredcomputer, line 252: Could you supply an example of what this would look like / when you would want it? IIUC, one might do jax.vmap(s, [1,2]) and expect to get `s'[1, s] | s'[2, s]` but that's O(n) to explore and we could fix that by letting `1:3` as a slice enter that position which I think would be superior to vamp in both in notational concision and efficiency unless I'm missing something (although, if your point is to expose the possibility as an exposition of the combinators' functional natures, go ahead)

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -13,199 +13,416 @@
...
-#######################
+            print(console.render((check1, check2)))
+            ```
+        """
+        return SelectionBuilder()
 
+    @classmethod
+    @typecheck
+    def r(cls, selections):
+        return reduce(or_, selections)
```
littleredcomputer, line 276: any chance you'd want select_nothing as the base case of the reduce so that the algebra can handle empty sets (similarly select_all for the `__and__`)

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -13,199 +13,416 @@
...
+    return True, select_all
 
-    ###################
-    # Pretty printing #
-    ###################
 
-    def __rich_tree__(self):
-        return rich_tree.Tree("[bold](AllSelection)")
+@typecheck
+def select_defer(
```
littleredcomputer, line 296: it's unclear what defer means in this context. What is different between this and the the expression `lambda: s` ? Is it because you want to check flags at runtime?

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -13,199 +13,416 @@
...
+        # head has type jnp.array with dtype = int, shape = ().
+        # slc is a statically known slice.
+        # we need to check
+        # slice := slice(lower, upper, step)
+        lower, upper, step = slc[0], slc[1], slc[2]
+        check1 = head >= lower if lower else True
+        check2 = head <= upper if upper else True
+
+        # TODO: possibly doesn't work.
+        check3 = head % step == lower if lower else True
```
littleredcomputer, line 370: Yeah, I think we need a test for this: I think the case `::k` doesn't work here (meaning every k'th element, no matter where the array starts, however long it is)

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -13,199 +13,416 @@
...
+
+@typecheck
+def select_and(s1: Selection, s2: Selection) -> Selection:
+    @Selection
+    @Pytree.partial(s1, s2)
+    @typecheck
+    def inner(s1: Selection, s2: Selection, head: AddressComponent):
+        check1, remaining1 = s1.has_addr(head)
+        check2, remaining2 = s2.has_addr(head)
+        check = staged_and(check1, check2)
```
littleredcomputer, line 387: why can't we return false at this point?

littleredcomputer: the asymmetry between and and or, i.e. whether we `defer(and(r1, r2))` vs `or(defer(r1), defer(r2))` is intereseting and might make for an illustrating comment

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -385,11 +530,33 @@ def model():
...
-        raise Exception(
-            f"`get_selection` is not implemented for choice map of type {type(self)}",
-        )
+        raise NotImplementedError
+
+    ###################
+    # Slice interface #
+    ###################
+
+    class Slice(Pytree):
```
littleredcomputer, line 537: Is this is use?

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -398,26 +565,164 @@ def get_selection(self) -> "Selection":
...
                 return submap
         else:
-            return self.__getitem__((addr,))
+            return submap
+
+    def __haskey__(self, addr: Address):
+        return self.has_submap(addr)
+
+
+class EmptyChoice(ChoiceMap):
```
littleredcomputer, line 582: Let's improve the coverage

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -398,26 +565,164 @@ def get_selection(self) -> "Selection":
...
+    def get_selection(self) -> Selection:
+        return Selection.a
+
+    def __rich_tree__(self):
+        tree = rich_tree.Tree("[bold](ChoiceValue)")
+        tree.add(gpp.tree_pformat(self.value))
+        return tree
+
+
+class SelectionChoiceMap(ValueLike, ChoiceMap):
```
littleredcomputer, line 653: We should either test this stuff or save it for another PR

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -539,22 +844,12 @@ def get_gen_fn(self) -> "GenerativeFunction":
...
 
-    @dispatch
-    def project(
-        self,
-        selection: NoneSelection,
-    ) -> FloatArray:
-        return 0.0
-
-    @dispatch
+    @abstractmethod
```
littleredcomputer, line 842: May want to deprecate the old signature if we're not rolling the version

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -854,9 +1119,24 @@ def _check():
...
+    @typecheck
+    def match(self, some: Callable) -> Any:
+        v = self.unmask()
+        return some(v)
+
+    @typecheck
+    def safe_match(self, none: Callable, some: Callable) -> Any:
+        return jax.lax.cond(
+            self.flag,
+            lambda v: some(v),
```
littleredcomputer, line 1130: just say `some`

src/genjax/_src/core/datatypes/generative.py
```diff
@@ -854,9 +1119,24 @@ def _check():
...
+            lambda v: none(),
```
littleredcomputer, line 1131: if the lambda gets a parameter, why does the caller of safe_match not get to see it? if it was supposed to be lambda: none(), just write none

littleredcomputer: The distinction between the algorithms match and safe match is not clear

tests/generative_functions/test_vector_datatypes.py
```diff
@@ -114,20 +114,19 @@ def chain(x):
 
         key, sub_key = jax.random.split(key)
         vec_tr = jax.jit(chain.simulate)(sub_key, (5, 0.0))
-        sel = genjax.indexed_select(jnp.array([0]), genjax.select("z"))
-        proj_score = vec_tr.project(sel)
+        sel = genjax.Selection.at[0, "z"]
+        proj_score = vec_tr.project(key, sel)
```
littleredcomputer, line 118: I still like tr.at[...].project(key) but another time I guess

tests/generative_functions/test_vector_datatypes.py
```diff
@@ -86,8 +86,8 @@ def kernel(x):
         map_over = jnp.arange(0, 50, dtype=float)
         key, sub_key = jax.random.split(key)
         vec_tr = jax.jit(kernel.simulate)(sub_key, (map_over,))
-        sel = genjax.select("z")
```
littleredcomputer, line 89: Do we want a test that asserts that `genjax.select("z")` and `Selection.at[..., "z"]` do the same thing?

tests/generative_functions/test_unfold_combinator.py
```diff
@@ -490,11 +510,18 @@ def chain(z_prev):
             key, sub_key = jax.random.split(key)
             (tr, w, _, _) = chain.update(sub_key, tr, choice, diffs)
 
-        assert tr.project(sel) == tr.get_score()
-        sel = genjax.indexed_select(jnp.array([0, 1, 2]), genjax.select("x", "z"))
-        assert tr.project(sel) == tr.get_score()
-        sel = genjax.indexed_select(jnp.array([0, 1, 2, 3, 4]), genjax.select("x", "z"))
-        assert tr.project(sel) == tr.get_score()
+        assert tr.project(key, sel) == tr.get_score()
```
littleredcomputer, line 513: this is where we'd want `:5`

src/genjax/_src/core/interpreters/staging.py
```diff
@@ -13,37 +13,50 @@
...
+###############################
+# Concrete Boolean arithmetic #
+###############################
 
-class Concrete(Pytree):
-    v: Any = Pytree.static()
 
-    def __hash__(self):
-        return hash(self.v)
+def staged_and(x, y):
```
sritchie, line 31: look at you with these optimizations... I'm surprised jax doesn't do this under the hood? Along with

`False and b` ==> `False`
`True or b` ==> `True`

etc

sritchie (commented):
I gave this a look but I am wildly unqualified to make any comments... deferring to @littleredcomputer here.

src/genjax/_src/shortcuts.py
```diff
@@ -131,20 +128,6 @@ def vector_choice_map(c: ChoiceMappable) -> VectorChoiceMap:
...
-    *choices: Selection,
-) -> IndexedSelection:
-    idx = jnp.atleast_1d(idx)
-    if len(choices) == 0:
-        return IndexedSelection(idx, AllSelection())
-    elif len(choices) == 1 and isinstance(choices[0], Selection):
-        return IndexedSelection(idx, choices[0])
-    else:
-        return IndexedSelection(idx, HierarchicalSelection.from_addresses(choices))
+    return Selection.f(addresses)
```
littleredcomputer, line 133: cool!

tests/generative_functions/test_unfold_combinator.py
```diff
@@ -31,7 +31,8 @@ def kernel(x):
         key, sub_key = jax.random.split(key)
         tr = jax.jit(kernel.simulate)(sub_key, (5, 0.1))
         unfold_score = tr.get_score()
-        assert jnp.sum(tr.project(genjax.select("z"))) == unfold_score
+        sel = genjax.Selection.a >> genjax.Selection.at["z"]
```
littleredcomputer: This is cool. Even if we didn't figure out slices in total generality we could have `:` map to `Selection.a` and that would clean this up a little bit

littleredcomputer (commented):
it's brilliant; I do think this has made the code easier to maintain, since the places were decisions get made are kind of concentrated now

tests/generative_functions/test_vector_datatypes.py
femtomc: These don't do the same thing -- `"z"` is a static address selection at "level 1", whereas `(..., "z")` is an "any selection at level 1, "z" at level 2"

src/genjax/_src/core/datatypes/generative.py
femtomc: We could! I'm a bit partial to the code organization now (at the cost of an indirection).

src/genjax/_src/core/datatypes/generative.py
femtomc: The DSL functions already do typechecking, so maybe I should just remove this.

src/genjax/_src/core/datatypes/generative.py
femtomc: I'm going to throw, and save this for a future PR where we really nail slices.

src/genjax/_src/core/datatypes/generative.py
femtomc: One issue is that we can't actually use the `r[0]` & `r[1:]` tuple idiom _for empty tuple_, which is a valid address.

tests/generative_functions/test_unfold_combinator.py
femtomc: I swapped this for `genjax.Selection.at[..., "z"]`

tests/generative_functions/test_unfold_combinator.py
femtomc: That's right, it's very annoying to specify the exact array.

src/genjax/_src/core/datatypes/generative.py
femtomc: I looked through the places where this check (& safety / coercion, due to the empty tuple thing) via `Selection.get_address_head`, etc) occurs, and outside of `core` it's just in `vector_datatypes` (for `IndexedChoiceMap`)

I feel comfortable saying that we're going to eliminate that anyways as part of the choice map changes.

tests/generative_functions/test_vector_datatypes.py
femtomc: Let's re-visit this -- the current way we're thinking about `project` is a bit off from Gen's semantics (but valid, for the traces which we currently work with), and it's not clear to me that this new syntax (for `project`) is what we'll always have.

PR #1003, opened 2024-03-22
docs: add keyring dependency for poetry

mhuebert:
For Poetry, we need to install keyring to use gcloud credentials to access the artifact repository (just like with pip).


sritchie:
Closing in favor of #1010 , sorry @mhuebert this is great and I'm glad you caught this!

PR #1011, opened 2024-03-26
WIP batching and masking design docs

georgematheos:
I am opening this PR so we can use the comments section to discuss this proposed design for batching and masking systems in genjax.


design_docs/batching.md
```diff
@@ -0,0 +1,60 @@
...
+- `c.simulate` and `c.importance` on  will yield an unbatched
+    trace.  This is totally standard as `c` will be an instance
+    of a subtype of `GenerativeFunction` (and indeed,
+    non-vmapped simulate and importance yielding an
+    unbatched trace is required by the GFI).
+- Let `T` be the return type of the kernel `k`.  Since
+    `k` must be a `JAXGenerativeFunction`, `T` will be a
+    pytree.
+    The return type of the generative function `c`
+    will be a batched instance of `k`, where the batch
```
jburnim: will be a batched instance of `T`?

design_docs/masking.md
```diff
@@ -0,0 +1,389 @@
...
+print(retval)
+# ^this^ prints an `Mask` object where `retval.flag`
+# is [1, 1, 1, 0, 0, 0, 0, 0, 0, 0].
+# `retval.value` will be a 10-vector where only the first 3
+# elements are semantically meaningful.
+
+# Update the trace to have 5 elements
+tr2 = tr.update(
+    key,
+    (jnp.arange(10) < 5, (jnp.zeros(10),))
```
jburnim: Does this call need to include an empty `choices` argument?  Or use `argdiffs=...` for passing the new arguments?

design_docs/masking.md
```diff
@@ -0,0 +1,389 @@
...
+
+The difference between `MaskedChoice` and `Mask` is that
+`MaskedChoice` is also a subclass of `genjax.Choice`, and
+it implements the choicemap interface, as follows.
+
+- `MaskedChoice.isempty()` returns `True` if all values in
+    the `flag` choicemap are `False`,
+    or if `value` is a `genjax.Choice` which is empty.
+- `MaskedChoice.get_value()` returns `value.get_value()`
+    if `flag.get_value()` is `True`, and raises an error otherwise.
```
jburnim: Is the idea that users would call use `MaskedChoice.get_value()` when they wanted this checking and would directly use `MaskedChoice.flag` and `MaskedChoice.value` when they wanted to handle the masking explicitly?  (E.g., are there cases where we have a vector for `flag` and `value`, and we want to grab them both and handle masking ourselves?)

Or that we expect code to always use `MaskedChoice.get_value`, but to use sequences of `get_submap` (and maybe `vmap`, etc.) to ensure they only ever call `get_value()` when `get_flag().get_value()` is True?

design_docs/masking.md
```diff
@@ -0,0 +1,389 @@
...
+The retval of `genjax.mask_combinator(k)` will be a `Mask(flag, val)`
+object where `val` is the retval of `k(a_0, ..., a_n)`,
+and `flag` is the same as the value of the `flag` argument passed to
+`mask_combinator(k)`.
+
+### Accessing values in `Mask` objects
+
+MVP implementation: users are responsible for checking the `flag` array
+and only accessing values where `flag` is `True`.
+I think it is okay to ship a release with this.
```
jburnim: +1

design_docs/masking.md
```diff
@@ -0,0 +1,389 @@
...
+
+The following pattern can be used to simulate having a variable-length
+map.
+```python
+masked_kernel = genjax.mask_combinator(k)
+# masked_kernel takes argument `(flag, (x,))`
+
+mapped_v2 = genjax.map_combinator(
+    in_axes=(0, (0,)),
+    max_length=10
```
jburnim: Is `max_length` needed in this example?

It seems like the max length is always implied by the (statically-known) lengths of the arguments to `mapped_v2`.

(And it seems like this should be true for any `map_combinator(mask_combinator(k))` -- the length of the inputs determines the max length of the map.)

design_docs/masking.md
```diff
@@ -0,0 +1,389 @@
...
+3. This proposal states that we should get the vector combinators
+    to have full GFI coverage, in the sense that any gen choicemap
+    should be able to be passed to `importance` and `update`.
+    Internally, this will be implemented by lifting the input choicemap
+    type to a more structured choicemap type that a lower-level
+    `combinator._importance` method can handle.
+    Should we (now or later) develop a better compositional type system
+    for reasoning about the more restrictive form of choicemaps
+    accepted by `_importance`?
+4. [Feel free to add questions...]
```
jburnim: Worth adding any questions around gradients -- along the lines of https://github.com/probcomp/dcolmap/issues/50 ?

jburnim (commented):
Looks great!  I left a few small comments.

design_docs/batching.md
georgematheos: Yes -- thanks for catching this!

design_docs/masking.md
georgematheos: Good point.

I think this gets at a way that the original design of Gen in Julia differs from JAX conventions.  This is probably a good thing to discuss on Monday.

In the original Gen design, for any `gf : GenerativeFunction` object (like `mapped_v2`), the full GFI would allow us to call `trace.update` to change the arguments to a trace of `gf` to any new argument accepted by `gf`.  In this case, that would include being allowed to update a trace by changing its arguments to an array of a different length.  That's why I proposed having `max_length` in the constructor--so the `gf` object constructed with `max_length` can consider differently-shaped arrays to be invalid inputs.

But I agree that this design feels like it adds needless complexity, in a way I think JAX usually tries to avoid.  So I'm not sure adding `max_length` to the constructor is actually a good move.

design_docs/masking.md
georgematheos: I can explain the idea here in a bit more detail when we discuss on Monday.

My goal is that in most code where users use the masking system directly (e.g. in the definitions of their generative functions), they will use `Mask`.
I have in mind that `MaskedChoice` will mostly be used in lower-level Gen code that needs to read out information from choicemaps.  `MaskedChoice` will mirror the `Choice` interface, so Gen code built for choicemaps can in fact support choicemaps with dynamic structure mediated through dynamic masking.  One example of this is having `importance` calls with constraint choicemaps which may have dynamic structure (and constrain different addresses on different runs).

Note that `get_submap`, `get_value`, `isempty`, and a couple other methods, compose the interface for being a `genjax.Choice` object.  (This is the "choicemap interface" in Gen.)  If I remember correctly, for any choice/choicemap, we should only call `choicemap.get_value()` if `choicemap.isempty()` is false, and this will carry to the `MaskedChoice` as well.

I don't know if there are use cases where users will directly access `MaskedChoice.flag` or not.  This might be good to discuss.

PR #1019, opened 2024-04-05
Maybe jitting callees inside invocations of the `genjax.interpreted` GFI

femtomc:


sritchie:
Closing this, we can re-open if we decide to focus on genjax.interpreted again.

PR #1034, opened 2024-04-08
Fix argument order repeat_combinator.py

mirkoklukas:
`RepeatCombinator` expects:
1. `inner`: JAXGenerativeFunction
2. `repeats`: Int = Pytree.static()

However `map_combinator` returns `RepeatCombinator(repeats, f)`.

I fixed that to `RepeatCombinator(f, repeats)`.


sritchie:
Nice catch. @femtomc is there a quick test we can add / suggest for this?

mirkoklukas:
@sritchie @femtomc Sorry not sure what the best process is: I closed this and opened an issue (#1035) and pointed to my branch (https://github.com/probcomp/genjax/compare/nightly...mkl-repeat-combinator) where I fixed some.

mirkoklukas:
There was another issue with repeat

PR #1036, opened 2024-04-11
Gm/unfold fix 20240411

georgematheos:
@femtomc I made progress adding tests for the Unfold combinator update fix we worked on together.  This PR contains the WIP fix, and the new tests.  I am not quite done debugging.  I wanted to post my progress here.

Note that there is currently one failing test for the branch where the unfold length decreases.  There are also 2 tests I have commented out which I know fail, because the discard is not constructed correctly in the case where the length decreases.


georgematheos:
@femtomc I've fixed bug in this.

One remaining shortcoming of this implementation is that it does not compute a discard choicemap correctly in the case where the length of the unfold combinator shrinks.  My understanding is this will be a bit annoying to add support for, because `EmptyChoice` has a different JAX shape than an `IndexedChoiceMap` (which is what we'd need to return when the size of the unfold shrinks).  It seems possible to me that we should wait to address this until we've improved the Masking system, as we discussed.  If you agree, maybe we should merge this into nightly, even though the implementation is incomplete (as it is better than what was there before)?  If so I can open an issue to remind us to fix this later.

femtomc:
@georgematheos I agree that we should merge this into `nightly` before re-visiting as part of masking / choice maps.

femtomc (approved):

PR #1084, opened 2024-05-06
Anywidget-based plotting

mhuebert:
A `genjax.studio.plot` tool for producing html/js-based plots.

Goals:
- A syntax which is easy to read and write, yet flexible 
- Plotting into a javascript environment we control, so that we can build interactive tools on top such as animation, scrubbing, python<>js dataflow, etc.

Approach:
- The [pyobsplot](https://github.com/juba/pyobsplot) library creates "stubs" in python which directly mirror the Observable Plot API. An AST-like "spec" is created in python and then interpreted in javascript.
- The [Observable Plot](https://observablehq.com/plot/) library does not have "chart types" but rather "marks", which are layered to produce a chart. These are nicely composable via `+` in Python.
- Compared to other common python plotting utilities, this is more functional and often has the effect of condensing and collecting plot-related code.

Basi

TODO
- [x] gen.studio.plot module
- [x] convert intro_to_genjax.py
- [x] convert colin/blocks notebook
- [ ] add interactivity to colin/blocks


mhuebert:
Closing in favor of https://github.com/probcomp/genjax/pull/1120

PR #1086, opened 2024-05-08
supply importance for Repeat with EmptyChoice

littleredcomputer:
this is needed for the Blocks demo where we use Repeat for the coefficients of a degree-N polynomial


femtomc (approved):
LGTM!

sritchie:
🚢

PR #1117, opened 2024-05-22
Documentation for GenJAX 0.4.0

femtomc:


sritchie:
@femtomc can you ask for reviews on these? Docs are an example where they're really written for other people, so would be good to get a user to at least confirm that the new docs make sense.

PR #1118, opened 2024-05-22
Aggressively collapse empty choice maps when constructing choice maps.

femtomc:


femtomc:
@littleredcomputer hold off on reviewing this

femtomc:
Ready now.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -902,7 +913,15 @@ def choice_map_or(
     c1: ChoiceMap,
     c2: ChoiceMap,
 ) -> ChoiceMap:
-    return ChoiceMap(OrChmFn(c1, c2))
+    return (
```
littleredcomputer: I have never seen the `x if y else z` syntax nested before, and I don't love it. How about something like 
```Python
match (c1.static_is_empty(), c2.static_is_empty):
    case True, True:
        return choice_map_empty
    case _, True:
        return c1
    case True, _:
        return c2
    case _, _:
        return ChoiceMap(OrChmFn(c1, c2))
```

littleredcomputer (commented):
I propose using `match` rather than nested `x if ...` because in the former case the condition is paired with the consequence _in that order_ which scans better than what's here which reverses that. No objection to unnested cases of `x if ...`

src/genjax/_src/core/generative/choice_map.py
femtomc: Agreed, changed.

PR #1120, opened 2024-05-23
convert intro notebook to studio.plot

mhuebert:
This PR:

- performs the minimal conversion of the intro notebook to gen.studio.plot
- updates each github actions workflow to authenticate with gcloud to we can install the (private) gen-studio dep


.github/workflows/ci.yml
```diff
@@ -25,6 +25,14 @@ jobs:
         with:
           python-version: 3.11.5
 
+      - id: auth
```
sritchie, line 28: I wish there was a way to share these across workflows... but I think if you do this credential setting in one job that everyone can share, the effects won't persist into the runtime of the new actions.

@mhuebert is that wrong?

.github/workflows/pyright.yml
```diff
@@ -10,17 +10,25 @@ jobs:
   pyright:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
+      - uses: actions/checkout@v4
```
sritchie, line 13: get the formatting consistent...

notebooks/active/intro.ipynb
```diff
@@ -247,20 +248,11 @@
...
-    "jitted = jit(full_model.simulate)\n",
-    "for ax in axes.flatten():\n",
-    "    key, sub_key = jrand.split(key)\n",
-    "    tr = jitted(key, (data,))\n",
-    "    x = data\n",
-    "    y = tr.get_retval()\n",
-    "    viz(ax, x, y, marker=\".\")\n",
-    "\n",
-    "plt.show()"
+    "key, *sub_keys = jrand.split(key, 10)\n",
```
sritchie, line 251: love it!

noxfile.py
```diff
@@ -61,6 +61,13 @@ def install_jaxlib(session):
 
 @session(python=python_version)
 def prepare(session):
+    session.run(
```
sritchie: this was the fix for the initial build errors.

sritchie (changes requested):
1. I think gen-studio should not include a `jaxlib` dependency. JAX wants users to install jaxlib separately depending on GPU vs CPU vs TPU, and the default one is CPU only and will mess with a GPU installation. This dependency is like a "provided" dependency in Java, except Python gives us no way to actually declare one of those...
2. wdyt of making gen-studio an 'extra', vs a main dependency? If someone doesn't want to use it it would be nice to keep the main build slim. They can always do

```python
pip install genjax[studio]
```

to bring studio along for the ride.

.github/workflows/ci.yml
mhuebert: @sritchie I'm not aware of a better way to do this. There are ways of passing data between workflows (see eg. https://stackoverflow.com/questions/67107607/github-actions-pass-data-from-workflow-to-another-workflow) but I don't see how this would let us avoid initializing google auth in each runtime.

mhuebert (commented):
thanks @sritchie for resolving the ci issues!

sritchie (approved):

PR #1128, opened 2024-05-24
fix in_axes typing, bug with _static_check_broadcastable

sritchie:
This PR:

- changes the `in_axes` type across the library to match JAX's type declaration
- handles the `int` and `None` cases in `_static_check_broadcastable`

### DIFF
--- a/src/genjax/_src/core/typing.py
+++ b/src/genjax/_src/core/typing.py
@@ -48,6 +48,8 @@ List = btyping.List
 Optional = btyping.Optional
 Type = btyping.Type
 
+# JAX Type alias.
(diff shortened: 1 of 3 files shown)


src/genjax/_src/core/typing.py
```diff
@@ -48,6 +48,8 @@
 Optional = btyping.Optional
 Type = btyping.Type
 
+# JAX Type alias.
+InAxes = int | None | Sequence[Any]
```
sritchie, line 52: I put this in `vmap_combinator`, but the import from core caused a circular dependency, so here it is for now.

femtomc (approved):

PR #1129, opened 2024-05-24
Collapse argdiffs into UpdateProblem -- finally, a unified interface.

femtomc:


femtomc:
This merges the declaration of argdiffs (information about how the arguments to a generative function have changed) into `UpdateProblem`, eliminating argdiffs as an argument in the signature of update, and greatly simplifying the way we can explain the type of argdiffs (now, as just another way to specify an `UpdateProblem` -- before, as this separate thing which sort of relates to `UpdateProblem`, but you need both to specify the old/new targets (now `UpdateProblem` completely specifies this information)).

PR #1142, opened 2024-05-29
Clarify types in combinator decorators and methods

sritchie:
This PR:

- updates the combinator methods to call the "constructor" versions of the combinators instead of the decorators. This allows the methods to return `GenerativeFunction` instead of `Callable | GenerativeFunction`. (Previously in cases where we had written `GenerativeFunction`, the type checker correctly pointed out that we were in error.)
- updates the combinator decorators to call dataclass constructors internally, also clarifying types
- For the derived combinators, replaces not-actually-decorator functions like `cond_combinator` with `CamelCase` versions like `CondCombinator` so users won't get confused
- Changes the `num_repeats` argument of `RepeatCombinator` to be `n`, not `num_repeats`.
- fixes a bug with `CondCombinator` where predicates other than explicit `0` or `1` would cause neither branch to succeed. Now, anything truthy selects the first branch, anything falsey selects the second branch.

(The old behavior was actually "buggy" — a `0` would select the first branch, a `1` would select the second, which doesn't make sense for branches called `if` and `else`.


src/genjax/_src/core/generative/core.py
```diff
@@ -827,113 +827,111 @@ def vmap(
         *args,
         in_axes: InAxes = 0,
     ) -> "GenerativeFunction":
-        from genjax import vmap_combinator
+        from genjax import VmapCombinator
```
sritchie, line 861: Changing these to the "constructor" forms allows pyright to see that we are in fact returning a `GenerativeFunction`, not the `Callable` part of the branch.

src/genjax/_src/core/generative/core.py
```diff
@@ -827,113 +827,111 @@ def vmap(
...
         self,
         gen_fn: "GenerativeFunction",
         *args,
     ) -> "GenerativeFunction":
-        from genjax import cond_combinator
+        from genjax import CondCombinator
 
         return (
-            cond_combinator(self, gen_fn)(*args)
+            CondCombinator(self, gen_fn)(*args)
```
sritchie: This `CondCombinator` is a function, not a new dataclass... but I still think that it makes sense to push users toward CamelCase versions of these if they want to explicitly pass the generative function in as a first arg.

Users have three options for combinators now:

- CamelCase, pass the gen fn explicitly
- decorator style, `vmap_combinator`
- methods

And we shouldn't encourage them to use the decorator style for explicit passing, etc... too confusing.

(Users don't care about derived vs non-derived combinators.)

src/genjax/_src/generative_functions/combinators/address_bijection_combinator.py
```diff
@@ -189,7 +189,10 @@ def address_bijection_combinator(
     /,
     *,
     address_bijection: dict,
-) -> Callable | AddressBijectionCombinator:
+) -> (
+    Callable[[GenerativeFunction], AddressBijectionCombinator]
```
sritchie, line 193: explicit return types, woohoo!

src/genjax/_src/generative_functions/combinators/cond_combinator.py
```diff
@@ -19,28 +19,35 @@
...
     def argument_mapping(b, *args):
-        idx = jnp.array(b, dtype=int)
+        idx = jnp.array(0 if bool(b) else 1, dtype=int)
         return (idx, *args)
 
-    inner_combinator = switch_combinator(if_gen_fn, else_gen_fn)
+    inner_combinator = SwitchCombinator((if_gen_fn, else_gen_fn))
 
-    return compose_combinator(
+    return ComposeCombinator(
```
sritchie, line 43: internally, calling the constructor forms helps the type system along.

src/genjax/_src/generative_functions/combinators/cond_combinator.py
```diff
@@ -19,28 +19,35 @@
...
         inner_combinator,
-        pre=argument_mapping,
+        argument_mapping,
         info="Derived combinator (Cond)",
     )
+
+
+@typecheck
+def cond_combinator(
```
sritchie, line 52: This isn't a decorator, so we should remove it in the next PR in favor of `CondCombinator`.

src/genjax/_src/generative_functions/combinators/cond_combinator.py
```diff
@@ -19,28 +19,35 @@
...
 
 @typecheck
-def cond_combinator(
+def CondCombinator(
     if_gen_fn: GenerativeFunction,
     else_gen_fn: GenerativeFunction,
 ) -> ComposeCombinator:
     def argument_mapping(b, *args):
-        idx = jnp.array(b, dtype=int)
+        idx = jnp.array(0 if bool(b) else 1, dtype=int)
```
sritchie: THIS is the "bugfix", if we agree. Before, the "predicate" was

- 0 for the `if` branch
- 1 for the `else` branch

anything else would select neither branch, definitely buggy.

To match python's predicate semantics, we now take anything truthy to mean `if`, anything falsey to mean `else`.

sritchie (commented):
Some notes to help the review.

src/genjax/_src/generative_functions/combinators/cond_combinator.py
MathieuHuot: In the same vein, @sritchie could you perhaps add a fix to this issue: https://genjax.gen.dev/index.html?

femtomc: This is going to run into JAX issues…

femtomc: The fix for this should be a type check assertion on the argument mapping function, enforcing that we expect either a Bool or a BoolArray

femtomc: Then, we should make sure what branch is selected agrees with the if-else semantics you described

littleredcomputer: Or we could use jax.numpy.bool_:
```
 jnp.bool_(jnp.array([0,1,2,3,0], dtype=float))
Array([False,  True,  True,  True, False], dtype=bool)
```

femtomc: We should provide a type checking assertion that we expect a scalar bool here.

femtomc: ( Either a Bool or a BoolArray with shape () )

femtomc: (Along with Colin’s fix)

src/genjax/_src/core/generative/core.py
```diff
@@ -949,9 +970,9 @@ def marginal(
         from genjax import marginal
 
         return (
-            marginal(self, select_or_addr=select_or_addr, algorithm=algorithm)(*args)
+            marginal(self, selection=select_or_addr, algorithm=algorithm)(*args)
```
sritchie, line 970: whoops, this was a bug in marginal with the wrong keyword arg name.

src/genjax/_src/core/generative/core.py
```diff
@@ -822,119 +818,144 @@ def treescope_color(self):
...
 
-        return (
-            vmap_combinator(self, in_axes=in_axes)(*args)
-            if args
-            else vmap_combinator(self, in_axes=in_axes)
-        )
+        Args:
+            in_axes ([`InAxes`][genjax.typing.InAxes], optional): indicates how
+            the underlying `vmap` patterns should be broadcast across the
+            input arguments to the generative function. Defaults to 0.
```
littleredcomputer: How about adding a hyperlink to the JAX discussion of this topic, since newbs will have questions about how to specify in_axes.

src/genjax/_src/core/typing.py
```diff
@@ -59,6 +59,20 @@
 
 Value = Any
 
+#################################
+# Trace-time-checked primitives #
+#################################
+
+ScalarBool = Annotated[
+    Bool | BoolArray,
+    Is[lambda arr: jnp.array(arr, copy=False).shape == ()],
```
littleredcomputer, line 68: If we must do this, may I propose `Is[jnp.isscalar]`:
```
>>> import jax.numpy as jnp
>>> jnp.isscalar(3.0)
True
>>> jnp.isscalar(jnp.array(3.0))
True
>>> jnp.isscalar(jnp.array([3.0]))
False
```

littleredcomputer (approved):

src/genjax/_src/core/generative/core.py
sritchie: Done, and though it's helpful I suspect this area of the docs will need much massaging...

src/genjax/_src/core/typing.py
sritchie: Done!

sritchie: Whoops, @littleredcomputer after pushing to CI I see that this fails the tests. A JAX tracer that stands in for a scalar (and passes the existing `Is` test) will fail `jnp.isscalar`...

https://github.com/probcomp/genjax/actions/runs/9498163121/job/26176282853?pr=1142

littleredcomputer (commented):
LGTM

PR #1168, opened 2024-06-13
Tidy combinator decorators, types, various renames (GEN-206, GEN-205)

sritchie:
This PR:

- adds docs for all combinator decorators to the docs site
- adds a `--strict` flag to the `mkdocs` task, so the build fails on docstring build errors
- removes the restrictions on which methods print on the docs page for `GenerativeFunction` etc
- discourages use of classes like `VmapCombinator` in favor of a suite of combinator decorators. Each of these is also the name of a method on `GenerativeFunction`:
  - `vmap`
  - `repeat`
  - `scan`
  - `mask`
  - `or_else`
  - `switch`
  - `mix`
  - `map_addresses`
  - `dimap`
  - `map`
  - `contramap`
- `ComposeCombinator` renamed to `DimapCombinator`
- `CondCombinator` renamed to `OrElseCombinator`
- `address_bijection` field renamed to `mapping`
- `scan`'s `max_length` parameter renamed to `n`, since it's not a max length but a total number of iterations
- each of the `_combinator.py` files loses its `_combinator` suffix, since they're already in a `combinators` folder
- `switch` decorator and method no longer take an explicit list, they're now variadic
- `genjax.typing` now binds the internal `Is` from `ScalarBool` as an explicit `ScalarShaped`
- `genjax.typing` exports `ScalarShaped`, `ScalarFloat`, `ScalarBool` now


src/genjax/_src/core/generative/core.py
```diff
@@ -1055,15 +1051,15 @@ def marginal(
         self,
         /,
         *,
-        select_or_addr: Optional[Any] = None,
+        selection: Optional[Any] = None,
```
littleredcomputer, line 990: Why not supply the default of Selection.all() here? It’s functional, so the non-reevaluation should be safe…

src/genjax/_src/core/generative/core.py
```diff
@@ -1035,17 +1035,13 @@ def dimap(
...
+        return f.dimap(pre=lambda *args: args, post=lambda _, ret: f(ret), info=info)
 
     def contramap(
         self, f: Callable, /, *, info: Optional[String] = None
     ) -> "GenerativeFunction":
         """Returns a new _tuple_ of args, be careful here!"""
-        import genjax
-
-        return genjax.dimap(pre=lambda *args: f(*args), info=info)(self)
+        return f.dimap(pre=lambda *args: f(*args), post=lambda _, ret: ret, info=info)
```
littleredcomputer: Does `pre=f` work here?

sritchie: Whoops I bet it does, what a newbie FP move on my part!!

src/genjax/_src/core/generative/core.py
sritchie: Oh, I did this because of the circular import triggered by putting the type up here. Is there a better way to do this? The `Any` is gross...

femtomc:
What happens to `vmap_combinator`? Are we okay calling it `vmap`?

src/genjax/_src/core/generative/core.py
```diff
@@ -930,47 +958,80 @@ def model(toss: bool):
...
 
-    def mix(self, gen_fn: "GenerativeFunction") -> "GenerativeFunction":
-        from genjax import MixtureCombinator
+    def mix(self, *fns: "GenerativeFunction") -> "GenerativeFunction":
+        import genjax
 
-        return MixtureCombinator(self, gen_fn)
+        return genjax.mix(self, *fns)
 
+    # TODO fix this, expand these args out.
```
sritchie: remove this one

src/genjax/_src/core/generative/core.py
```diff
@@ -930,47 +958,80 @@ def model(toss: bool):
...
 
     #####################
     # GenSP / inference #
     #####################
 
     def marginal(
         self,
-        *args,
-        select_or_addr: Optional[Any] = None,
+        /,
```
sritchie, line 988: TODO force discipline with these

docs/library/generative_functions.md
```diff
@@ -74,22 +79,22 @@ While the programmatic `StaticGenerativeFunction` language is powerful, its rest
 ## Derived combinators
 
 
-::: genjax.ComposeCombinator
+::: genjax.DimapCombinator
```
sritchie: change to point to decorators

docs/cookbook/index.md
```diff
@@ -0,0 +1,3 @@
+# Cookbook
```
sritchie, line 1: We need this to kill a warning that fails the build in strict mode.

docs/library/core.md
```diff
@@ -4,26 +4,10 @@
 This page describes the set of core concepts and datatypes in GenJAX, including Gen's generative datatypes and concepts ([`GenerativeFunction`][genjax.core.GenerativeFunction], [`Trace`][genjax.core.Trace], [`Sample`][genjax.core.Sample], [`Constraint`][genjax.core.Constraint], and [`UpdateProblem`][genjax.core.UpdateProblem]), the core JAX compatibility datatypes ([`Pytree`][genjax.core.Pytree], [`Const`][genjax.core.Const], and [`Closure`][genjax.core.Closure]), as well as functionally inspired `Pytree` extensions ([`Mask`][genjax.core.Mask], [`Sum`][genjax.core.Sum]), and GenJAX's approach to "static" (JAX tracing time) typechecking ([`typecheck`][genjax.typing.typecheck]).
 
 ::: genjax.core.GenerativeFunction
-    options:
```
sritchie, line 7: it's better to show all the methods!

docs/library/generative_functions.md
```diff
@@ -45,51 +46,16 @@ For any serious work, you'll want a way to combine generative functions together
...
-::: genjax.mixture_combinator
-    options:
-        show_root_heading: true
-
-::: genjax.address_bijection_combinator
-    options:
-        show_root_heading: true
+Each of the combinators below is implemented as a decorator. `GenerativeFunction` instances make each combinator available as a method with the same name.
+
+::: genjax.vmap
```
sritchie, line 51: lots of decorators...

mkdocs.yml
```diff
@@ -139,5 +139,5 @@ markdown_extensions:
       generic: true
   - toc:
       title: On this page
-      toc_depth: 2
+      toc_depth: 3
```
sritchie, line 142: this shows off the methods.

noxfile.py
```diff
@@ -194,7 +194,7 @@ def mkdocs(session: Session) -> None:
     build_dir = Path("site")
     if build_dir.exists():
         shutil.rmtree(build_dir)
-    session.run("poetry", "run", "mkdocs", "build", external=True)
+    session.run("poetry", "run", "mkdocs", "build", "--strict", external=True)
```
sritchie, line 197: Thanks to @femtomc for spotting this

src/genjax/_src/core/generative/core.py
```diff
@@ -818,22 +822,17 @@ def treescope_color(self):
     # Combinators #
     ###############
 
-    def vmap(self, in_axes: InAxes = 0) -> "GenerativeFunction":
+    def vmap(self, /, *, in_axes: InAxes = 0) -> "GenerativeFunction":
```
sritchie, line 825: Enforce these like we do elsewhere...

src/genjax/_src/core/generative/core.py
```diff
@@ -858,29 +857,35 @@ def model(x):
             print(tr.render_html())
             ```
         """
-        from genjax import VmapCombinator
+        import genjax
 
-        return VmapCombinator(self, in_axes=in_axes)
+        return genjax.vmap(in_axes=in_axes)(self)
```
sritchie, line 862: all of these use the decorators internally.

src/genjax/_src/inference/sp.py
```diff
@@ -275,12 +275,8 @@ def estimate_logpdf(
...
-    /,
-    *,
-    selection: Optional[Selection] = None,
-    algorithm: Optional[Algorithm] = None,
-) -> Callable | GenerativeFunction:
+    selection: Optional[Selection] = None, algorithm: Optional[Algorithm] = None
+) -> Callable[[GenerativeFunction], Marginal]:
     @Pytree.partial(selection)
     def decorator(
         selection: Optional[Selection],
```
sritchie: I think this was broken before, yeah? the decorator should only be passed one input, the generative function.

PR #1169, opened 2024-06-15
scan, fold, iterate, iterate_final methods / decorators (GEN-291, GEN-292)

sritchie:
This PR implements four combinators on top of `scan` that I believe make more intuitive sense. (the current scan is overloading various behaviors.)

`accumulate` and `reduce`, like in FP:

```python
>>> import genjax
>>> import jax.numpy as jnp

>>> @genjax.gen
... def add(carry, x):
...     return carry + x
... 

# normal call
>>> add.simulate(key, (0,2)).get_retval()
Array(2, dtype=int32, weak_type=True)

# accumulate takes the initial accumulator AND a vector of new entries, then returns a vector of all accumulators seen:
>>> add.accumulate().simulate(key, (0, jnp.ones(4))).get_retval()
Array([1., 2., 3., 4.], dtype=float32)

# reduce has the same sig but only returns the last value:
>>> add.reduce().simulate(key, (0,jnp.ones(10)).get_retval()
Array(10., dtype=float32)
```

`iterate` and `iterate_final`:

```python
>>> @genjax.gen
... def inc(prev):
...     return prev + 1
... 

# normal call
>>> inc.simulate(key, (0,)).get_retval()
Array(1, dtype=int32, weak_type=True)

# iterate takes the initial value (same signature!) and returns a vector of [f(x), f(f(x)), ...]
>>> inc.iterate(n=4).simulate(key, (0,)).get_retval()
Array([1, 2, 3, 4], dtype=int32, weak_type=True)

# iterate_final gives you the final thing, no intermediates
>>> inc.iterate_final(n=10).simulate(key, (0,)).get_retval()
Array(10, dtype=int32, weak_type=True)
```


littleredcomputer (approved):
LGTM

PR #1173, opened 2024-06-21
Docs for ScanCombinator, decorator, methods

sritchie:
This PR:

- Beefs up documentation on `ScanCombinator`, the `genjax.scan` decorator and the `scan` method on `GenerativeFunction`
- renames `max_length` to `length` on `ScanCombinator`, and makes it optional (so that we don't have to specify this if we supply an array over which to scan)
- adds `reverse` and `unroll` arguments that we can pass through to `jax.lax.scan`, with defaults supplied

### DIFF
--- a/docs/library/combinators.md
+++ b/docs/library/combinators.md
@@ -1,8 +1,45 @@
 # Combinators: structured patterns of composition
 
-While the programmatic `StaticGenerativeFunction` language is powerful, its restrictions can be limiting. Combinators are a way to express common patterns of composition in a more concise way, and to gain access to effects which are common in JAX (like `jax.vmap`) for generative computations.
+While the programmatic [`genjax.StaticGenerativeFunction`][] language is powerful, its restrictions can be limiting. Combinators are a way to express common patterns of composition in a more concise way, and to gain access to effects which are common in JAX (like `jax.vmap`) for generative computations.
 
-Each of the combinators below is implemented as a decorator. `GenerativeFunction` instances make each combinator available as a method with the same name.
+Each of the combinators below is implemented as a method on [`genjax.GenerativeFunction`][] and as a standalone decorator.
+
+You should strongly prefer the method form. Here's an example of the `vmap` combinator created by the [`genjax.GenerativeFunction.vmap`][] method:
+
+Here is the `vmap` combinator used as a method. `square_many` below accepts an array and returns an array:
+
+```python exec="yes" html="true" source="material-block" session="combinators"
+import jax, genjax
+
+@genjax.gen
+def square(x):
+    return x * x
+
+square_many = square.vmap()
+```
+
+Here is `square_many` defined with [`genjax.vmap`][], the decorator version of the `vmap` method:
+
+```python exec="yes" html="true" source="material-block" session="combinators"
+@genjax.vmap()
+@genjax.gen
+def square_many_decorator(x):
+    return x * x
+```
+
+!!! warning
+    We do _not_ recommend this style, since the original building block generative function won't be available by itself. Please prefer using the combinator methods, or the transformation style shown below.
+
+If you insist on using the decorator form, you can preserve the original function like this:
+
+```python exec="yes" html="true" source="material-block" session="combinators"
+@genjax.gen
+def square(x):
+    return x * x
+
+# Use the decorator as a transformation:
+square_many_better = genjax.vmap()(square)
+```
 
 ## `vmap`-like Combinators
 
--- a/docs/library/combinators.md
+++ b/docs/library/combinators.md
@@ -1,0 +1,31 @@
+# Combinators: structured patterns of composition
+
+While the programmatic `StaticGenerativeFunction` language is powerful, its restrictions can be limiting. Combinators are a way to express common patterns of composition in a more concise way, and to gain access to effects which are common in JAX (like `jax.vmap`) for generative computations.
+
+Each of the combinators below is implemented as a decorator. `GenerativeFunction` instances make each combinator available as a method with the same name.
+
+## `vmap`-like Combinators
+
+::: genjax.vmap
+::: genjax.repeat
+
+## `scan`-like Combinators
+
+::: genjax.scan
+
+## Control Flow Combinators
+
+::: genjax.or_else
+::: genjax.switch
+
+## Various Transformations
+
+::: genjax.map_addresses
+::: genjax.dimap
+::: genjax.map
+::: genjax.contramap
+
+## The Rest
+
+::: genjax.mask
+::: genjax.mix
--- a/src/genjax/_src/core/generative/core.py
+++ b/src/genjax/_src/core/generative/core.py
@@ -892,11 +892,11 @@ class GenerativeFunction(Pytree):
         - `a` may be a primitive, an array type or a pytree (container) type with array leaves
         - `b` may be a primitive, an array type or a pytree (container) type with array leaves.
 
-        All traced values are nested under an index.
+        The values traced by each call to the original generative function will be nested under an integer index that matches the loop iteration index that generated it.
 
         For any array type specifier `t`, `[t]` represents the type with an additional leading axis, and if `t` is a pytree (container) type with array leaves then `[t]` represents the type with the same pytree structure and corresponding leaves each with an additional leading axis.
 
-        When the type of `xs` (denoted `[a]` above) is an array type or None, and the type of `ys` (denoted `[b]` above) is an array type, the semantics of the returned [`genjax.GenerativeFunction`][] are given roughly by this Python implementation:
+        When the type of `xs` in the snippet below (denoted `[a]` above) is an array type or None, and the type of `ys` in the snippet below (denoted `[b]` above) is an array type, the semantics of the returned [`genjax.GenerativeFunction`][] are given roughly by this Python implementation:
 
         ```python
         def scan(f, init, xs, length=None):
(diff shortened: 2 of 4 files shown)


docs/library/combinators.md
```diff
@@ -0,0 +1,31 @@
+# Combinators: structured patterns of composition
+
+While the programmatic `StaticGenerativeFunction` language is powerful, its restrictions can be limiting. Combinators are a way to express common patterns of composition in a more concise way, and to gain access to effects which are common in JAX (like `jax.vmap`) for generative computations.
```
sritchie: These feel nice to live off in their own file. More discoverable, and appropriate especially as we add more.

src/genjax/_src/core/generative/core.py
```diff
@@ -970,14 +1056,14 @@ def map(
     ) -> "GenerativeFunction":
         import genjax
 
-        return genjax.map(f=f, info=info)
+        return genjax.map(f=f, info=info)(self)
```
sritchie, line 1059: whoops, this was a bug! I know I need tests here, I'll get these in in a further PR (this is tracked in Linear).

docs/library/combinators.md
```diff
@@ -0,0 +1,31 @@
...
+
+Each of the combinators below is implemented as a decorator. `GenerativeFunction` instances make each combinator available as a method with the same name.
+
```
derifatives, line 43: Does it make any sense to *briefly* explain these & the differences between them, oir at least signpost more in-depth documentation? As in, if someone's reading this and doesn't already know the difference between vmap and repeat, or the difference between map, dimap, and contramap, do they know how to find out?

src/genjax/_src/core/generative/core.py
```diff
@@ -877,17 +877,103 @@ def repeat(self, /, *, n: Int) -> "GenerativeFunction":
...
+        unroll: int | bool = 1,
+    ) -> "GenerativeFunction":
+        """
+        When called on a [`genjax.GenerativeFunction`][] of type `(c, a) -> (c, b)`, returns a new [`genjax.GenerativeFunction`][] of type `(c, [a]) -> (c, [b])` where
+
+        - `c` is a loop-carried value, which must hold a fixed shape and dtype across all iterations
+        - `a` may be a primitive, an array type or a pytree (container) type with array leaves
+        - `b` may be a primitive, an array type or a pytree (container) type with array leaves.
+
+        All traced values are nested under an index.
```
derifatives: Is it obvious what this means?

src/genjax/_src/core/generative/core.py
```diff
@@ -877,17 +877,103 @@ def repeat(self, /, *, n: Int) -> "GenerativeFunction":
...
+
+        For any array type specifier `t`, `[t]` represents the type with an additional leading axis, and if `t` is a pytree (container) type with array leaves then `[t]` represents the type with the same pytree structure and corresponding leaves each with an additional leading axis.
+
+        When the type of `xs` (denoted `[a]` above) is an array type or None, and the type of `ys` (denoted `[b]` above) is an array type, the semantics of the returned [`genjax.GenerativeFunction`][] are given roughly by this Python implementation:
```
derifatives: Is this the first mention of `xs` in this docstring? (Did you initially have the Python function first?) It reads like a "forward reference" as written.

src/genjax/_src/core/generative/core.py
```diff
@@ -877,17 +877,103 @@ def repeat(self, /, *, n: Int) -> "GenerativeFunction":
...
+                ys.append(y)
+            return carry, np.stack(ys)
+        ```
+
+        Unlike that Python version, both `xs` and `ys` may be arbitrary pytree values, and so multiple arrays can be scanned over at once and produce multiple output arrays. `None` is actually a special case of this, as it represents an empty pytree.
+
+        The loop-carried value `c` must hold a fixed shape and dtype across all iterations (and not just be consistent up to NumPy rank/shape broadcasting and dtype promotion rules, for example). In other words, the type `c` in the type signature above represents an array with a fixed shape and dtype (or a nested tuple/list/dict container data structure with a fixed structure and arrays with fixed shape and dtype at the leaves).
+
+        Args:
+            n: optional integer specifying the number of loop iterations, which (if supplied) must agree with the sizes of leading axes of the arrays in `xs`. If supplied then the returned generative function can take `None` in place of `xs`, as seen in the Python example above.
```
derifatives: Not sure about "as seen in the Python example above." The Python code doesn't return a generative function. Maybe rephrase?

src/genjax/_src/core/generative/core.py
```diff
@@ -877,17 +877,103 @@ def repeat(self, /, *, n: Int) -> "GenerativeFunction":
...
+            import genjax
+
+
+            @genjax.gen
+            def random_walk(prev, _):
+                x = genjax.normal(prev, 1.0) @ "x"
+                return x, None
+
+
+            scan_fn = random_walk.scan(n=1000)
```
derifatives: I would prefer to call the generative function `random_walk_step` and to rename `scan_fn` as `random_walk`, but it's up to you.

src/genjax/_src/core/generative/core.py
```diff
@@ -877,17 +877,103 @@ def repeat(self, /, *, n: Int) -> "GenerativeFunction":
...
+
+
+            @genjax.gen
+            def add_and_square(sum, x):
+                new_sum = sum + x
+                return new_sum, sum * sum
+
+
+            # notice no `n` parameter supplied:
+            scan_fn = add_and_square.scan()
```
derifatives: Can we come up with an example where it's important that it be a generative function? This example would be fine as ordinary JAX, right?

src/genjax/_src/generative_functions/combinators/scan.py
```diff
@@ -117,14 +118,47 @@ class CheckerboardProblem(UpdateProblem):
 
 @Pytree.dataclass
 class ScanCombinator(GenerativeFunction):
-    """> `ScanCombinator` accepts a kernel_gen_fn generative function, as well as a static
-    maximum unroll length, and provides a scan-like pattern of generative computation.
+    """
+    `ScanCombinator` wraps a `kernel_gen_fn` [`genjax.GenerativeFunction`][] of type `(c, a) -> (c, b)` in a new [`genjax.GenerativeFunction`][] of type `(c, [a]) -> (c, [b])`, where
```
derifatives, line 122: I am confused about the relationship between `scan` and `ScanCombinator`. The docstrings seem largely identical. Is it possible that we have the full docstring only in whichever one is more user facing (probably `ScanCombinator`?), and have the other more minimal docstring point to the full one ("for details, see the docstring of [[other place]].") This is slightly less user friendly but protects against the two going out-of-sync. I don't feel strongly.

derifatives: (And again when it shows up a third time.)

derifatives (approved):
Looks good over all. I left comments but none are blocking IMO.

src/genjax/_src/generative_functions/combinators/scan.py
littleredcomputer: That's a good point, but we would like good doc to show up in the IDE in both cases. Perhaps that is possible to accomplish while still DRY

src/genjax/_src/core/generative/core.py
```diff
@@ -877,17 +877,103 @@ def repeat(self, /, *, n: Int) -> "GenerativeFunction":
...
+
+        Args:
+            n: optional integer specifying the number of loop iterations, which (if supplied) must agree with the sizes of leading axes of the arrays in `xs`. If supplied then the returned generative function can take `None` in place of `xs`, as seen in the Python example above.
+
+            reverse: optional boolean specifying whether to run the scan iteration forward (the default) or in reverse, equivalent to reversing the leading axes of the arrays in both `xs` and in `ys`.
+
+            unroll: optional positive int or bool specifying, in the underlying operation of the scan primitive, how many scan iterations to unroll within a single iteration of a loop. If an integer is provided, it determines how many unrolled loop iterations to run within a single rolled iteration of the loop. If a boolean is provided, it will determine if the loop is competely unrolled (i.e. `unroll=True`) or left completely unrolled (i.e. `unroll=False`).
```
littleredcomputer, line 922: potentially we could lean on JAX documentation here, rather than copying it (for one thing, it could change). We might begin by announcing that the purpose of this is to _adapt_ jax.lax.scan to the GF environment, and mention that it is intended to provide a framework for a variety of loop/accumulate computations that can be fully accelerated

littleredcomputer (approved):

docs/library/combinators.md
sritchie: Done! For the examples you gave I'll be adding docs to those one at a time in separate PRs, and at that point it makes sense to have a short overview at the top of each group of combinators.

src/genjax/_src/core/generative/core.py
sritchie: expanded this a bit

src/genjax/_src/core/generative/core.py
sritchie: noted that it's a forward ref

src/genjax/_src/core/generative/core.py
sritchie: rephrased

src/genjax/_src/core/generative/core.py
sritchie: fixed

PR #1174, opened 2024-06-23
Beef up remaining combinator docs, genjax.{mix,switch} signature change

sritchie:
This PR:

- **breaking!** Turns `genjax.mix` and `genjax.switch` from decorators into into functions from `*gen_fn -> gen_fn`. The decorator form was clunky and not something I want to encourage. Folks can use the methods if they want `f.mix(g)` , or `mix(f, g)`. But `mix(g)(f)` is no longer a thing.
- adds documentation and examples for all combinators, decorators and combinator instance methods

None of the examples have assertions, but these are additional blocks of code that can prevent breakages, as a bit of a backstop.


tests/generative_functions/test_static_gen_fn.py
```diff
@@ -573,6 +573,7 @@ def outlier(prob):
         key = jax.random.PRNGKey(314159)
         proposal = make_gen_fn()
         _tr = proposal.simulate(key, (0.3,))
+        # TODO
```
sritchie, line 576: we need to turn these into proper tests...

src/genjax/_src/core/generative/core.py
```diff
@@ -1025,19 +1081,134 @@ def model(toss: bool):
...
+                return genjax.normal(x, 1.0) @ "y"
+
+
+            @genjax.gen
+            def component2(x):
+                return genjax.normal(x, 2.0) @ "y"
+
+
+            # Create mixture model
+            mixture = component1.mix(component2)
```
littleredcomputer, line 1200: I find it irksome that the logits are not specified when the mixture is created, but I understand that this is the prevailing style

src/genjax/_src/generative_functions/combinators/address_bijection.py
```diff
@@ -64,6 +64,41 @@ def get_score(self):
...
+
+
+        @genjax.gen
+        def model():
+            x = genjax.normal(0.0, 1.0) @ "x"
+            y = genjax.normal(x, 1.0) @ "y"
+            return x + y
+
+
+        mapped_model = model.map_addresses(mapping={"new_x": "x", "new_y": "y"})
```
littleredcomputer, line 93: it seems counterintuitive to not have the map go the other way: written this way, you could have "x" map to more than one thing, since it can occur as a value more than once.

littleredcomputer: also, who does this :) ?

littleredcomputer (commented):
we do need more testing urgently. I am hoping for a bright idea that might allow us to write them more efficiently

littleredcomputer (approved):

PR #1175, opened 2024-06-24
Update URLs to GenJAX website

horizon-blue:
The official GenJAX website has been moved to https://genjax.gen.dev/. However, many places in our documentations still points to https://probcomp.github.io/genjax, which currently leads to the 404 page.

As a simple fix, I replaced all occurrences of `"probcomp.github.io/genjax"` with `"genjax.gen.dev"` in the current GenJAX repo.

## Test Plan

I've manually clicked into each of the new URLs and verified that all of them are valid :).


sritchie:
Thank you @horizon-blue!!!

PR #1185, opened 2024-06-30
Remove AddressBijection + map_addresses, relax decorator type bounds (GEN-322)

sritchie:
This PR:

- removes `AddressBijectionCombinator` and the associated `map_addresses` method/decorator
- removes `AddrMapChm`, `choice_map_address_function`, `ChoiceMap.with_addr_map`
- implements `RepeatCombinator` without the need for the `"_internal"` address (and removing our only usage of `map_addresses`
- changes all combinator method / decorator return types to `GenerativeFunction`. I did this because this gives us more flexibility to change the implementation later without breaking type compatibility, and no one should be explicitly relying on the implementation details anyway 

I removed `AddressBijectionCombinator` because it had a few bugs (mapping `x` to `new_x` left `x` available in the choice map, for example), and because ironing out the behavior of a combinator that could flatten addresses was quite hairy.

For example, if we map with

```python
{...: "_ignore",
 "a": "b",
 (): "random"}

```
it seems like this is valid and will
- skip any address one level down except `_ignore`, which will get flattened
- if I call `get_value()` it’ll get converted to `.get_submap("random").get_value()`
- `"a"`’s mapping will never be seen by anyone

@femtomc notes that we shouldn't have `...` and `()` in the mapping at the same time, but we didn't do any validation and this was trouble waiting to happen imo.


src/genjax/_src/generative_functions/combinators/repeat.py
```diff
@@ -18,44 +18,30 @@
...
-    @gen
-    def expanded_gen_fn(_: IntArray, args: Tuple):
-        return gen_fn(*args) @ "_internal"
-
-    return expanded_gen_fn.vmap(in_axes=(0, None)).contramap(
-        argument_mapping, info="Derived combinator (Repeat)"
+    return (
+        gen_fn.contramap(lambda _idx, args: args)
+        .vmap(in_axes=(0, None))
+        .contramap(lambda *args: (jnp.zeros(n), args))
```
littleredcomputer, line 39: may we call these `retvals` instead of `args`? Does that make sense?

littleredcomputer (approved):
lgtm, with one required caveat: author a test that shows the contramaps of repeat doing their jobs. The tests of repeat I see do test choices made, but not the retval, so let's add a test for that.

src/genjax/_src/generative_functions/combinators/repeat.py
sritchie: Addressed in slack with this note:

- the wrapped gen_fn wants its args
- we want to vmap on some dummy index
- the outer contramap sticks jnp.zeros(n) on as a first arg,
- the inner contramap strips off the index before it hits the gen_fn

so no return value mod.

PR #1187, opened 2024-07-01
Fixing a couple of bugs in first batch of cookbook entries

MathieuHuot:
There's a bug in my usage of the switch combinator which changed since I wrote the notebook so I need to fix it in 2 notebooks (conditionals and dimap combinator), also need to add some more explanations in the dimap combinator.

Only look at commit [5349a0e](https://github.com/probcomp/genjax/pull/1187/commits/5349a0ecd62f8d632cb51435c4f22b34a0e2f629) and more recent ones.


MathieuHuot:
@sritchie  sorry about earlier, it's ready now.

sritchie (approved):

PR #1192, opened 2024-07-01
make gfi_boundary preserve types (GEN-313, GEN-336)

sritchie:
This PR:

- Makes `ChoiceMap` the return type of `get_choices` (with a `@typecheck` annotation so this will fail at runtime if `get_sample` returns something different)
- modifies `gfi_boundary` to preserve the input and output args of the wrapped method, so the IDE can see proper types


littleredcomputer (approved):
lgtm

PR #1193, opened 2024-07-01
kill dataclass type errors with typevar (GEN-337)

sritchie:
This PR kills 9 pyright errors of the form

```
error: "Any | type[Any]" is not a class (reportGeneralTypeIssues)
```

by

- modifying `Pytree.dataclass` to ONLY return a decorator, not `type | decorator`
- adding a type variable so that the type of the wrapped class is preserved


littleredcomputer (commented):
There's an art to getting the decorator to work correctly with and without arguments (esp. if one accepts that such arguments should be keyword-only) as discussed [here](https://stackoverflow.com/questions/3888158/making-decorators-with-optional-arguments#comment65959042_24617244) and [here](https://stackoverflow.com/questions/653368/how-to-create-a-decorator-that-can-be-used-either-with-or-without-parameters)... the typing is subtle though, as in this example from dataclasses.py:
```
@overload
def dataclass(cls: None, /) -> Callable[[type[_T]], type[_T]]: ...
@overload
def dataclass(cls: type[_T], /) -> type[_T]: ...
```
I don't think it's essential but since we are reusing the name dataclass we should strive for minimum divergence from the base concept

sritchie:
@littleredcomputer mind taking another look? I took your advice and cleaned this up, same results, no extra parens on dataclass.

littleredcomputer (commented):
lgtm

PR #1194, opened 2024-07-01
Export Address and AddressComponent from genjax._src.core.generative (GEN-315)

sritchie:
These appear in public signatures and should be exported.


littleredcomputer (approved):
lgtm

PR #1199, opened 2024-07-03
Fix dimap extra tupling (GEN-338)

sritchie:
This PR removes an extra level of tupling in the dimap combinator's `update` implementation. This fixes #1169 's failing test as well.

I added tests of the returned trace's score and retval. I don't remember/know how the score relates to the weight, but getting a weight test into here would probably be a good idea.


littleredcomputer (approved):
lgtm

PR #1200, opened 2024-07-03
Enable debugging notebook

sritchie:
This PR adds the debugging notebook to the quarto build. I got past the previous error by turning the final cell into a non-executable markdown code block.


MathieuHuot (approved):
lgtm!

PR #1204, opened 2024-07-04
Fix incorrect mask update weight (GEN-340)

arijit-dasgupta:
The `update` method of the `Mask` combinator now returns the weight of a trace with all 4 combinations of Argdiffs accounted for. 

Following the pattern of (old trace's `check` --> updated trace's `check`), they are:

1. True --> True
2. True --> False
3. False --> True
4. False -- False

I worked out the appropriate weights for each case with @MathieuHuot, but it would be good if someone could review this too. I also added a test under `tests/generative_functions/test_mask_combinator.py` for these 4 cases.


arijit-dasgupta:
Big thanks to @mugamma for helping me understand the maybe monad to motivate my last commit!

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -133,17 +134,33 @@ def update_change_target(
         match trace:
             case MaskTrace():
                 inner_trace = trace.inner
+                init_check = trace.check
```
littleredcomputer: Let's just say `if not trace.check` here

littleredcomputer (commented):
Looks good to me @arijitnoobstar but regrettably I cannot comment on the weight calculations

src/genjax/_src/generative_functions/combinators/mask.py
femtomc: @arijitnoobstar Can you fix this comment before I approve to merge?

arijit-dasgupta: Just to confirm: you mean removing match/case for `init_check == False` to `if not trace.check`?

arijit-dasgupta:
@femtomc @littleredcomputer I realized that I needed to dispatch the update depending on `trace.check`. My earlier version of Match/Case did not actually trigger the importance problem (I realize now that it can only dispatch match-case on Types in jax and not the actual values). I also can't just do an `if trace.check` conditional because of the JAX constraints. Hence I added a dispatch that calls the right update depending on whether update is trying to run Importance or if the `trace.check` is `True` or `False` (via match-case and `jax.lax.cond` respectively. 

This continues to preserve the weights correctly + resimulates the inner gen_fn when something masked is being updated (a.k.a the random choices made in the unmasked state are thrown away and resimulated with the provided key). This follows the idea that when a computation is unmasked, we sample from the gen_fn it unmasks over instead of taking the dummy sampled values that was masked over. 

Here are a couple motivations for this design choice:
1. If I update a trace (lets call it `tr_1`) from being unmasked to masked (lets call it `tr_2`), I don't just hide the computation in the hopes that I can reveal it again in the future with the exact same random choices, i basically discard away any computation. Now if I were to update that very trace (`tr_2`) from masked to unmasked (lets call it `tr_3`)with a different random key, it should resimulate the computation with different randomness given the fact that I had 'forgotten' the earlier computation. If I do not resimulate (from `tr_2` --> `tr_3`) and merely update the inner gen_fn and return that trace as `tr_3`, any choices that are not constrained by the choicemap in my update problem would be the exact same as in `tr_1` because the update did not change those values. This would happen regardless of the random key provided. I am assuming the case for Exact Densities here, not sure how much this would matter if the inner generative functions used `GenSP` distributions or any kind of unbiased estimates. @mugamma helped me think about this point.
2. This plays in nicely with SMC with the use of a masked scan combinator, where the unmasking at a specific timestep (T=t) would efficiently sample the particles for that timestep (since the resimulation happens only at the step where the sampling is needed) while doing update for T=t. @MathieuHuot helped motivate this need to me.

The resimulation for the `trace.check == False` is done via an update call from the `EmptyTarget` (basically how `.importance` is defined) in the dispatched `update_change_target_from_false` method.

```
inner_trace = EmptyTrace(self.gen_fn)
imp_update_problem = ImportanceProblem(update_problem)

premasked_trace, w, _, _ = self.gen_fn.update(
    key, inner_trace, GenericProblem(tuple(inner_argdiffs), imp_update_problem)
)
```

The one thing I am not sure about is: what is the appropriate retdiff and backwards proposal for the `trace.check == False` case? I tried returning the retdiff and bwd_proposal from the update call from the `EmptyTarget` (from the code above), but that led to pytree shape issues from the jax.lax.cond call because of Type mismatches. This tells me that the structure of the retdiff and bwd_proposal for an`ImportanceProblem` and a normal update might have differing types. To work around this, I did a second update on the trace (from the importance-like update call) with the same update problem (just without wrapping the update problem with `ImportanceProblem` like earlier). This is shown below:

```
_, _, retdiff, bwd_problem = self.gen_fn.update(
   key, premasked_trace, GenericProblem(tuple(inner_argdiffs), update_problem)
  )
```

This helps to avoid that pytree error, but i'm worried that the `retdiff` and `bwd_problem` might be wrong as they do not propose back to the empty trace (as I would expect in a normal `importance` call). At the same time, since the backwards imply proposing back to the null/masked trace, i'm not sure if the `retdiff` or `bwd_problem` even matters in this case. This is hard to avoid because of the need to use `jax.lax.cond` which requires the preservation of pytrees returned from the `trace.check == True` case. If there is a way to better dispatch the `trace.check` without breaking JAX, that would be ideal, but i'm not sure that is possible.

One good news from this update (my last [commit](https://github.com/probcomp/genjax/pull/1204/commits/bf729b8941368c4dc2c86bcdc46c7bdcaa13d1e9) ) is that SMC is working quite nicely on my toy example notebook with the `masked_scan_combinator` since the update move is essentially sampling the time step that matters by simply modifying the argdiff and giving the observation for that timestep.

I anything in this comment is confusing or unclear, i'll be happy to talk about it in-person or via call :-)

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -139,18 +140,77 @@ def update_change_target(
...
+    @typecheck
+    def update_dispatch(
+        self,
+        key: PRNGKey,
+        trace: Trace,
+        update_problem: UpdateProblem,
+        argdiffs: Argdiffs,
+    ) -> Tuple[Trace, Weight, Retdiff, UpdateProblem]:
+        match update_problem:
+            case ImportanceProblem(constraint):
```
sritchie: @arijitnoobstar for the future, go ahead and install the pre-commit hooks - they'll catch errors like this with the ruff linter and help fix them. https://genjax.gen.dev/developing.html#commit-hooks

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -135,32 +137,95 @@ def update_change_target(
...
+
+    @typecheck
+    def update_change_target_from_false(
+        self,
+        key: PRNGKey,
+        trace: Trace,
+        update_problem: UpdateProblem,
+        argdiffs: Argdiffs,
+    ) -> Tuple[Trace, Weight, Retdiff, UpdateProblem]:
+        (check, *_) = Diff.tree_primal(argdiffs)
```
littleredcomputer: I would say `check = Diff.tree_primal(argdiffs)[0]`

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -135,32 +137,95 @@ def update_change_target(
...
+        (check_diff, *inner_argdiffs) = argdiffs
```
littleredcomputer: Regrettably (as you know since you use `tuple()` below) restructuring a tuple with `*` results in a list. May I propose this idiom: 
```
check_diff = argdiffs[0]
inner_argdiffs = argdiffs[1:]
```
or (even better)
```
check_diff, inner_argdiffs = argdiffs[0], argdiffs[1:]
```

The use of slicing in this way preserves tuple-nature

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -135,32 +137,95 @@ def update_change_target(
...
+
+        inner_trace = EmptyTrace(self.gen_fn)
+        imp_update_problem = ImportanceProblem(update_problem)
+
+        premasked_trace, w, _, _ = self.gen_fn.update(
```
littleredcomputer, line 176: I think it is worth a _comment_, in the code, that discusses the need for both updates and how the ultimate result is derived from the two of them.

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -135,32 +137,95 @@ def update_change_target(
...
         key: PRNGKey,
         trace: Trace,
         update_problem: UpdateProblem,
     ) -> Tuple[Trace, Weight, Retdiff, UpdateProblem]:
+        assert isinstance(trace, MaskTrace) or isinstance(trace, EmptyTrace)
+
         match update_problem:
             case GenericProblem(argdiffs, subproblem):
-                return self.update_change_target(key, trace, subproblem, argdiffs)
+                return self.update_dispatch(key, trace, subproblem, argdiffs)
```
littleredcomputer: `update_dispatch` just uses `match` again with finer granularity? Why not fold the sub-match into this place right here. Toss in a one-line comment explaining what's being done in each case.

tests/generative_functions/test_mask_combinator.py
```diff
@@ -53,3 +55,45 @@ def model(x):
...
+        assert w == -tr.get_score()
+
+        # pre-update mask arg is False
+        tr = jax.jit(model.simulate)(key, (False, 2.0))
+        # mask check arg transition: False --> True
+        argdiffs = U.g(
+            (Diff.unknown_change(True), Diff.no_change(tr.get_args()[1])), C.n()
+        )
+        w = tr.update(key, argdiffs)[1]
+        assert w == tr.inner.update(key, C.n())[1] + tr.inner.get_score()
```
littleredcomputer, line 83: I'm curious as to the point of calling `update(..., C.n())` here: shouldn't it be an identity of gen that `tr ~= tr.update(..., C.n())`?

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -135,32 +137,95 @@ def update_change_target(
                 inner_trace = trace.inner
             case EmptyTrace():
                 inner_trace = EmptyTrace(self.gen_fn)
+            case _:
+                raise Exception(f"Unexpected trace type: {trace}")
```
littleredcomputer: use `NotImplementedError`

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -135,32 +137,95 @@ def update_change_target(
...
+
+        return (
+            MaskTrace(self, premasked_trace, check),
+            w,
+            Mask.maybe(check_diff, retdiff),
+            MaskedProblem(check, bwd_problem),
+        )
+
+    @typecheck
+    def update_change_target_from_false(
```
sritchie, line 161: @femtomc , can you make a note here about an error that we should throw for this incorrect move?

src/genjax/_src/generative_functions/combinators/mask.py
sritchie: done

sritchie (approved):

PR #1208, opened 2024-07-06
Prepend initial accumulator in accumulate, iterate (GEN-346)

sritchie:
This PR modifies the `iterate` and `accumulate` combinators to attach the initial accumulator at the front of the array returned by these generative functions.

`jax.lax.scan` does NOT do this because the type signature of `(s -> a -> (s, b)) -> s -> [a] -> (s, [b])` doesn't allow it. (The accumulated type here is `b`, not `s`).

BUT in `iterate` and `accumulate`, `s == b`, and returning the initial accumulator matches the behavior of these functions everywhere they appear (in Clojure, Python, Haskell, Scala etc.).

Thanks to @littleredcomputer for suggesting this and @arijitnoobstar for noting that this would be useful for him as well.

The PR also fixes some docs typos.

### DIFF
--- a/src/genjax/_src/generative_functions/combinators/scan.py
+++ b/src/genjax/_src/generative_functions/combinators/scan.py
@@ -655,10 +655,10 @@ def prepend_initial_acc(args, ret):
     init_acc = args[0]
     xs = ret[1]
 
-    def concat_with_init(init, arr):
-        return jnp.concatenate([jnp.array([init]), arr])
+    def cat(init, arr):
+        return jnp.concatenate([jnp.array(init)[jnp.newaxis], arr])
 
-    return jax.tree.map(concat_with_init, init_acc, xs)
+    return jax.tree.map(cat, init_acc, xs)
 
 
 @typecheck
--- a/src/genjax/_src/generative_functions/combinators/scan.py
+++ b/src/genjax/_src/generative_functions/combinators/scan.py
@@ -634,6 +634,33 @@ def scan(
     return decorator
 
 
+def prepend_initial_acc(args, ret):
+    """
+    Prepends the initial accumulator value to the array of accumulated values.
+
+    This function is used in the context of scan operations to include the initial
+    accumulator state in the output, effectively providing a complete history of
+    the accumulator's values throughout the scan.
+
+    Args:
+        args: A tuple containing the initial arguments to the scan operation. The first element is expected to be the initial accumulator value.
+        ret: A tuple containing the final accumulator value and an array of intermediate accumulator values from the scan operation.
+
+    Returns:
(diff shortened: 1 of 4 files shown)


src/genjax/_src/generative_functions/combinators/scan.py
```diff
@@ -634,6 +634,33 @@ def decorator(f):
...
+    Note:
+        This function uses JAX's tree mapping to handle nested structures in the accumulator, allowing it to work with complex accumulator types.
+    """
+    init_acc, _ = args
+    xs = ret[1]
+
+    def concat_with_init(init, arr):
+        return jnp.concatenate([jnp.array([init]), arr])
+
+    return jax.tree.map(concat_with_init, init_acc, xs)
```
sritchie: @littleredcomputer stare at this with me and let me know if this seems right, and if I should add more tests...

tests/generative_functions/test_scan_combinator.py
```diff
@@ -166,19 +166,19 @@ def test_add(self, key):
         result = add.simulate(key, (0, 2)).get_retval()
         assert result == 2
 
-    def test_scan(self, key):
+    def test_accumulate(self, key):
```
sritchie, line 221: I had forgotten to rename these tests!

src/genjax/_src/generative_functions/combinators/scan.py
littleredcomputer: If you're trying to add an axis to `init` the idiomatic way is `init[jnp.newaxis]`. If you're worried about whether init is a bare scalar (and we should be) then maybe we want `jnp.atleast_1d(init)[jnp.newaxis]`.

littleredcomputer: And we will absolutely want tests for where init is a row vector and a small matrix (for the latter iterating the fibonacci seed matrix with multiplication-by-itself would make a good test). How about this:

```
def cat(x,y):
     return jnp.concatenate((jnp.array(x)[jnp.newaxis], y))

>>> cat (3, jnp.array([4,5,6]))
Array([3, 4, 5, 6], dtype=int32)
>>> cat (jnp.array(3), jnp.array([4,5,6]))
Array([3, 4, 5, 6], dtype=int32)
>>> cat (jnp.array([4,5]), jnp.array([[6,7],[8,9]]))
Array([[4, 5],
       [6, 7],
       [8, 9]], dtype=int32)
```

In this case we do not want `jnp.atleast1d` since that, combined with newaxis, would promote a bare scalar two steps in rank which is one too many

littleredcomputer (approved):
LGTM

PR #1209, opened 2024-07-06
Fix dimap args / retval mapping issue (GEN-345)

sritchie:
This PR:

- adds generic types to `DimapTrace`, `DimapCombinator` and all of the methods that create these types
- fixes two bugs exposed by the typing where the args failed to be transformed before being passed to `retval_mapping`
- adds tests that back up this behavior
- fixes up some `Callable` type bounds

### DIFF
--- a/src/genjax/_src/core/generative/core.py
+++ b/src/genjax/_src/core/generative/core.py
@@ -38,7 +38,6 @@ from genjax._src.core.typing import (
     Is,
     List,
     Optional,
-    ParamSpec,
     PRNGKey,
     String,
     Tuple,
@@ -53,8 +52,7 @@ register_exclusion(__file__)
 if TYPE_CHECKING:
     import genjax
 
-_P = ParamSpec("_P")
-_T = TypeVar("_T")
+_C = TypeVar("_C", bound=Callable)
 
 #####################################
 # Special generative function types #
@@ -528,7 +526,7 @@ class GenerativeFunction(Pytree):
         return jtu.tree_map(lambda v: jnp.zeros(v.shape, dtype=v.dtype), data_shape)
 
     @classmethod
-    def gfi_boundary(cls, c: Callable[_P, _T]) -> Callable[_P, _T]:
+    def gfi_boundary(cls, c: _C) -> _C:
         return gfi_boundary(c)
 
     @abstractmethod
--- a/src/genjax/_src/core/generative/core.py
+++ b/src/genjax/_src/core/generative/core.py
@@ -53,6 +53,9 @@ if TYPE_CHECKING:
     import genjax
 
 _C = TypeVar("_C", bound=Callable)
+ArgTuple = TypeVar("ArgTuple", bound=tuple)
+R = TypeVar("R")
+S = TypeVar("S")
 
 #####################################
 # Special generative function types #
@@ -1407,9 +1410,9 @@ class GenerativeFunction(Pytree):
         self,
         /,
         *,
-        pre: Callable[..., Any],
-        post: Callable[..., Any],
-        info: Optional[String] = None,
+        pre: Callable[..., ArgTuple],
+        post: Callable[[ArgTuple, R], S],
+        info: String | None = None,
     ) -> "GenerativeFunction":
         """
         Returns a new [`genjax.GenerativeFunction`][] and applies pre- and post-processing functions to its arguments and return value.
@@ -1460,7 +1463,7 @@ class GenerativeFunction(Pytree):
         return genjax.dimap(pre=pre, post=post, info=info)(self)
 
     def map(
-        self, f: Callable[..., Any], *, info: Optional[String] = None
+        self, f: Callable[[R], S], *, info: String | None = None
     ) -> "GenerativeFunction":
         """
         Specialized version of [`genjax.dimap`][] where only the post-processing function is applied.
@@ -1501,7 +1504,7 @@ class GenerativeFunction(Pytree):
         return genjax.map(f=f, info=info)(self)
 
     def contramap(
-        self, f: Callable[..., Any], *, info: Optional[String] = None
+        self, f: Callable[..., ArgTuple], *, info: String | None = None
     ) -> "GenerativeFunction":
         """
         Specialized version of [`genjax.GenerativeFunction.dimap`][] where only the pre-processing function is applied.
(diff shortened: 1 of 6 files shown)


src/genjax/_src/generative_functions/combinators/dimap.py
```diff
@@ -190,7 +191,7 @@ def assess(
     ) -> Tuple[Score, Any]:
         inner_args = self.argument_mapping(*args)
         w, inner_retval = self.inner.assess(sample, inner_args)
-        retval = self.retval_mapping(args, inner_retval)
+        retval = self.retval_mapping(inner_args, inner_retval)
```
sritchie, line 202: in `simulate`, we pass `inner_args`, so we should do the same.

src/genjax/_src/generative_functions/combinators/dimap.py
```diff
@@ -153,6 +153,7 @@ def update_change_target(
         def closed_mapping(args, retval):
             return self.retval_mapping(args, retval)
 
+        # TODO I think closed_mapping here is being called with `args`, not `inner_args`?
```
sritchie: I don't see how `incremental` works here, BUT I had a test fail and it seemed like `closed_mapping` was being passed `args` here.

src/genjax/_src/generative_functions/combinators/dimap.py
```diff
@@ -176,9 +177,11 @@ def update(
     ) -> Tuple[Trace, Weight, Retdiff, UpdateProblem]:
         match update_problem:
             case GenericProblem(argdiffs, subproblem):
+                # TODO shouldn't we transform args here too?
```
sritchie: actually I think this is why (see comment above).

sritchie (commented):
Adding some notes, @femtomc . We obviously will need tests too to check that we're passing the transformed args into retval_diff.

src/genjax/_src/generative_functions/combinators/dimap.py
sritchie: answer is, no, since this is a DimapTrace coming in, so args are already transformed.

src/genjax/_src/core/generative/core.py
```diff
@@ -53,8 +52,7 @@
 if TYPE_CHECKING:
     import genjax
 
-_P = ParamSpec("_P")
-_T = TypeVar("_T")
+_C = TypeVar("_C", bound=Callable)
```
sritchie, line 55: turns out that if the paramspec and return don't change we can just do this.

src/genjax/_src/core/traceback_util.py
```diff
@@ -13,16 +13,15 @@
 # limitations under the License.
 
 import jax._src.traceback_util as traceback_util
-from beartype.typing import Callable, ParamSpec, TypeVar
+from beartype.typing import Callable, TypeVar
 
-_P = ParamSpec("_P")
-_T = TypeVar("_T")
+_C = TypeVar("_C", bound=Callable)
```
sritchie, line 18: same note here.

src/genjax/_src/generative_functions/combinators/dimap.py
```diff
@@ -31,41 +31,45 @@
...
 
+ArgTuple = TypeVar("ArgTuple", bound=tuple)
+R = TypeVar("R")
+S = TypeVar("S")
+
 
 @Pytree.dataclass
-class DimapTrace(Trace):
-    dimap_combinator: "DimapCombinator"
+class DimapTrace(Trace, Generic[ArgTuple, S]):
```
sritchie, line 48: might as well track these with generic types!

src/genjax/_src/generative_functions/combinators/dimap.py
```diff
@@ -130,28 +134,34 @@ def update_change_target(
...
         tr, w, inner_retdiff, bwd_problem = self.inner.update(
             key, inner_trace, GenericProblem(inner_argdiffs, update_problem)
         )
+
         inner_retval_primals = Diff.tree_primal(inner_retdiff)
         inner_retval_tangents = Diff.tree_tangent(inner_retdiff)
 
-        def closed_mapping(args, retval):
-            return self.retval_mapping(args, retval)
+        def closed_mapping(args: tuple, retval: R) -> S:
```
sritchie, line 161: boom, the types identified THIS bug:

src/genjax/_src/generative_functions/combinators/dimap.py
```diff
@@ -186,11 +196,11 @@ def update(
...
         self,
         sample: Sample,
-        args: Tuple,
-    ) -> Tuple[Score, Any]:
+        args: tuple,
+    ) -> tuple[Score, Any]:
         inner_args = self.argument_mapping(*args)
         w, inner_retval = self.inner.assess(sample, inner_args)
-        retval = self.retval_mapping(args, inner_retval)
+        retval = self.retval_mapping(inner_args, inner_retval)
```
sritchie, line 202: and this one too!

src/genjax/_src/generative_functions/combinators/dimap.py
```diff
@@ -201,9 +211,9 @@ def assess(
 
 def dimap(
     *,
-    pre: Callable[..., Any] = lambda *args: args,
-    post: Callable[..., Any] = lambda _args, retval: retval,
-    info: Optional[String] = None,
+    pre: Callable[..., ArgTuple] = lambda *args: args,
```
sritchie, line 213: and now the type checker will complain if pre and post don't match.

tests/generative_functions/test_dimap_combinator.py
```diff
@@ -21,16 +21,17 @@ class TestDimapCombinator:
     def test_dimap_update_retval(self):
         # Define pre- and post-processing functions
         def pre_process(x, y):
-            return (x + 1, y * 2)
+            return (x + 1, y * 2, y * 3)
 
-        def post_process(_, retval):
+        def post_process(args, retval):
+            assert len(args) == 3, "post_process has to receive transformed args."
```
sritchie, line 27: happy to add a more detailed test, but this will fail without this fix.

tests/generative_functions/test_dimap_combinator.py
```diff
@@ -44,6 +45,10 @@ def model(x, y):
             -2.5092335 == trace.get_retval()
         ), "initial retval is a square of random draw"
 
+        assert (trace.get_score(), trace.get_retval()) == dimap_model.assess(
```
sritchie, line 48: this will test the assess codepath.

femtomc (approved):

PR #1212, opened 2024-07-09
fixes to ScanCombinator._update

littleredcomputer:
This is a fix to GEN-339: 

- removes an extra layer of array being applied to a scanned function argument during update calculations
- equalizes the Diff type of scan input and output to UknownChange in both cases


arijit-dasgupta:
This is super useful! I had this same bug (which I realize I forgot to report, my bad) and I had to do a little workaround to index the scan args from the list that was somehow wrapped around it

femtomc (approved):

PR #1214, opened 2024-07-10
Hotfix for TFP broadcasting in ExactDensity.

femtomc:


sritchie:
@femtomc , can you retroactively get a review on this? Some notes I'd have:

- can we add a test / tests demonstrating what this is fixing?
- please link the associated linear issue by changing the title to include GEN-XXX (I think there is one open for this)
- `logpdf` is still broken, right?
- can you add a description to the PR so that it's easy to write release notes?

PR #1215, opened 2024-07-11
cookbook entry on "Nested approximate marginalisation & RAVI stacks" (GEN-134)

MathieuHuot:


MathieuHuot:
@arijitnoobstar  already gave great comments about how to improve a couple of cookbook entries including this one but I'll need to think about how to do that. As this one now works, I suggest dealing with these revisions in another PR.

sritchie (approved):

PR #1217, opened 2024-07-12
Clean up easy pyright errors

sritchie:
This PR:

- upgrades pyright in the pre-commit hook to the latest version (1.1.371)
- Changes all bare uses of `Callable` to `Callable[..., Any]`... this is not ideal, but we can do another pass knocking out any unfounded use of `Any` in the codebase
- Changes `get_sample` calls to `get_choices` where the caller is assuming that the return value is a choicemap. I know that @femtomc wants to deprecate `get_choices` calls, but the types at play make it clear that we're not ready for this yet.

I want to get to a place where pyright failing for a PR is a notable event, not something we ignore.

Locally, this takes me from 173 errors in pre-commit to 124 errors.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,6 +37,6 @@ repos:
         types_or: [ python, pyi, jupyter ]
 
   - repo: https://github.com/RobertCraigie/pyright-python
-    rev: v1.1.370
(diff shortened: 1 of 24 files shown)


littleredcomputer (approved):
LGTM

PR #1228, opened 2024-07-17
Adding resampling strategies to the compositional SMC library (GEN-382)

MathieuHuot:
Added 4 resampling strategies to the SMC library.
These are: multinomial, systematic, stratified, residual.
They extend an abstract class `ESSResamplingStrategy` which extends `SMCAlgorithm`.

They can be used as follows (code in tests):
```

@genjax.gen
def model():
      _ = genjax.flip(0.5) @ "x"
      _ = genjax.flip(0.7) @ "y"

target = Target(model, (), C["y"].set(True))
alg = ImportanceK(target, k_particles=100)
multinomial_resample = MultinomialResampling(self.alg, 20, ess_threshold=1.0)
multinomial_resample.run_smc(self.key)
```

The 4 strategies run but they definitely need more testing and polish. There's some TODOs in the file `smc.py` that explains what need to be checked or done. They also currently don't implement cSMC.


sritchie:
@MathieuHuot @femtomc what's the status on these, guys? Do we want to get these in after @femtomc 's new push on edit? Otherwise can we merge `main` and get a review in?

MathieuHuot:
> @MathieuHuot @femtomc what's the status on these, guys? Do we want to get these in after @femtomc 's new push on edit? Otherwise can we merge `main` and get a review in?

I'd like a version of these on main if possible, but @femtomc I don't know if we want to have them available through `alg` as I wrote, have them as separate functions and let people figure out how/when to use it, or try to include them as part of `edit` somehow? I think I remember you mentioning an issue with the integration with `edit` that these resampling strategies don't work in general, they need a particle collection. But in light of the new developments/refactor do you have any new ideas/insights regarding this issue?

sritchie:
@MathieuHuot , ping on this — I know we want these in `main` but how should we proceed?

MathieuHuot:
Thanks for the ping. This was blocked by a few things.
- we needed to clarify the batching semantics for GenJAX ✅
- we needed to define what's a batched trace and what its guarantees are ✅
- we needed to make progress on clearing the meaning of `edit` and its mathematical guarantees (somewhat ✅)
- the general version requires thinking about batched traces of different batch-dimension before and after resampling. This is tricky to reason about and we have not solved this ❌
- the fallback would be to implement the N particles to N particles resampling only (or maybe for some leading axes given by the user). This seems doable but has not been solved in this draft ❌

This is not the most urgent on the list of todos (people use a very simple categorical resampling by hand for now and that's doing ok) so I think I'll close this for now.

PR #1233, opened 2024-07-17
Replace all "assert True" statements with simple tests

### LINKED ISSUE #1183: [BUG] `assert True` tests need to be replaced with proper tests
**Describe the bug!**

 This exists in a number of places in testing, and we should replace these with actual tests.

sritchie:
Closes #1183.

### DIFF
--- a/tests/generative_functions/test_switch_combinator.py
+++ b/tests/generative_functions/test_switch_combinator.py
@@ -36,9 +36,8 @@ class TestSwitchCombinator:
 
         key = jax.random.PRNGKey(314159)
         key, sub_key = jax.random.split(key)
-        _tr = model.simulate(sub_key, ())
-        # TODO
-        assert True
+        tr = model.simulate(sub_key, ())
+        assert 0.5672885 == tr.get_retval()
 
     def test_switch_combinator_simulate(self):
         @genjax.gen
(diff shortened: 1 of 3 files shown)


tests/generative_functions/test_switch_combinator.py
```diff
@@ -244,6 +243,5 @@ def model():
 
         key = jax.random.PRNGKey(314159)
         key, sub_key = jax.random.split(key)
-        _tr = model.simulate(sub_key, ())
-        # TODO
-        assert True
+        tr = model.simulate(sub_key, ())
+        assert 0.0 == tr.get_retval()
```
littleredcomputer, line 247: I think in this case you should also assert that `tr.get_choices()['b']` is truthy, since that is the antecedent to the retval being exactly zero, and also assert on the score

tests/generative_functions/test_switch_combinator.py
```diff
@@ -36,9 +36,8 @@ def model():
 
         key = jax.random.PRNGKey(314159)
         key, sub_key = jax.random.split(key)
-        _tr = model.simulate(sub_key, ())
-        # TODO
-        assert True
+        tr = model.simulate(sub_key, ())
```
littleredcomputer, line 39: The assertion we might like to make here is that a simulation of this N times always gives the same answer even though the flips could be different? Although if you're just testing that nothing died, this is enough

PR #1234, opened 2024-07-18
Support masks in ChoiceMapBuilder

SiegeLordEx:
I don't know if this is a good idea or a correct way to approach this. I just wanted to use a ChoiceMapBuilder for a `genjax.mask` and this seemed a good way to do it.

### DIFF
--- a/src/genjax/_src/core/generative/choice_map.py
+++ b/src/genjax/_src/core/generative/choice_map.py
@@ -54,7 +54,8 @@ register_exclusion(__file__)
 #################
 
 StaticAddressComponent = String
-DynamicAddressComponent = Int | IntArray
+MaskAddressComponent = Bool | BoolArray
+DynamicAddressComponent = Int | IntArray | MaskAddressComponent
 AddressComponent = StaticAddressComponent | DynamicAddressComponent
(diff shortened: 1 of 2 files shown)


femtomc:
Thanks for the PR! This is the right idea, but has a few issues -- and we're in the middle of a sprint for other issues, can we discuss next week?

The main one is that the notion of the mask flag is not an address, per say -- it's just a bit of metadata wrapped around a choice map.

sritchie:
Hey @SiegeLordEx , I have my head around ChoiceMapBuilder and would love to know more about what you're trying to do here. @littleredcomputer and I are looking at this next week.

sritchie:
@SiegeLordEx , I think for the case you have in the tests I would write

```python
C["a"].set(C.v(3.0).mask(jnp.array(True)))
```

or

```python
C.v(3.0).mask(jnp.array(True)).extend("a")
```

after #1323 gets merged. I see what you're thinking here... I agree with @femtomc that the syntax proposed here is confusing the idea of a mask with an address. I think we want the syntax in `[]` to become numpy's slicing syntax over time.

I'll close, but please re-open if I'm off base here and we can keep discussing.

PR #1239, opened 2024-07-22
Get serious about `UpdateRequest` (GEN-373)

femtomc:
Key files to look at for changes:

* `src/genjax/_src/core/generative/core.py` (changes to the core datatypes, the way the interfaces are organized, and overloads which keep the common usage of choice map with the interfaces observationally unchanges)
* `src/genjax/_src/generative_functions/distributions/distribution.py` (changes to `Distribution` which uses the new interface organization)
* ... similar for other generative function definition files.

## TODO
- [ ] address the pyright directive in distribution.py, line 438; make sure that the newly constrained `update_masked` issue is handled by the new edit stuff


src/genjax/_src/adev/core.py
```diff
@@ -41,7 +41,7 @@
     Is,
     List,
     PRNGKey,
-    Tuple,
+    tuple,
```
sritchie: you can delete this from `typing` and remove the import everywhere, since `tuple` is in python stdlb

src/genjax/__init__.py
```diff
@@ -11,15 +11,19 @@
...
+"""GenJAX is a probabilistic programming system constructed by combining the
+concepts of Gen with the program transformation and hardware accelerator
+compilation capabilities of JAX."""
 
 # This __init__ file exports GenJAX's public API.
 # For the internals, see _src.
 
 from importlib import metadata
 
+from beartype.claw import beartype_this_package
```
sritchie, line 23: pull into its own PR - adding this and removing all `typecheck` annotations.

pyproject.toml
```diff
@@ -154,10 +171,10 @@ convention = "google"
 
 [tool.ruff.lint]
 preview = true
-extend-select = ["I"]
+extend-select = ["I", "RUF"]
```
sritchie: pull into its own PR

src/genjax/_src/adev/core.py
```diff
@@ -41,52 +41,55 @@
...
+
+    These objects support forward sampling, but also come equipped with a
+    strategy that interacts with ADEV's AD transformation to return a JVP
+    estimate.
+
     """
 
     @abstractmethod
-    def sample(self, key, *args):
+    def sample(self, key, *arguments):
```
sritchie: that's fine but extract into your "nonsense additions" pr

src/genjax/_src/core/pytree.py
```diff
@@ -110,31 +115,36 @@ def dataclass(
...
 
 
             print(MyClass(10, jnp.array(5.0)).render_html())
             ```
+
         """
 
         return pz.pytree_dataclass(
             incoming,
+            overwrite_parent_init=True,
```
sritchie, line 129: this also goes into the beartype PR, whoops

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -20,76 +20,126 @@
...
     """
 
-    flag: BoolArray
-    value: Any
+    flag: Bool | BoolArray
+    value: V
 
     @classmethod
-    def maybe(cls, f: BoolArray, v: Any):
+    def maybe(cls, f: Bool | BoolArray, v: V) -> V | "Masked[V]":
```
sritchie: I like that these are more permissive — I am NOT a huge fan of the `|` on the exit without better typing. Can we either

- always return a `Masked`, or
- add an `@overload` annotation?

sritchie: I also want to generate docstrings for all of these. Cursor is very good at this, I am happy to do this with you as a final step and get these in.

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -20,76 +20,126 @@
...
             case _:
-                return Mask(f, v)
+                if staged_check(f):
+                    return v
+                else:
+                    return Masked(f, v)
 
     @classmethod
-    def maybe_none(cls, f: BoolArray, v: Any):
+    def maybe_none(cls, f: Bool | BoolArray, v: Any):
```
sritchie: let's do a good type dance here too...

src/genjax/_src/core/generative/functional_types.py
sritchie: OR we can make a new named type defined as `V | Masked[V]`, vs adding this explicitly everywhere

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -105,7 +155,7 @@ def _check():
         optional_check(_check)
         return self.value
 
-    def unsafe_unmask(self):
+    def unsafe_unmask(self) -> V:
```
sritchie, line 155: if they both throw, why would we use this one?

if you want both, can you move this above `unmask` and call it instead of `self.value` in `unmask`?

src/genjax/_src/core/interpreters/forward.py
```diff
@@ -99,9 +95,9 @@ class InitialStylePrimitive(FlatPrimitive):
     def __init__(self, name, batch_semantics=None):
         super().__init__(name)
 
-        def fun_impl(*args, **params):
-            consts, args = jax_util.split_list(args, [params["num_consts"]])
-            return jc.eval_jaxpr(params["_jaxpr"], consts, *args)
+        def fun_impl(*arguments, **params):
```
sritchie: change `args` to `arguments` could be its own PR to tidy this up

src/genjax/_src/core/interpreters/incremental.py
```diff
@@ -41,15 +43,19 @@
...
+    Generic,
     IntArray,
     List,
     Optional,
-    Tuple,
+    TypeVar,
     Value,
     static_check_is_concrete,
-    typecheck,
+    tuple,
```
sritchie: no need for `tuple` export

src/genjax/_src/core/pytree.py
```diff
@@ -198,8 +212,8 @@ def _inner(v):
         )
 
     @staticmethod
-    def partial(*args):
-        return lambda fn: Closure(args, fn)
+    def partial(*arguments):
```
sritchie: you could also ditch this change, I think everyone knows what args is?

src/genjax/_src/core/serialization/backend.py
```diff
@@ -16,37 +16,39 @@
...
+
     """
 
     @abc.abstractmethod
     def serialize(self, tr: Trace):
         pass
 
     @abc.abstractmethod
-    def deserialize(self, bytes, gen_fn: GenerativeFunction, args: Tuple):
+    def deserialize(self, bytes, gen_fn: GenerativeFunction, arguments: tuple):
```
sritchie: honestly I would revert the `args` / `arguments` flip...

docs/library/core.md
```diff
@@ -60,25 +56,23 @@ The most primitive way to encode _runtime uncertainty_ about a piece of data is
...
 
 ::: genjax.core.Mask
-    options:
-        show_root_heading: true
-        members:
-          - unmask
-          - match
+options:
+show_root_heading: true
+members: - unmask - match
```
sritchie: This is a weird unrolling of the list syntax, and it looks like we lost indentation... can we revert those chunks?

src/genjax/__init__.py
sritchie: (I did this in #1266 , so we can delete this diff here)

src/genjax/_src/adev/core.py
```diff
@@ -40,25 +40,27 @@
...
 )
 
 DualTree = Annotated[
     Any,
     Is[lambda v: Dual.static_check_dual_tree(v)],
 ]
-"""
-`DualTree` is the type of `Pytree` argument values with `Dual` leaves.
-"""
+"""`DualTree` is the type of `Pytree` argument values with `Dual` leaves."""
```
sritchie, line 49: is it possible to revert all of these formatting changes? They're distracting and if easy via vscode or something I bet we could toss them out.

src/genjax/_src/core/generative/__init__.py
```diff
@@ -59,35 +72,49 @@
...
+    "ProjectedChoiceMap",
+    "Projection",
     "Retdiff",
     "Retval",
     "Sample",
+    "SampleCoercableToChoiceMap",
     "Score",
     "Selection",
     "SelectionBuilder",
+    "SelectionProjection",
```
sritchie, line 112: types, types, types!

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -20,75 +20,123 @@
...
+    Callable,
+    Generic,
     IntArray,
+    TypeVar,
     static_check_bool,
     static_check_is_concrete,
-    typecheck,
 )
 
+V = TypeVar("V")
```
sritchie, line 37: add a comment about what these typevars are

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -20,75 +20,123 @@
...
     ## Usage of invalid data
 
-    If you use invalid `Mask(False, data)` data in inference computations, you may encounter silently incorrect results.
+    If you use invalid `Masked(False, data)` data in inference computations, you may encounter silently incorrect results.
+
     """
 
-    flag: BoolArray
-    value: Any
+    flag: bool | BoolArray
```
sritchie, line 68: wdyt of making this ONLY a `BoolArray`, and then converting it on the way in, via `maybe`? I think we talked about this...

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -20,75 +20,123 @@
...
             case _:
-                return Mask(f, v)
+                if staged_check(f):
+                    return v
+                else:
+                    return Masked(f, v)
 
     @classmethod
-    def maybe_none(cls, f: BoolArray, v: Any):
+    def maybe_none(cls, f: bool | BoolArray, v: Any):
```
sritchie, line 83: this should take a `v: V`, yeah?

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -20,75 +20,123 @@
...
             else None
             if static_check_bool(f)
-            else Mask.maybe(f, v)
+            else Masked.maybe(f, v)
         )
 
+    @classmethod
+    def lift(
+        cls,
+        f: Callable[[V], R],
```
sritchie, line 97: if we are going to have these `|` type signatures, I think we need to `@overload` to show that it's actually two possible signatures:

`V => R`

or

`Masked[V] => Masked[R]`

not four possible signatures, right?

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -20,75 +20,123 @@
...
+            match v:
+                case Masked(flag, value):
+                    return Masked(flag, f(value))
+                case _:
+                    return f(v)
+
+        return wrapped
+
+    @classmethod
+    def lift2(
```
sritchie, line 109: how about this, how many actual possible signatures do we have?

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -104,21 +152,22 @@ def _check():
...
-    def unsafe_unmask(self):
+    def unsafe_unmask(self) -> V:
         # Unsafe version of unmask -- should only be used internally,
         # or carefully.
         return self.value
 
 
 @Pytree.dataclass(match_args=True)
-class Sum(Pytree):
+class DynamicEnum(Pytree):
```
sritchie, line 162: what is dynamic about it?

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -174,44 +223,38 @@ def uncertain_idx(idx):
     """
 
     idx: IntArray | Diff
```
sritchie, line 225: this feels suss... what's the meaning of a `DynamicEnum` with a diff?

src/genjax/_src/core/interpreters/__init__.py
```diff
@@ -11,12 +11,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""This module contains definitions for interpreters which act as program transformers
-when staged out by JAX.
+"""This module contains definitions for interpreters which act as program
```
sritchie, line 14: revert docs formatting, I can't tell if there are new docs here or not

src/genjax/_src/core/interpreters/incremental.py
```diff
@@ -39,15 +41,18 @@
...
+    Generic,
     IntArray,
     Optional,
+    TypeVar,
     Value,
     static_check_is_concrete,
-    typecheck,
 )
 
+V = TypeVar("V")
```
sritchie, line 53: docs on what these are, please

src/genjax/_src/core/interpreters/incremental.py
```diff
@@ -215,7 +247,7 @@ def _inner(v):
             if static_check_is_change_tangent(v):
```
sritchie, line 247: could do

```
return static_check_is_change_tangent(v) and isinstance(v, _NoChange)
```

src/genjax/_src/core/interpreters/staging.py
```diff
@@ -38,25 +39,63 @@ def staged_check(v):
     return static_check_is_concrete(v) and v
 
 
-def staged_and(x, y):
+def staged_and(
```
sritchie, line 42: can you name this `bool | BoolArray` type if we are going to use it all over?

src/genjax/_src/core/interpreters/staging.py
```diff
@@ -38,25 +39,63 @@ def staged_check(v):
...
+
+@overload
+def staged_or(
+    x: bool,
+    y: BoolArray,
+) -> BoolArray:
+    pass
+
+
+@overload
```
sritchie, line 81: maybe you could kill this fourth one with

```python
@overload
def staged_or(
    x: bool,
    y: bool,
) -> bool:
    pass


@overload
def staged_or(
    x: bool | BoolArray,
    y: bool,
) -> BoolArray:
    pass


@overload
def staged_or(
    x: bool,
    y: bool | BoolArray,
) -> BoolArray:
    pass
```

src/genjax/_src/core/pytree.py
```diff
@@ -109,31 +113,36 @@ def dataclass(
 
 
             @Pytree.dataclass
-            @typecheck  # Enforces type annotations on instantiation.
+            # Enforces type annotations on instantiation.
```
sritchie, line 116: no more need for this comment

src/genjax/_src/core/pytree.py
```diff
@@ -109,31 +113,36 @@ def dataclass(
...
+        `default_factory`, to customize how the field is instantiated when an
+        instance of the dataclass is instantiated.` Fields which are provided
+        with default values must come after required fields in the dataclass
+        declaration.
 
         Examples:
             ```python exec="yes" html="true" source="material-block" session="core"
             @Pytree.dataclass
-            @typecheck  # Enforces type annotations on instantiation.
+            # Enforces type annotations on instantiation.
```
sritchie, line 145: remove comment

src/genjax/_src/generative_functions/combinators/dimap.py
```diff
@@ -106,101 +123,94 @@ def transformed_normal_draw(mean, std):
         )
         print(tr.render_html())
         ```
+
     """
 
-    inner: GenerativeFunction
-    argument_mapping: Callable[[tuple], ArgTuple] = Pytree.static()
-    retval_mapping: Callable[[ArgTuple, R], S] = Pytree.static()
+    inner: GenerativeFunction[A_, S, R_, C, P, U]
```
sritchie, line 129: nice!!

src/genjax/_src/generative_functions/combinators/dimap.py
```diff
@@ -106,101 +123,94 @@ def transformed_normal_draw(mean, std):
...
+        inner_tr, w, bwd_projection = self.inner.importance_edit(
+            key, constraint, inner_args
+        )
+        inner_retval = inner_tr.get_retval()
+        retval = self.retval_mapping(inner_args, inner_retval)
+        trace = DimapTrace(
+            self,
+            inner_tr,
+            args,
+            retval,
```
sritchie, line 172: I don't feel capable of giving good reviews of the math motivating any of these!

src/genjax/_src/generative_functions/combinators/dimap.py
```diff
@@ -238,7 +249,7 @@ def pre_process(x, y):
             return (x + 1, y * 2)
 
 
-        def post_process(args, retval):
+        def post_process(arguments, retval):
```
sritchie, line 252: why switch "args" for "arguments" everywhere?

src/genjax/_src/generative_functions/combinators/dimap.py
```diff
@@ -210,12 +220,13 @@ def assess(
 
 def dimap(
     *,
-    pre: Callable[..., ArgTuple] = lambda *args: args,
-    post: Callable[[ArgTuple, R], S] = lambda _, retval: retval,
+    pre: Callable[..., A] = lambda *args: args,
+    post: Callable[[A, R_], R] = lambda _, retval: retval,
     info: String | None = None,
 ) -> Callable[[GenerativeFunction], GenerativeFunction]:
```
sritchie, line 226: how about adding the type parameters back on to the input and returned generative function? I think libraries should be as explicit as possible with these things

src/genjax/_src/generative_functions/combinators/dimap.py
```diff
@@ -303,21 +315,22 @@ def map_model(x):
...
 
     return dimap(pre=lambda *args: args, post=post, info=info)
 
 
 def contramap(
-    f: Callable[..., ArgTuple],
+    f: Callable[..., A],
     *,
     info: String | None = None,
 ) -> Callable[[GenerativeFunction], GenerativeFunction]:
```
sritchie, line 331: Now that we have the type params can we add them here?

src/genjax/_src/generative_functions/combinators/dimap.py
```diff
@@ -303,21 +315,22 @@ def map_model(x):
...
-    def post(_, x: R) -> S:
+    def post(_, x: R_) -> R:
         return f(x)
 
     return dimap(pre=lambda *args: args, post=post, info=info)
 
 
 def contramap(
-    f: Callable[..., ArgTuple],
+    f: Callable[..., A],
```
sritchie, line 328: don't we want `Callable[[A_], A]` or something like that?

src/genjax/_src/generative_functions/combinators/repeat.py
```diff
@@ -17,21 +17,18 @@
...
-    typecheck,
 )
 
-register_exclusion(__file__)
 
-
-def RepeatCombinator(gen_fn: GenerativeFunction, /, *, n: Int) -> GenerativeFunction:
-    """
-    A combinator that samples from a supplied [`genjax.GenerativeFunction`][] `gen_fn` a fixed number of times, returning a vector of `n` results.
+def RepeatCombinator(gen_fn: GenerativeFunction, /, *, n: int) -> GenerativeFunction:
```
sritchie, line 25: return type parameters,

src/genjax/_src/generative_functions/combinators/repeat.py
```diff
@@ -40,10 +37,12 @@ def RepeatCombinator(gen_fn: GenerativeFunction, /, *, n: Int) -> GenerativeFunc
     )
 
 
-@typecheck
-def repeat(*, n: Int) -> Callable[[GenerativeFunction], GenerativeFunction]:
-    """
-    Returns a decorator that wraps a [`genjax.GenerativeFunction`][] `gen_fn` of type `a -> b` and returns a new `GenerativeFunction` of type `a -> [b]` that samples from `gen_fn `n` times, returning a vector of `n` results.
+def repeat(*, n: int) -> Callable[[GenerativeFunction], GenerativeFunction]:
```
sritchie, line 40: type params on return value?

src/genjax/_src/inference/smc.py
```diff
@@ -239,18 +238,20 @@ def estimate_reciprocal_normalizing_constant(
...
     This initializes a 1-particle `ParticleCollection` by importance sampling from `target` using `q`.
 
     Any choices in `target.p` not in `q` will be sampled from the internal proposal distribution of `p`,
     given `target.constraints` and the choices sampled by `q`.
+
     """
 
     target: Target
-    q: SampleDistribution | None = Pytree.field(default=None)
+    q: Optional[SampleDistribution] = Pytree.field(default=None)
```
sritchie, line 254: I think Python likes `| None` now, I'm fine with this switch but | None is apparently "idiomatic"

src/genjax/_src/inference/smc.py
```diff
@@ -642,14 +642,11 @@ def get_gen_fn(self) -> GenerativeFunction:
 
 
 @Pytree.dataclass
-@typecheck
 class AttachCombinator(GenerativeFunction):
     gen_fn: GenerativeFunction
     importance_move: SMCMove = Pytree.static(default=DeferToInternal())
     update_move: SMCMove = Pytree.static(default=DeferToInternal())
 
-    @GenerativeFunction.gfi_boundary
```
sritchie, line 651: what were these again? why can they go away?

src/genjax/_src/inference/sp.py
```diff
@@ -217,45 +256,43 @@ def estimate_reciprocal_normalizing_constant(
...
             )
 
-            return (Z, latent_choices)
+            return (Z, ChoiceMapSample(latent_choices))
 
-    @typecheck
     def estimate_logpdf(
         self,
         key: PRNGKey,
-        constraint: ChoiceMap,
```
sritchie, line 258: args => arguments, but constraint to `v`??

src/genjax/_src/inference/vi.py
```diff
@@ -86,7 +85,11 @@ def logpdf(v: Any, *args: Any) -> FloatArray:
 
 
 def logpdf(gen_fn):
-    return lambda v, *args: gen_fn.assess(ChoiceMap.value(v), args)[0]
+    return lambda v, *args: gen_fn.assess(
+        jax.random.PRNGKey(0),
```
sritchie, line 89: THIS does not look right!! Can we replace this with a proper key? Do we need to take one in as an arg?

src/genjax/_src/inference/vi.py
```diff
@@ -134,19 +137,16 @@ def logpdf(gen_fn):
...
-"""
+"""The type of gradient estimates returned by sampling from gradient estimators
+for loss terms."""
 
 
-@typecheck
 def ELBO(
     guide: SampleDistribution,
     make_target: Callable[..., Target],
 ) -> Callable[[PRNGKey, Arguments], GradientEstimate]:
```
sritchie, line 147: wrong return type now, yeah? this returns value and gradient estimate?

src/genjax/_src/inference/vi.py
```diff
@@ -134,19 +137,16 @@ def logpdf(gen_fn):
...
-    """
-    Return a function that computes the gradient estimate of the ELBO loss term.
-    """
+    """Return a function that computes the value and gradient estimate of the
+    ELBO loss term."""
 
     def grad_estimate(
```
sritchie, line 151: name needs to change too

tests/generative_functions/test_scan_combinator.py
```diff
@@ -18,301 +18,16 @@
...
+    ChoiceMapEditRequest,
+    IndexEditRequest,
+    SelectionRegenerateRequest,
+)
 from genjax import SelectionBuilder as S
-from genjax import UpdateProblemBuilder as U
 from genjax.typing import FloatArray
 
 
-@genjax.iterate(n=10)
```
sritchie, line 27: let's get all these tests back in, this was probably a mistake

tests/generative_functions/test_scanlike_combinators.py
```diff
@@ -0,0 +1,299 @@
...
+        @genjax.iterate(n=10)
+        @genjax.gen
+        def scanner(x):
+            z = genjax.normal(x, 1.0) @ "z"
+            return z
+
+        key = jax.random.PRNGKey(314159)
+        key, sub_key = jax.random.split(key)
+        for i in range(1, 5):
+            tr, _w = jax.jit(scanner.importance)(sub_key, C[i, "z"].set(0.5), (0.01,))
```
sritchie, line 69: never mind I see that they're moved!

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -445,20 +432,59 @@ def a(
...
+    def at(self) -> "ChoiceMapAddressIndex":
+        return self
+
+    def filter(self):
+        sels = map(lambda addr: SelectionBuilder[addr], self.addrs)
+        or_sel = reduce(or_, sels)
+        return self.choice_map.filter(or_sel)
+
+
+class ChoiceMap(Generic[V], Pytree):
```
sritchie, line 487: `V` makes sense for valuechoicemap but not for a nested one... I still feel like this mashup of get_value and has_value into here is odd.

src/genjax/_src/core/generative/core.py
```diff
@@ -994,16 +955,22 @@ def add_and_square_step(sum, x):
...
             print(tr.render_html())
             ```
+
         """
         import genjax
 
-        return genjax.scan(n=n, reverse=reverse, unroll=unroll)(self)
+        return genjax.scan(
+            n=n,
+            reverse=reverse,
```
sritchie, line 967: this reverse flag is still busted, yeah?

src/genjax/_src/core/generative/core.py
```diff
@@ -1261,14 +1235,15 @@ def normal_draw(mean):
             )
             print(tr.render_html())
             ```
+
         """
         import genjax
 
         return genjax.mask(self)
 
     def or_else(self, gen_fn: "GenerativeFunction", /) -> "GenerativeFunction":
```
sritchie, line 1247: type parameters?

src/genjax/_src/core/generative/core.py
```diff
@@ -1588,47 +1573,547 @@ def wrapped(*args):
...
+    def complement(self) -> "SelectionProjection":
+        return SelectionProjection(~self.selection)
+
+
+#######################
+# Edit specifications #
+#######################
+
+
+class EditRequest(Pytree):
```
sritchie, line 1753: is there some other natural place to put these vs core?

src/genjax/_src/generative_functions/combinators/scan.py
```diff
@@ -191,41 +232,48 @@ def random_walk(prev, _):
...
+        tuple[Ca, Sc2],
+        ChoiceMapConstraint,
+        SelectionProjection | ChoiceMapProjection,
+        ChoiceMapEditRequest,
+    ]
 
     # Only required for `None` carry inputs
-    length: Optional[Int] = Pytree.static()
+    length: Optional[int] = Pytree.static()
     reverse: bool = Pytree.static(default=False)
```
sritchie, line 249: my busted reverse lives on

src/genjax/_src/generative_functions/static.py
```diff
@@ -21,56 +21,69 @@
...
     PRNGKey,
-    typecheck,
+    TypeVar,
+    overload,
 )
 
-register_exclusion(__file__)
+A = TypeVar("A", bound=Arguments)
+S = TypeVar("S", bound=Sample)
+R = TypeVar("R", bound=Retval)
```
sritchie, line 73: the `Retval` bound of `Any` I think can go away, yeah? and the `Retval` type? Maybe there are some others like this too that can go away.

sritchie (commented):
First crop of comments

src/genjax/_src/generative_functions/combinators/scan.py
femtomc: We can fix it in another PR -- we need to make sure the logic of the visitation order is propagated to getting constraints and edit requests for some of the interfaces.

src/genjax/__init__.py
femtomc: It's annoying that I did this here -- but there's a test case which won't pass unless I either add this line, or restore all `typecheck` annotations, which I removed like a dumb dumb

Can we funnel this through despite this?

femtomc: I can also add a single typecheck back to the method which throws for the test case

src/genjax/__init__.py
```diff
@@ -20,6 +20,10 @@
...
+
+beartype_this_package()
+
```
littleredcomputer, line 26: Are we confident enough in our test suite to do this? I guess the thing to do would be to get George, Pavel & co. to try this before we blast it into main. My concern is that this will enable a lot of new failure cases that our test suite may not provoke

src/genjax/_src/adev/core.py
```diff
@@ -263,12 +265,12 @@ def static_check_dual_tree(v):
 
 @Pytree.dataclass
 class ADInterpreter(Pytree):
-    """The `ADInterpreter` takes a `Jaxpr`,
-    propagates dual numbers through it, while also performing a CPS transformation,
-    to compute forward mode AD.
+    """The `ADInterpreter` takes a `Jaxpr`, propagates dual numbers through it,
+    while also performing a CPS transformation, to compute forward mode AD.
```
littleredcomputer, line 270: I would really love to see an example of what this means _concretely_, something like $x + y \mapsto k(x + y, \delta x + \delta y)$

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -403,9 +395,8 @@ class ChoiceMapNoValueAtAddress(Exception):
 
 @Pytree.dataclass
 class _ChoiceMapBuilder(Pytree):
-    addr: ExtendedAddress = ()
+    addr: Optional[ExtendedAddress] = Pytree.field(default=None)
```
littleredcomputer, line 398: In my installation I am always advised that Optional is deprecated in favor of `T | None`, making me wonder if we need to nail down the type checking parameters for developer

sritchie:
No problem, keep it!

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -445,20 +432,59 @@ def a(
...
 
 
-class ChoiceMap(Sample, Constraint):
+##########################
+# AddressIndex interface #
+##########################
+
+
+@Pytree.dataclass
+class ChoiceMapAddressIndex(Pytree):
```
littleredcomputer, line 460: In what sense is the word Index used here? Like an index into an array? Or is it an index of ChoiceMapAddresses?

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -445,20 +432,59 @@ def a(
...
+    choice_map: "ChoiceMap"
+    addrs: list[Address]
+
+    def __getitem__(self, addr: AddressComponent | Address) -> "ChoiceMapAddressIndex":
```
littleredcomputer, line 464: I would think of getitem as returning a member of this collection but it looks like it is returning an augmented object? Maybe this would be better written `append`

littleredcomputer: Maybe this is for a kind of selector syntax (comment helpful here)

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -445,20 +432,59 @@ def a(
...
+        addr = addr if isinstance(addr, tuple) else (addr,)
+        return ChoiceMapAddressIndex(
+            self.choice_map,
+            [*self.addrs, addr],
+        )
+
+    def set(self, v):
+        new = self.choice_map
+        for addr in self.addrs:
+            new = ChoiceMapBuilder.a(addr, v) + new
```
littleredcomputer, line 474: Do we get back an object of the same type? It looks like "not", since we used to have a list of addresses, and now we have fanned them out?

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -445,20 +432,59 @@ def a(
...
+        return new
+
+    @property
+    def at(self) -> "ChoiceMapAddressIndex":
```
littleredcomputer, line 478: when is this helpful?

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -445,20 +432,59 @@ def a(
...
+        return self
+
+    def filter(self):
```
littleredcomputer, line 481: does this convert the choice into a filter that involves the same addresses? I'm confused, since we than apply the filter to "ourselves"

I get the impression that "set" and "filter" are not things that could chain together, because it's hard to see what the type effects of these operations are. As we scan along, we get CMs of various types

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -635,53 +658,20 @@ def idx(cls, addr: DynamicAddressComponent, v: Any) -> "ChoiceMap":
...
-            return self.choice_map.filter(or_sel)
+    def kwargs(cls, **kwargs) -> "ChoiceMap":
+        return ChoiceMap.dict(kwargs)
 
     @property
-    def at(self) -> AddressIndex:
-        """Access the `ChoiceMap.AddressIndex` mutation interface. This allows
+    def at(self) -> ChoiceMapAddressIndex:
+        """Access the `ChoiceMapAddressIndex` mutation interface. This allows
         users to take an existing choice map, and mutate it _functionally_.
```
littleredcomputer, line 675: I am beginning to wonder whether the Builder pattern wouldn't save us a _lot_ of headache here (the idea borrowed from [Protocol Buffers, which are immutable in Java](https://protobuf.dev/getting-started/javatutorial/#builders-messages)).  In the builder state, you can do anything, but you have to call build to bake the immutable object. To modify one you already have, you call `.toBuilder()`. to restore mutability. That of course involves a lot of copying, which JAX won't care about at tracing time, but I worry that what we have here will be O(N^2) on the client pre-compilation side since each mutation means we must copy every array. 

Whereas, `.build() / .toBuilder()` puts the phase transition in the users' hands, and also lets them work in a mutable style when that is useful. 

In this idea, the mutable form could just be plain ol' python objects (dicts etc)

littleredcomputer: This could relieve choice map objects from any responsibility to support their own assembly/disassembly at lower levels. I'm a big fan of immutability but I don't like hidden copy-on-write as much

src/genjax/_src/core/generative/core.py
```diff
@@ -20,296 +20,301 @@
...
-"""
-`Retval` is the type of return values from the return value function of a generative function. It is a type alias for `Any`, and is used to improve readability and parsing of interface specifications.
+"""`Retval` is the type of return values from the return value function of a
+generative function.
+
+It is a type alias for `Any`, and is used to improve
+readability and parsing of interface specifications.
 """
 
 Argdiffs = Annotated[
```
littleredcomputer, line 110: Is there not a Diff type we can use here, straight up? Argdiffs and Retdiffs seem to be the exact same type, with a runtime typecheck for Diff. How about just using a Diff type? Or is it a Pytree of Diffs?

src/genjax/_src/core/generative/core.py
```diff
@@ -20,296 +20,301 @@
...
-    """
-    An `UpdateProblem` is a request to update a trace of a generative function. Generative functions respond to instances of subtypes of `UpdateProblem` by providing an [`update`][genjax.core.GenerativeFunction.update] implementation.
+###########
+# Samples #
+###########
 
-    Updating a trace is a common operation in inference processes, but naively mutating the trace will invalidate the mathematical invariants that Gen retains. `UpdateProblem` instances denote requests for _SMC moves_ in the framework of [SMCP3](https://proceedings.mlr.press/v206/lew23a.html), which preserve these invariants.
-    """
 
+class Sample(Generic[C], Pytree):
```
littleredcomputer, line 139: So a `Sample[C]` is something that can be converted to type `C` if we call `to_constraint()` on it, and we know nothing else about it.

Q: Since nothing can be done with this object other than convert it to type C, why not just use the type C directly?

I have surveyed the four implementers of Sample. Each of them implements to_constraint by sending all of their instance variables to the constructor for type C. I claim then that this type is just an alias for C, with a delayed constructor. But no rationale for this is apparent.

src/genjax/_src/core/generative/core.py
```diff
@@ -20,296 +20,301 @@
...
-    An IntervalConstraint encodes the constraint that the value output by a
-    distribution on the reals lies within a given interval.
+Pj = TypeVar("Pj", bound="Projection")
 
-    Formally, `IntervalConstraint(a, b)` represents the constraint (`x` $\\mapsto$ `a` $\\leq$ `x` $\\leq$ `b`, `True`).
-    """
 
-    a: FloatArray
-    b: FloatArray
+class Projection(Generic[Pj, S1, S2], Pytree):
```
littleredcomputer, line 247: How does this differ from `Callable[[S1], S2]`?  From this documentation I'd have expected methods `project :: S1 -> S2` and `unproject :: S2 -> S1` which can throw if this projection is not also injective?  

What is `Pj`? What does `complement` do?

littleredcomputer: ... reading ahead: it seems that the complement takes elements that are not mapped by the projection and sends them... where? I can understand that the complement of the IdentityProjection is empty (since everybody got sent somewhere), and that the complement of the EmptyProjection is the IdentityProjection (since nobody got sent anywhere, and we have that $S2\subset S1$, the identity would work as the complement, but in what other case is it well defined to choose a destination for the elements other than those that have been projected?

src/genjax/_src/core/generative/core.py
```diff
@@ -449,14 +468,33 @@ def get_gen_fn(self) -> "GenerativeFunction":
 #######################
 
 
-class GenerativeFunction(Pytree):
-    """
-    `GenerativeFunction` is the type of _generative functions_, the main computational object in Gen.
+class GenerativeFunction(
```
littleredcomputer, line 471: I've scanned ahead and it is not clear that we get much if anything out of type parameter `A`. A is definitely under the user's control anyway, it will change constantly as the user works, and I can't see how we could do much type inference with it inside our code.

src/genjax/_src/core/generative/core.py
```diff
@@ -449,14 +468,33 @@ def get_gen_fn(self) -> "GenerativeFunction":
...
+        # function's P distribution on -- formally, this type is always restricted to be the type
+        # of samples from P -- but in the code, S can sometimes be a larger set of types.
+        S,
+        # the space of return values
+        R,
+        # the space of constraints to `GenerativeFunction.importance_edit`
+        C,
+        # the space of projections returned by `GenerativeFunction.importance_edit`
+        # and the space of projections to `GenerativeFunction.project_edit`
+        P,
```
littleredcomputer, line 485: It looks like `P` and `R` overlap somehow.

One thing this PR brings out is that Update and Importance are the same thing except that Importance has no "origin" target. This makes me think that the type level is not the right place to distinguish them.

src/genjax/__init__.py
MathieuHuot: last option seems good?

src/genjax/_src/core/generative/core.py
```diff
@@ -748,73 +667,114 @@ def model():
...
-        Under the hood, creates an [`UpdateProblem`][genjax.core.UpdateProblem] which requests that the generative function respond with a move from the _empty_ trace (the only possible value for _empty_ target $\\delta_\\emptyset$) to the target induced by the generative function for constraint $C$ with arguments $a$.
-        """
-        tr, w, _, _ = self.update(
-            key,
-            EmptyTrace(self),
-            GenericProblem(Diff.unknown_change(args), ImportanceProblem(constraint)),
+    def importance(
+        self,
+        key: PRNGKey,
+        choice_map: "ChoiceMap | ChoiceMapConstraint",
```
littleredcomputer, line 709: Why advertise that we take both types, if we're going to force the thing into a ChoiceMapConstraint anyway? Just say the type we really care about, which is clearly ChoiceMapConstraint in this case.

littleredcomputer: The fact that it is even possible to just shovel any ChoiceMap into the ChoiceMapConstraint constructor makes me wonder what the material difference between the types actually is!

The declaration says

```
class ChoiceMapConstraint(Generic[C], Constraint, ChoiceMap[Constraint]):
    # TODO: Any here is to maintain backwards compatibility.
    choice_map: ChoiceMap[Sample | Constraint | Any]
``` 

But here `Any` gives the game away, `Constraint` has no behavior and is apparently a marker type (ok), and then we have Sample, which is carrying the load here (I worry that the distinction between Sample and ChoiceMap is pedantic. How can I understand the distinction _in the code_?)

src/genjax/_src/core/generative/core.py
```diff
@@ -748,73 +667,114 @@ def model():
...
-        """
+        args: A,
+    ) -> tuple[S, Score, R]:
+        """Samples a [`Sample`][genjax.core.Sample] and any untraced randomness
+        $r$ from the generative function's distribution over samples ($P$), and
+        returns the [`Score`][genjax.core.Score] of that sample under the
+        distribution, and the [`Retval`][genjax.core.Retval] of the generative
+        function's return value function $f(r, t, a)$ for the sample and
+        untraced randomness."""
         tr = self.simulate(key, args)
```
littleredcomputer, line 744: doesn't seem useful. Are people supposed to override this with custom proposals? Why not just return `tr`?

littleredcomputer, line 744: Nobody uses this and it would seem better to return `tr`, so why not just call simulate instead. If you're meant to override it, would it be difficult to return the trace in that case?

src/genjax/_src/core/generative/core.py
```diff
@@ -1168,16 +1141,18 @@ def inc(x):
...
 
     def iterate_final(
-        self, /, *, n: Int, unroll: int | bool = 1
+        self, /, *, n: int, unroll: int | bool = 1
     ) -> "GenerativeFunction":
-        """
-        Returns a decorator that wraps a [`genjax.GenerativeFunction`][] of type `a -> a` and returns a new [`genjax.GenerativeFunction`][] of type `a -> a` where
+        """Returns a decorator that wraps a [`genjax.GenerativeFunction`][] of
+        type `a -> a` and returns a new [`genjax.GenerativeFunction`][] of type
+        `a -> a` where.
```
littleredcomputer, line 1155: s/\\./:/

src/genjax/_src/core/generative/core.py
```diff
@@ -1588,47 +1576,547 @@ def wrapped(*args):
...
+# Choice map specific extensions #
+##################################
+
+
+@Pytree.dataclass(match_args=True)
+class ChoiceMapSample(Generic[S], Sample, ChoiceMap[Sample]):
+    # TODO: Any here is to maintain backwards compatibility.
+    choice_map: ChoiceMap[Sample | Any]
+
+    def to_constraint(self) -> "ChoiceMapConstraint":
```
littleredcomputer, line 1589: A ChoiceMap _is_ a constraint: could we add the marker type to the class definition?

src/genjax/_src/core/generative/core.py
```diff
@@ -1588,47 +1576,547 @@ def wrapped(*args):
...
+        return ChoiceMapConstraint(submap)
+
+
+@Pytree.dataclass
+class ProjectedChoiceMap(ChoiceMap[Sample]):
+    choice_map_projection: ChoiceMap[Projection]
+    choice_map: ChoiceMap[Sample]
+
+    def get_value(self) -> Sample:
+        leaf = self.choice_map.get_value()
```
littleredcomputer, line 1666: get_value is an interesting case type wise since it used to be specific to Value choice maps, and now it is everywhere. The default value is "effecively None". Would it make sense to make `return None` the default implementation?

It is unclear how the type variable Projection enters here. What if it where `_` ? Is it understood that Projection is a kind of sample? What does `leaf_proj.project(leaf)` do? are we walking two structures in parallel?

src/genjax/_src/core/generative/core.py
```diff
@@ -1588,47 +1576,547 @@ def wrapped(*args):
...
+        key: PRNGKey,
+        trace: EmptyTrace[G],
+        args: A,
+    ) -> tuple[
+        Trace[G, A, Sample, Retval],
+        Weight,
+        Retdiff,
+        "ChoiceMapProjectionProjectRequest",
+    ]:
+        gen_fn = trace.get_gen_fn()
```
littleredcomputer, line 1912: So initially in the model of having the implementation of `tr.update(u) being `u.update(tr)` we find that we reflect back to `tr.get_gen_fn().update(u)`, so the work is still a part of the GF. Is that because the types of edit here are primitive, in the sense that _only being_ the GF involved can help complete this computation?

src/genjax/_src/core/generative/core.py
```diff
@@ -1588,47 +1576,547 @@ def wrapped(*args):
...
+    propagate_incremental: bool = Pytree.static(default=False)
+
+    def edit(
+        self,
+        key: PRNGKey,
+        trace: Trace[G, A, S, R],
+        args: A,
+    ) -> tuple[Trace[G, A, S, R], Weight, Retdiff, "IncrementalChoiceMapEditRequest"]:
+        gen_fn = trace.get_gen_fn()
+        return gen_fn.edit(key, trace, self, args)
```
littleredcomputer, line 1950: It's hard to see how `gen_fn.edit` learns the value of `propagate_incremental`: does it peek? Or does it not matter? It looks like that value needs to make its way into `IncrementalChoiceMapEditRequestHandler`

src/genjax/_src/core/generative/core.py
```diff
@@ -1588,47 +1576,547 @@ def wrapped(*args):
...
+    projection: Selection
+
+    def edit(
+        self,
+        key: PRNGKey,
+        trace: Trace[G, A, S, R],
+        args: A,
+    ) -> tuple[Trace[G, A, S, R], Weight, Retdiff, ChoiceMapEditRequest]:
+        gen_fn = trace.get_gen_fn()
+        return gen_fn.edit(key, trace, self, args)
```
littleredcomputer, line 1964: this implementation looks entirely generic, except that there's a `projection` instance variable. Should this be the default implementation of EditRequest, and the the body of this class could be `pass`?

src/genjax/_src/core/interpreters/staging.py
littleredcomputer: Does `Bool` cover both cases?

src/genjax/_src/core/interpreters/staging.py
```diff
@@ -38,25 +39,63 @@ def staged_check(v):
...
         static_check_is_concrete(x)
         and static_check_is_concrete(y)
-        and isinstance(x, Bool)
-        and isinstance(y, Bool)
+        and isinstance(x, bool)
+        and isinstance(y, bool)
     ):
         return x and y
     else:
         return jnp.logical_and(x, y)
```
littleredcomputer, line 54: `jnp.logical_and` works just fine on naked bools. Can we just return jnp.logical_and(x,y) always? I cannot construct an example where the return value would be materially different (it will be a rank 0 bool array, of course, when it might have returned a naked bool, but this is JAX, so that's OK?)

src/genjax/_src/core/interpreters/staging.py
littleredcomputer: I heartily endorse the idea of just directly returning `jnp.logical_or(x, y)`. It works fine. Prove me wrong with an example

littleredcomputer: We can't leave this here, because it leaves the reader with the impression that something dark is going on that nobody wants to talk about. What are we afraid will happen if we call `jnp.logical_thing` with bools?

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -12,73 +12,100 @@
...
-    mask_combinator: "MaskCombinator"
-    inner: Trace
-    check: BoolArray
-
-    def get_args(self):
+class MaskedTrace(
+    Generic[G, A, R, U],
+    Trace[
+        "MaskedCombinator[A, R, U]",
+        tuple[bool | BoolArray, A],
```
littleredcomputer, line 62: So here's the case where A is used ... interesting. We're saying that the argument type of this trace starts with a bool as its initial argument?

Can it be an array? Is this really what we want, since the function cannot be invoked with this modified arg list, can it? 

Wouldn't it be more sensible to have get_mask return that value rather than prepending it to args, which seems like lying since the GF wasn't called that way?

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -108,139 +135,151 @@ def masked_normal_draw(mean):
...
+            R,
+            ChoiceMapConstraint,
+            SelectionProjection | ChoiceMapProjection,
+            U,
+        ],
+        A,
+        R,
+        U,
+    ]:
         check, *inner_args = args
```
littleredcomputer, line 167: Aha! it's not actually the GF that gets the extra argument, it's simulate which eats the argument and then calls the GF! I think we need to be clearer about this. The user's GF may be invoked in masked and unmasked settings. In either case, there is no visible extra argument, nothing the user can test.

littleredcomputer: I'm beginning to wonder whether we shouldn't store the mask state somewhere else than the beginning of the arg list. Hm....

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -108,139 +135,151 @@ def masked_normal_draw(mean):
...
+        #               w' = 0.0
+        #
+        # For the transition True -> True, we apply the move to the existing
+        # unmasked trace. In that case, the weight is just the weight of the move.
+        #
+        #               w' = w
+        #
+        # In any case, we always apply the move... we're not avoiding
+        # that computation
+
```
littleredcomputer, line 267: this is helpful.

littleredcomputer (commented):
in progress, got as far as just before scan.py

src/genjax/_src/adev/core.py
MathieuHuot: Not sure this is what you're asking for, but there's a fairly detailed example on p29 of our paper (Fig 9, the continuation occurs with D(normal_REPARAM) which adds lines of code around the end of the program): https://arxiv.org/pdf/2406.15742 (I can explain notations or write with more simple notations if that's helpful, we're using the notation from the paper YOLO[58] which formalizes JAX's AD system).

There's a perhaps cleaner to read example on p18 of the ADEV paper: https://arxiv.org/pdf/2212.06386 (ignore "do" and "E", and \widetilde{dl} is a continuation). 

There's also the ADEV cookbook waiting for review if you're interested for a more visual example: https://github.com/probcomp/genjax/pull/1250

src/genjax/_src/core/generative/choice_map.py
MathieuHuot: Same, I would +1 this idea.

src/genjax/__init__.py
sritchie: yes let's do that, @littleredcomputer 's comment is a good reminder that this is probably going to hose users of genjax (since we know all of our types aren't solid, we have 200 type errors)

src/genjax/_src/core/generative/core.py
sritchie: @littleredcomputer one place it comes in is in contramap, where you have to make sure that your `pre` function generates the correct types.

PR #1243, opened 2024-07-23
Adopt @huebert's Pythonic base class into Pytree.

littleredcomputer:
add a unit test for Pytree.partial with scan (see GEN-339)

### DIFF
--- a/src/genjax/_src/core/pytree.py
+++ b/src/genjax/_src/core/pytree.py
@@ -498,25 +498,34 @@ class Closure(Pytree):
 
 
 def nth(x: Pytree, idx: Int):
+    """Returns a Pytree in which `[idx]` has been applied to every leaf."""
     return jtu.tree_map(lambda v: v[idx], x)
 
 
-class Pythonic(Pytree):
+class PythonicPytree(Pytree):
     """
     A class that adds support for bracket indexing/slicing, sequence-like operations,
-    and concatenation to make working with pytrees more Pythonic.
+    and concatenation to make working with Pytrees more Pythonic. The base class is
+    appropriate for Pytrees which have a unform shape over leaves (or at least each
+    leaf's initial axis should have the same length).
     """
 
     def __getitem__(self, idx):
+        """Return a pytree in which each leaf has been sliced by `idx`."""
         return nth(self, idx)
 
     def __len__(self):
+        """Return the "length" of the Pytree. This should only be used on
+        Pytrees which have a uniform shape over leaves; it operates by
+        returning the length of the "first" leaf."""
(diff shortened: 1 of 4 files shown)


src/genjax/_src/core/pytree.py
```diff
@@ -494,3 +495,35 @@ def f(v, x):
 
     def __call__(self, *args, **kwargs):
         return self.fn(*self.dyn_args, *args, **kwargs)
+
+
+def nth(x: Pytree, idx: Int):
+    return jtu.tree_map(lambda v: v[idx], x)
+
+
+class Pythonic(Pytree):
```
littleredcomputer: Maybe the name should be PythonicPytree

littleredcomputer: I tried making this a decorator but learned for myself what had stopped @mhuebert from doing so: we would need a type incantation that says "This function accepts a _class_ K and returns a class K' with additional member functions." That sounds an awful lot like saying "Class Pythonic extends Pytree in the following way," and a base class is the least tricky way of accomplishing this.

src/genjax/_src/core/pytree.py
```diff
@@ -494,3 +495,35 @@ def f(v, x):
 
     def __call__(self, *args, **kwargs):
         return self.fn(*self.dyn_args, *args, **kwargs)
+
+
+def nth(x: Pytree, idx: Int):
```
sritchie, line 500: can we add docstrings to these, noting what they do for pytree inheritors? can be short, even LLM-generated if you're on Cursor!

src/genjax/_src/core/pytree.py
sritchie: that's fine! I like that name `PythonicPytree` as well.

tests/generative_functions/test_scan_combinator.py
```diff
@@ -338,3 +338,48 @@ def model(k):
...
         assert w < -100.0
+
+
+class TestScanWithParameters:
+    @pytest.fixture
+    def key(self):
+        return jax.random.PRNGKey(314159)
+
+    @staticmethod
+    def gen_partial(gen_fn, closed_over):
```
sritchie: I don't see this used anywhere...

src/genjax/_src/core/pytree.py
littleredcomputer: done

src/genjax/_src/core/pytree.py
littleredcomputer: done

mhuebert (approved):
LGTM!

PR #1246, opened 2024-07-23
Adding a cookbook entry on resampling strategies for SMC (GEN-381)

MathieuHuot:
This can only be finished after #1228 is merged as it will reuse the components from that library to showcase the use of different strategies from the library.


MathieuHuot:
This PR was barely started and with all the changes in `edit` and  the new way of thinking about editRequest, I don't think we'd write it in the style of the "compositional SMC library" as in Gen.jl but rather using external inference logic or using `edit`.

PR #1247, opened 2024-07-24
Add `partial_apply` to `StaticGenerativeFunction`

sritchie:
This PR adds an instance method `partial_apply` that partially applies the supplied arguments to the generative function. I'd have called this `partial`, but Python doesn't allow an instance method to have the same name as an existing static method.

We need something like this to be able to partially apply generative functions in a JAX-compatible manner. Calling `Pytree.partial` alone will not do the right thing.


femtomc:
I'm still not entirely sure how `update` works on this object (and many of the other derived combinators).

There's an ideal world where "it all just works" -- but I think we should stop adding things until we understand this better (fair that @littleredcomputer may not need to use `update` -- and can use the partial trick I showed)

MathieuHuot:
> I'm still not entirely sure how `update` works on this object (and many of the other derived combinators).
> 
> There's an ideal world where "it all just works" -- but I think we should stop adding things until we understand this better (fair that @littleredcomputer may not need to use `update` -- and can use the partial trick I showed)

@femtomc we can think more about the general case when the time comes but I think it's ok for different parts of the system to have different level of automation support for now: right now we're really trying to build a smooth and fast modelling language, pretty much with no automation. 

The SMC library will be a big jump in enabling cool automation and giving a framework for programmable inference that the automation can understand, leverage, and be combined with. 

On this particular case, if we're thinking of Gen style update, it should be easy: we call update on the inner generative function where there are some args for which we know there's no argdiff (the ones provided in the curry combinator).

femtomc:
> On this particular case, if we're thinking of Gen style update, it should be easy: we call update on the inner generative function where there are some args for which we know there's no argdiff (the ones provided in the curry combinator).

But that's not how that works -- if the curried generative function is created inside another generative function, and closes over arguments from the outer generative function which may have incremental information ... if I update a trace from the outer generative function, what happens?

There's an analogy to AD on higher order functions.

MathieuHuot:
> > On this particular case, if we're thinking of Gen style update, it should be easy: we call update on the inner generative function where there are some args for which we know there's no argdiff (the ones provided in the curry combinator).
> 
> But that's not how that works -- if the curried generative function is created inside another generative function, and closes over arguments from the outer generative function which may have incremental information ... if I update a trace from the outer generative function, what happens?
> 
> There's an analogy to AD on higher order functions.

ok good point about the nested case. It shouldn't affect my other points though. 
Also it shouldn't be hard to figure out: accumulate the argdiff from the curry with the argdiff for the rest of args, bundle that up in a big argdiff for the inner generative function and that should do.

PR #1250, opened 2024-07-29
Adding a first example as a cookbook entry of using ADEV for correctly differentiating probabilistic programs

MathieuHuot:


MathieuHuot:
@sritchie I think this doesn't need a detailed review, and can easily be merged. It's just a simple first cookbook to show some interaction between GenJAX and differentiation.

sritchie:
Sounds good, fix the build errors and let's merge!

sritchie (approved):

MathieuHuot:
@sritchie I don't understand the remaining linter bug for ruff. Locally it's not complaining. Is it possible my local ruff busted or something? I thought it'd be updated automatically with poetry but maybe not>

MathieuHuot:
> @sritchie I don't understand the remaining linter bug for ruff. Locally it's not complaining. Is it possible my local ruff busted or something? I thought it'd be updated automatically with poetry but maybe not>

ok nevermind it was indeed the ruff version on my end which was not up to date.

PR #1253, opened 2024-07-31
Fixing some problems in cookbooks after comments from Rif (GEN-388)

MathieuHuot:
- Fixing the misleading explanation of `update` in the `methods_of_generative_fun.ipynb` cookbook (will still require a entire cookbook or two to explain the details and subtleties, especially for the expressivity of generalised interface)
- Fixing the usage of random keys to be the intended usage in JAX (new key every time, even when not strictly necessary).

### DIFF
--- a/notebooks/cookbook/basics/generative_fun.ipynb
+++ b/notebooks/cookbook/basics/generative_fun.ipynb
@@ -168,6 +168,7 @@
    "outputs": [],
    "source": [
     "fast_beta_bernoulli_process.simulate(key, (1.0,))\n",
+    "key, subkey = jax.random.split(key)\n",
     "jitted(key, (1.0,))"
    ]
   },
@@ -177,8 +178,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "key, subkey = jax.random.split(key)\n",
     "%timeit beta_bernoulli_process.simulate(key, (1.0,))\n",
+    "key, subkey = jax.random.split(key)\n",
     "%timeit fast_beta_bernoulli_process.simulate(key, (1.0,))\n",
+    "key, subkey = jax.random.split(key)\n",
     "%timeit jitted(key, (1.0,))"
    ]
   },
@@ -204,7 +208,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.11.4"
+   "version": "3.11.6"
   }
  },
  "nbformat": 4,
--- a/notebooks/cookbook/basics/generative_fun.ipynb
+++ b/notebooks/cookbook/basics/generative_fun.ipynb
@@ -55,7 +55,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "key = jax.random.PRNGKey(314159)"
+    "key = jax.random.PRNGKey(0)"
    ]
   },
   {
@@ -71,7 +71,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "tr = beta_bernoulli_process.simulate(key, (1.0,))"
+    "key, subkey = jax.random.split(key)\n",
+    "tr = beta_bernoulli_process.simulate(subkey, (1.0,))"
    ]
   },
   {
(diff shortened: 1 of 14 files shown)


notebooks/cookbook/basics/generative_fun.ipynb
```diff
@@ -168,6 +168,7 @@
    "outputs": [],
    "source": [
     "fast_beta_bernoulli_process.simulate(key, (1.0,))\n",
+    "key, subkey = jax.random.split(key)\n",
```
derifatives, line 156: This still doesn't look right to me? I'd expect you to call `jitted` on `subkey` (which is never reused), rather than `key`, which is used again later? (Similar comment in many other places.)

notebooks/cookbook/basics/some_oddities.ipynb
```diff
@@ -99,9 +103,9 @@
     "    return v\n",
     "\n",
     "\n",
-    "key = jax.random.PRNGKey(0)\n",
+    "key, subkey = jax.random.split(key)\n",
     "arg = (3.0,)  # 3 is not a valid probability but a valid logit\n",
-    "keys = jax.random.split(key, 30)\n",
+    "keys = jax.random.split(subkey, 30)\n",
     "# simulate 30 times\n",
     "jnp.array([g.simulate(key, arg).get_sample()[\"v\"] for key in keys])"
```
derifatives: Why not:
```
jax.vmap(g.simulate, in_axes=(0, None))(keys, arg).get_sample()
```

derifatives (commented):
This is a partial review, since it looks to me like you're still maybe not getting the key discipline right? Feel free to reach out on slack if you want to discuss.

notebooks/cookbook/basics/some_oddities.ipynb
MathieuHuot: I should totally do that. 
I don't remember the reason but it may have been at the time because of the subtlety of a trace obtained by a `vmap` not being a proper trace (e.g. the score will be a vector and not a single value because of the automatic array of struct -> struct of array con version that `vmap` does. You then need to use some treemap to recover what you want).

MathieuHuot:
@derifatives I think I fixed properly the usage of random keys this time.

sritchie (approved):

notebooks/cookbook/basics/methods_of_generative_fun.ipynb
```diff
@@ -327,7 +328,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We can also update a trace. This is for instance useful for performance optimizations in Metropolis-Hastings algorithms where often most of the trace doesn't change between time steps.\n",
+    "We can also update a trace. This is for instance useful as a performance optimization in Metropolis-Hastings algorithms where often most of the trace doesn't change between time steps.\n",
```
derifatives, line 331: Mixed feelings on this, since as of this writing I don't believe update actually performs this operation?

notebooks/cookbook/basics/methods_of_generative_fun.ipynb
```diff
@@ -452,14 +454,17 @@
    "source": [
     "jitted_update = jit(beta_bernoulli_process.update)\n",
     "\n",
-    "new_trace, weight_diff, ret_diff, discard_choice = jitted_update(key, old_trace, update)"
+    "key, subkey = jax.random.split(key)\n",
+    "new_trace, weight_diff, ret_diff, discard_choice = jitted_update(\n",
```
derifatives, line 458: Nit: I prefer to avoid naming arguments I'm never going to use.

notebooks/cookbook/expressivity/conditionals.ipynb
```diff
@@ -386,11 +389,13 @@
     "    return v\n",
     "\n",
     "\n",
-    "key = jax.random.PRNGKey(0)\n",
+    "key, subkey = jax.random.split(key)\n",
```
derifatives, line 392: Note for many places, if you're going to make a bunch of new calls, you can do (and I'd recommend):

key, tr1_key, tr2_key, tr3_key = jax.random.split(key, 4)

notebooks/cookbook/expressivity/custom_distribution.ipynb
```diff
@@ -124,8 +125,10 @@
     "\n",
     "\n",
     "probs = jnp.array([0.5, 0.5])\n",
-    "model.simulate(key, (probs,))\n",
-    "jit(model.simulate)(key, (probs,))"
+    "key, subkey = jax.random.split(key)\n",
```
derifatives, line 128: In a situation like this, where you're just calling jitted and unjitted versions of the same function, I'd probably stylistically choose to reuse the key, because conceptually it's the same call. (I'm not sure under what circumstances JAX would definitely give the same answer.)

notebooks/cookbook/expressivity/iterating_computation.ipynb
```diff
@@ -449,10 +460,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "sub_keys = jax.random.split(key, 3)\n",
+    "key, subkey = jax.random.split(key)\n",
+    "keys = jax.random.split(subkey, 3)\n",
```
derifatives, line 464: I usually do this a lot (first split the key, and then make an array of keys from one side of the split). I find this utility worthwhile in most notebooks:

```
def and_new_keys(key, N):
    key, tmp_key = jax.random.split(key)
    new_keys = jax.random.split(tmp_key, N)
    return key, new_keys
```

derifatives (commented):
Looks great. Left a few minor comments / thoughts, but nothing that needs changing. I didn't read every single line.

PR #1259, opened 2024-08-05
gen-420 and some type fixes

littleredcomputer:
cleans up a few type errors and adds a missing else case


src/genjax/_src/core/generative/choice_map.py
```diff
@@ -266,7 +266,7 @@ def select_none():
 
 @Pytree.dataclass
 class StaticSel(Selection):
-    addr: StaticAddressComponent = Pytree.static()
+    addr: StaticAddressComponent | EllipsisType = Pytree.static()
```
femtomc: Change to `ExtendedStaticAddressComponent` for consistency

femtomc (changes requested):
Single LoC change otherwise looks great!

src/genjax/_src/inference/smc.py
```diff
@@ -293,7 +293,7 @@ class ImportanceK(SMCAlgorithm):
     importance sampling."""
 
     target: Target
-    q: Optional[SampleDistribution] = Pytree.field(default=Pytree.const(None))
+    q: SampleDistribution = Pytree.field(default=Pytree.const(None))
```
femtomc: `None` is officially allowed in JAX, so I think we should restore the `Optional`, and just have `Pytree.field(default=None)`

littleredcomputer: OK. (To make that work, I have changed the definition of `static_check_none`). If you don't think anyone would ever use `Const(None)` then we can remove that method and just check `is None` at the call sites

littleredcomputer:
I also fixed a problem where smc was defining the instance variable `target`, which was shadowing the method `GenerativeFunction.target()` in an incompatible way (method vs. member). Since nobody was using the method and it was undocumented I deleted it.

littleredcomputer:
@femtomc ptal

femtomc (approved):

PR #1261, opened 2024-08-06
Vmap update importance processes args instead of argdiiffs now

arijit-dasgupta:
Previously in `update_importance`, the argdiffs were passed instead of args (which the function expected). This fix should extract the primals and ensure that the `args` field of the Vmap trace being returned are args and not argdiffs.

### DIFF
--- a/src/genjax/_src/generative_functions/combinators/vmap.py
+++ b/src/genjax/_src/generative_functions/combinators/vmap.py
@@ -182,20 +182,21 @@ class VmapCombinator(GenerativeFunction):
         self,
         key: PRNGKey,
         choice_map: ChoiceMap,
-        args: tuple,
+        argdiffs: tuple,
     ) -> tuple[Trace, Weight, Retdiff, UpdateProblem]:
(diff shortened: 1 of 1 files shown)


sritchie:
@arijitnoobstar I will take a look at this today!

src/genjax/_src/generative_functions/combinators/vmap.py
```diff
@@ -182,32 +182,33 @@ def update_importance(
         self,
         key: PRNGKey,
         choice_map: ChoiceMap,
-        args: tuple,
+        argdiffs: tuple,
```
littleredcomputer, line 185: @femtomc do we want to supply a `Problem` instance here? As in: GenericProblem = UpdateProblem + Argdiffs, here we want an ImportanceProblem with argdiffs perhaps?

sritchie:
@arijitnoobstar , really sorry for letting this sit for so long. @littleredcomputer fixed this issue in #1304 , and is going to adjust that to make you a co-author. We'll get this out in the release that we're attempting to get out this week.

Issue #1224, opened 2024-07-15
[BUG] Numpy Version should be <2.0.0


ahiser1117:
`Tensorflow-probability` has not been updated to work with `Numpy 2.0.0` and uses depreciated functions.

The current `pip install` of GenJAX installs `Numpy 2.0.0` and causes issues.

[Link to issue on Tensorflow-Prob GitHub](https://github.com/tensorflow/probability/issues/1814)

PR #1265, opened 2024-08-13
Fix numpy dep, jax dep

sritchie:
This PR

- fixes a reversion from https://github.com/probcomp/genjax/commit/5ff564fae93d2c562e4941dbbf3d479a940a9d0a in the jax required dependency
- moves the numpy dep from dev to mainline, fixing #1224


MathieuHuot (approved):
Looks good to me!

PR #1272, opened 2024-08-20
Retval of Gaussian Mixture in Cookbook is the incorrect order

ahiser1117:
`random_weighted` needs to return (weight, sample).

Current implementation has (sampled, weight).


sritchie:
re-assigning to @MathieuHuot

MathieuHuot:
Thanks @ahiser1117, good catch! I'll propagate the fix to the other cookbook that uses it and merge.

MathieuHuot (approved):

sritchie:
@MathieuHuot i think the docs build failed, would you mind taking a look?

PR #1273, opened 2024-08-20
GEN-482 (fix pyright errors in GenJAX)

littleredcomputer:
- pyright was giving different results in (VS Code, pre-commit) vs. Github action. I adjusted the GitHub action to follow Pylance, as recommended by the action author, which solved that problem. Previously, upright version 1.1.358 was in use; following Pylance moves that version to 1.1.373.
- Sample is replaced with ChoiceMap at all points where type errors follow the use of the former. I understand that ChoiceMap is a narrower concept than Sample, but we do not have enough type infrastructure in GenJAX to support the distinction presently
- the `inference` library is temporarily excluded, as it is not ready for prime-time 
- the handling of static booleans is now handled by `Flag`. The point of `Flag` is to concentrate the logic for making concreteness optimizations in one place. I believe that the systematic use of Flag has corrected a few spots where concreteness optimization had been overlooked, provoking the elision of a few `unmask` calls in the unit test suite. These require special attention to make sure that my conjecture is true.
- Constructing a Mask requires a Flag. It may be better to have the Mask constructor promote bare booleans to Flag types, or, to add a `post_init` assertion that the type is correct. It may be, however, that the user shouldn't be creating their own Masks, and so we can rely on the type checking for library development from now on to enforce correct Mask creation.
- pyright is (perhaps overly) scrupulous about the names of parameters in overridden methods, and doesn't allow overriding `*args` with a sequence of better names. Can't fight city hall, so I have changed this.
- To clear some errors I have resorted to a pattern of `assert isinstance(x, Foo), f"A Foo is required here, not {x}, [because ...]"`, which I think we should adopt since it's harder to figure out what to do if beartype gets involved.
-

### DIFF
--- a/.github/workflows/pyright.yml
+++ b/.github/workflows/pyright.yml
@@ -33,4 +33,4 @@ jobs:
 
       - uses: jakebailey/pyright-action@v1
         with:
-          version: 1.1.358
+          pylance-version: latest-release
--- a/notebooks/cookbook/basics/choicemap_creation_selection.ipynb
+++ b/notebooks/cookbook/basics/choicemap_creation_selection.ipynb
@@ -10,7 +10,7 @@
    "source": [
     "---\n",
     "title: Choicemaps\n",
-    "subtitle: What's a choicemap, and how do I create one or select from it? \n",
+    "subtitle: What's a choicemap, and how do I create one or select from it?\n",
     "---"
    ]
   },
@@ -324,7 +324,7 @@
     "\n",
     "\n",
     "trace = jax.jit(model.simulate)(key, (0.5,))\n",
-    "trace.get_sample()[(\"s\", \"v2\")]"
+    "trace.get_sample()[\"s\", \"v2\"]"
    ]
   },
   {
@@ -340,7 +340,7 @@
     "\n",
     "image = jnp.zeros([2, 3], dtype=jnp.float32)\n",
     "trace = sample_image.simulate(key, (image,))\n",
-    "trace.get_sample()[..., ..., \"new_pixel\"]"
+    "trace.get_choices()[..., ..., \"new_pixel\"]"
    ]
   },
   {
@@ -359,7 +359,7 @@
     "\n",
     "\n",
     "trace = hmm.simulate(key, (0.0, None))\n",
-    "trace.get_sample()[..., \"z\"], trace.get_sample()[3, \"y\"]"
+    "trace.get_choices()[..., \"z\"], trace.get_choices()[3, \"y\"]"
    ]
   },
   {
@@ -378,7 +378,7 @@
     "\n",
     "\n",
     "trace = model.simulate(key, (0.3,))\n",
-    "trace.get_sample()[..., \"x\"]"
+    "trace.get_choices()[..., \"x\"]"
    ]
   },
   {
--- a/notebooks/cookbook/basics/choicemap_creation_selection.ipynb
+++ b/notebooks/cookbook/basics/choicemap_creation_selection.ipynb
@@ -324,7 +324,7 @@
     "\n",
     "\n",
     "trace = jax.jit(model.simulate)(key, (0.5,))\n",
-    "trace.get_sample()[\"s\", \"v2\"]"
+    "trace.get_sample()[\"s\", \"v1\"]"
    ]
   },
   {
--- a/notebooks/cookbook/expressivity/conditionals.ipynb
+++ b/notebooks/cookbook/expressivity/conditionals.ipynb
@@ -397,9 +397,9 @@
     "key, subkey = jax.random.split(key)\n",
     "tr3 = jitted(subkey, (2.2,))\n",
     "(\n",
-    "    tr1.get_sample()[(\"s\", \"v1\")],\n",
-    "    tr2.get_sample()[(\"s\", \"v2\")],\n",
-    "    tr3.get_sample()[(\"s\", \"v3\")],\n",
+    "    tr1.get_sample()[\"s\", \"v1\"],\n",
+    "    tr2.get_sample()[\"s\", \"v2\"],\n",
+    "    tr3.get_sample()[\"s\", \"v3\"],\n",
     ")"
    ]
   },
@@ -428,7 +428,7 @@
     "jitted = switch_model_v2.simulate\n",
     "key, subkey = jax.random.split(key)\n",
     "tr = jitted(subkey, (0.0,))\n",
-    "tr.get_sample()[(\"switch\", \"v1\")]"
+    "tr.get_sample()[\"switch\", \"v1\"]"
    ]
   }
  ],
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,7 +109,7 @@ pythonVersion = "3.11"
 venvPath = "."
 venv = ".venv"
 include = ["src", "tests"]
-exclude = ["**/__pycache__"]
+exclude = ["**/__pycache__", "noxfile.py", "src/genjax/_src/inference/*.py"]
 defineConstant = { DEBUG = true }
 typeCheckingMode = "standard"
 reportMatchNotExhaustive = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,7 +109,10 @@ pythonVersion = "3.11"
 venvPath = "."
 venv = ".venv"
 include = ["src", "tests"]
-exclude = ["**/__pycache__", "noxfile.py", "src/genjax/_src/inference/*.py"]
+# TODO: inference has been excluded because this code my not be ready for type-correctness.
+#       msgpack dependency is optional, so it doesn't typecheck
+#       typecheck doesn't see noxfiles's import of nox
+exclude = ["**/__pycache__", "noxfile.py", "src/genjax/_src/inference/*.py", "src/genjax/_src/core/serialization/msgpack.py"]
 defineConstant = { DEBUG = true }
 typeCheckingMode = "standard"
 reportMatchNotExhaustive = true
--- a/src/genjax/_src/adev/primitives.py
+++ b/src/genjax/_src/adev/primitives.py
@@ -87,8 +87,9 @@ def reinforce(sample_func, logpdf_func):
 
 @Pytree.dataclass
 class FlipEnum(ADEVPrimitive):
-    def sample(self, key, p):
-        return 1 == tfd.Bernoulli(probs=p).sample(seed=key)
+    def sample(self, key, *args):
+        (probs,) = args
+        return 1 == tfd.Bernoulli(probs=probs).sample(seed=key)
 
     def jvp_estimate(
         self,
@@ -127,7 +128,7 @@ flip_enum = FlipEnum()
 @Pytree.dataclass
 class FlipMVD(ADEVPrimitive):
     def sample(self, key, *args):
-        p = args[0]
+        p = (args,)
         return 1 == tfd.Bernoulli(probs=p).sample(seed=key)
 
     def jvp_estimate(
@@ -153,7 +154,8 @@ flip_mvd = FlipMVD()
 
 @Pytree.dataclass
 class FlipEnumParallel(ADEVPrimitive):
-    def sample(self, key, p):
+    def sample(self, key, *args):
+        (p,) = args
         return 1 == tfd.Bernoulli(probs=p).sample(seed=key)
 
     def jvp_estimate(
@@ -189,7 +191,8 @@ flip_enum_parallel = FlipEnumParallel()
 
 @Pytree.dataclass
 class CategoricalEnumParallel(ADEVPrimitive):
-    def sample(self, key, probs):
+    def sample(self, key, *args):
+        (probs,) = args
         return tfd.Categorical(probs=probs).sample(seed=key)
 
     def jvp_estimate(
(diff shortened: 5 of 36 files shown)


sritchie:
@littleredcomputer can I bug you to add a project description & title? then when we squash merge folks can get context here

littleredcomputer:
> @littleredcomputer can I bug you to add a project description & title? then when we squash merge folks can get context here

You can once I'm finished; then I'll add a resume of the changes and things to look out for. (Pyright is now green, but there are two failing unit tests)

.github/workflows/pyright.yml
```diff
@@ -33,4 +33,4 @@ jobs:
 
       - uses: jakebailey/pyright-action@v1
         with:
-          version: 1.1.358
```
sritchie, line 36: interesting, does this pin to whatever pylance is using?

notebooks/cookbook/expressivity/conditionals.ipynb
```diff
@@ -397,9 +397,9 @@
     "key, subkey = jax.random.split(key)\n",
     "tr3 = jitted(subkey, (2.2,))\n",
     "(\n",
-    "    tr1.get_sample()[(\"s\", \"v1\")],\n",
-    "    tr2.get_sample()[(\"s\", \"v2\")],\n",
-    "    tr3.get_sample()[(\"s\", \"v3\")],\n",
+    "    tr1.get_sample()[\"s\", \"v1\"],\n",
```
sritchie: since I don't know the API well yet, are the before/after here equivalent?

pyproject.toml
```diff
@@ -109,7 +109,10 @@ pythonVersion = "3.11"
 venvPath = "."
 venv = ".venv"
 include = ["src", "tests"]
-exclude = ["**/__pycache__"]
+# TODO: inference has been excluded because this code my not be ready for type-correctness.
+#       msgpack dependency is optional, so it doesn't typecheck
```
sritchie, line 113: it should typecheck fine if we modify the `poetry install` command to include `--all-extras` as an option.

Looking now, we already do that in `pyright.yml`:

image

so this SHOULD be able to typecheck in CI just fine.

pyproject.toml
```diff
@@ -109,7 +109,10 @@ pythonVersion = "3.11"
...
+#       typecheck doesn't see noxfiles's import of nox
+exclude = ["**/__pycache__", "noxfile.py", "src/genjax/_src/inference/*.py", "src/genjax/_src/core/serialization/msgpack.py"]
```
sritchie, line 115: do you want to add `#noqa` surgically, vs excluding these full files?

src/genjax/_src/adev/primitives.py
```diff
@@ -239,7 +242,8 @@ def _inner(probs, primals):
 
 @Pytree.dataclass
 class NormalREPARAM(TailCallADEVPrimitive):
-    def sample(self, key, loc, scale_diag):
+    def sample(self, key, *args):
```
sritchie, line 245: nice, this is all solid.

src/genjax/_src/adev/primitives.py
```diff
@@ -392,8 +396,8 @@ def _inner(alpha, beta):
 class Baseline(ADEVPrimitive):
     prim: ADEVPrimitive
 
-    def sample(self, key, b, *args):
```
sritchie, line 395: odd that this got past ruff before, which usually calls out unused variables correctly.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -207,8 +205,8 @@ def maybe(cls, flag: Bool | BoolArray, s: "Selection") -> "Selection":
 
 @Pytree.dataclass
 class AllSel(Selection):
-    def check(self) -> Bool | BoolArray:
-        return True
+    def check(self) -> Flag:
```
sritchie, line 206: praise Flag

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -269,11 +266,11 @@ class StaticSel(Selection):
     addr: ExtendedStaticAddressComponent = Pytree.static()
     s: Selection = Pytree.field()
 
-    def check(self) -> Bool | BoolArray:
-        return False
+    def check(self) -> Flag:
+        return flag(False)
```
sritchie, line 268: thank you!!

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -353,10 +348,8 @@ class OrSel(Selection):
     s1: Selection
     s2: Selection
 
-    def check(self) -> Bool | BoolArray:
-        check1 = self.s1.check()
-        check2 = self.s2.check()
-        return staged_or(check1, check2)
+    def check(self) -> Flag:
```
sritchie, line 349: love this API

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -794,10 +789,10 @@ class XorChm(ChoiceMap):
     c1: ChoiceMap
     c2: ChoiceMap
 
-    def get_value(self) -> Optional[Any]:
+    def get_value(self) -> Any:
```
sritchie, line 790: is this really all we can say about the return type? I guess it is for now until we track type params through.

I'm excited to add what we can salvage of the generics on top of this base.

src/genjax/_src/core/generative/core.py
```diff
@@ -49,7 +46,7 @@
 if TYPE_CHECKING:
     import genjax
 
-_C = TypeVar("_C", bound=Callable)
+_C = TypeVar("_C", bound=Callable[..., Any])
```
sritchie, line 49: nice, I thought I had caught these bare callables

src/genjax/_src/core/generative/core.py
```diff
@@ -1712,7 +1715,7 @@ def update(
     @typecheck
     def assess(
         self,
-        sample: Sample,
+        sample: "genjax.ChoiceMap",
```
sritchie, line 1718: nice, so assess now at the top level takes a `ChoiceMap`? This wasn't ready to generalize to the tells-us-nothing sample I suppose

src/genjax/_src/core/interpreters/staging.py
```diff
@@ -34,40 +39,72 @@
...
+def flag(b: bool | BoolArray):
+    return Flag(b, not isinstance(b, jc.Tracer))
+
+
+@Pytree.dataclass(eq=False)
+class Flag(Pytree):
+    f: bool | BoolArray
+    concrete: bool = Pytree.static()
+
+    def and_(self, f: "Flag") -> "Flag":
```
sritchie, line 66: any interest in overloading `__or__` and `__and__` here so folks can use `&` and `|` or is that too cute?

src/genjax/_src/core/interpreters/staging.py
```diff
@@ -113,13 +149,21 @@ def cached_stage_dynamic(flat_fun, in_avals):
     return typed_jaxpr
 
 
+# This function has been cloned from api_util, since it is not exported from that module
```
sritchie, line 161: you could do this, or could also do `#noqa` on that import... this is probably better.

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -118,9 +118,9 @@ def simulate(
         key: PRNGKey,
         args: tuple,
     ) -> MaskTrace:
-        check, *inner_args = args
-        tr = self.gen_fn.simulate(key, tuple(inner_args))
-        return MaskTrace(self, tr, check)
+        check, inner_args = args[0], args[1:]
```
sritchie, line 121: nice

src/genjax/_src/generative_functions/static.py
```diff
@@ -604,7 +604,7 @@ def assess(
         self,
         sample: ChoiceMap,
         args: tuple,
-    ) -> tuple[ArrayLike, Any]:
```
sritchie, line 607: nice, was this a bug before?

src/genjax/_src/inference/smc.py
```diff
@@ -665,21 +667,21 @@ def importance(
         key: PRNGKey,
         constraint: Constraint,
         args: tuple,
-    ) -> tuple[Trace, Weight, UpdateProblem]:
+    ) -> tuple[Trace, Weight]:
```
sritchie, line 670: this looks like a bigger change that could use some attention from @femtomc

src/genjax/pretty.py
```diff
@@ -47,19 +47,21 @@ def _pytree_handler(node, subtree_renderer):
                     "background_color must be provided if background_pattern is"
                 )
 
-            def wrap_block(block):
+            def wrapper1(block):
```
sritchie: are there more descriptive names we can use for these two wrappers?

tests/generative_functions/test_switch_combinator.py
```diff
@@ -222,7 +222,6 @@ def f2():
         # Just select 0 in all branches for simplicity:
         tr = jax.vmap(s.simulate, in_axes=(0, None))(keys, (0, (), ()))
         y = tr.get_choices()["y"]
-        y = y.unmask()
```
sritchie, line 225: are masks unwrapped automatically now?

sritchie (commented):
First pass!

.github/workflows/pyright.yml
littleredcomputer: Yes (apparently with a lag of a day or two)

notebooks/cookbook/expressivity/conditionals.ipynb
littleredcomputer: This fix was automatically applied by Ruff: `[ ]` syntax is provided to the `getitem` method as a tuple, so the parens are superfluous

tests/generative_functions/test_switch_combinator.py
littleredcomputer: I am hoping that what has happened is that systematic use of `Flag` has uncovered overlooked concreteness optimizations, and that the previous Masks were incorrect. Special attention required here

src/genjax/pretty.py
littleredcomputer: fixed

src/genjax/_src/generative_functions/static.py
littleredcomputer: I don't think the previous types were "wrong" but pyright insists that overriders respect the super type here. Taking a second look I have changed to `[Score, Retval]` which is even better

src/genjax/_src/core/interpreters/staging.py
littleredcomputer: ruff deletes `#noqa` on imports :(

src/genjax/_src/core/interpreters/staging.py
littleredcomputer: I could be talked into this

src/genjax/_src/core/generative/core.py
littleredcomputer: I am hoping we can reconsider the type question by building on a green state

src/genjax/_src/adev/primitives.py
littleredcomputer: this was an odd case! It was used to drop an arg, so maybe that was enough of an "effect"?

pyproject.toml
littleredcomputer: I think that #noqa is the wrong choice here, since it suggests that what's being done is intentional. The inference library doesn't really seem ready to publish to me, but maybe we need a new Issue to cover the expansion of uprightness to this module

pyproject.toml
littleredcomputer: I will try this

pyproject.toml
sritchie: issue sounds great

pyproject.toml
littleredcomputer: this didn't work (the poetry-install hook does not appear to run, despite the preconditions in the documentation having been met)

notebooks/cookbook/basics/choicemap_creation_selection.ipynb
```diff
@@ -324,7 +324,7 @@
     "\n",
     "\n",
     "trace = jax.jit(model.simulate)(key, (0.5,))\n",
-    "trace.get_sample()[\"s\", \"v2\"]"
+    "trace.get_sample()[\"s\", \"v1\"]"
```
sritchie, line 327: notice this change, was this a bug?

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -329,10 +325,8 @@ class AndSel(Selection):
     s1: Selection
     s2: Selection
 
-    def check(self) -> Bool | BoolArray:
-        check1 = self.s1.check()
-        check2 = self.s2.check()
-        return staged_and(check1, check2)
+    def check(self) -> Flag:
+        return self.s1.check().and_(self.s2.check())
```
sritchie, line 328: if you do `__and__`, you can still call it directly as a method if you don't want the explicit `&`

sritchie (approved):

notebooks/cookbook/basics/choicemap_creation_selection.ipynb
littleredcomputer: PRNG instability? The trace does contain the one but not the other. I looked in the past to see if there was evidence that something else had changed but this build has been broken for a while so I didn't find any evidence to the contrary.

PR #1277, opened 2024-08-22
add R generics throughout

sritchie:
This PR adds an `R` generic to `Distribution` and `Trace` and a number of other classes that these depend on.


pyproject.toml
```diff
@@ -117,6 +117,8 @@ defineConstant = { DEBUG = true }
 typeCheckingMode = "standard"
 reportMatchNotExhaustive = true
 reportMissingImports = true
+# reportMissingTypeArgument = true
```
sritchie, line 120: turning this back on exposes many more places where we'll need to wire an `R` in.

src/genjax/_src/core/generative/core.py
```diff
@@ -48,9 +49,19 @@
 
 _C = TypeVar("_C", bound=Callable[..., Any])
 ArgTuple = TypeVar("ArgTuple", bound=tuple)
+
+# Generative Function type variables
 R = TypeVar("R")
```
sritchie, line 54: TODO: import R vs writing a new R in files? does this matter?

src/genjax/_src/core/generative/core.py
```diff
@@ -48,9 +49,19 @@
...
+"""
+Generic denoting the return type of a generative function.
+"""
+
 S = TypeVar("S")
 
+Carry = TypeVar("Carry")
```
sritchie, line 61: these don't all need to be here, we don't use X yet

src/genjax/_src/core/generative/core.py
```diff
@@ -511,26 +513,26 @@ def importance_sampling(key, constraint):
...
         return IgnoreKwargs(self)
 
     def get_trace_shape(self, *args) -> Any:
         return get_trace_shape(self, args)
 
-    def get_empty_trace(self, *args) -> Trace:
-        data_shape = self.get_trace_shape(*args)
-        return jtu.tree_map(lambda v: jnp.zeros(v.shape, dtype=v.dtype), data_shape)
+    def get_empty_trace(self, *args) -> Trace[R]:
+        trace_template = self.get_trace_shape(*args)
```
sritchie, line 533: @femtomc , let me know if we should change the `shape` name here

src/genjax/_src/core/generative/core.py
```diff
@@ -835,7 +837,8 @@ def treescope_color(self):
     # Combinators #
     ###############
 
-    def vmap(self, /, *, in_axes: InAxes = 0) -> "GenerativeFunction":
+    # TODO think through, or note, that the R that comes out will have to be bounded by pytree.
```
sritchie, line 839: todo we need to document the slight weirdness around pytree-shaped things...

src/genjax/_src/core/generative/core.py
littleredcomputer: good question. my inclination would be to declare them as needed -> fix all this in Python 3.12, and skip the part where we import one-letter names. Alternative: add them to genjax.typing

src/genjax/_src/core/generative/core.py
```diff
@@ -912,13 +915,13 @@ def normal_draw(mean):
         return genjax.repeat(n=n)(self)
 
     def scan(
-        self,
+        self: "GenerativeFunction[tuple[Carry, Y]]",
```
littleredcomputer, line 917: This one got the crowd dancin'

src/genjax/_src/core/interpreters/staging.py
```diff
@@ -106,11 +107,13 @@ def concrete_false(self):
...
-    def where(self, t, f):
+    def where(self, t: ArrayLike, f: ArrayLike) -> ArrayLike:
         """Return t or f according to the truth value contained in this flag
         in a manner that works in either the concrete or dynamic context"""
-
-        return jax.lax.select(jnp.all(self.f), t, f)
+        if self.concrete:
+            return t if self.f else f
+        else:
+            return jax.lax.select(jnp.all(self.f), t, f)
```
littleredcomputer: Would you kindly apply the following patch:

```
diff --git a/tests/core/test_staging.py b/tests/core/test_staging.py
index 93311634..84dd1851 100644
--- a/tests/core/test_staging.py
+++ b/tests/core/test_staging.py
@@ -75,3 +75,5 @@ class TestFlag:
     def test_where(self):
         assert Flag(True, concrete=True).where(3.0, 4.0) == 3
         assert Flag(False, concrete=True).where(3.0, 4.0) == 4
+        assert Flag(True, concrete=False).where(3.0, 4.0) == 3
+        assert Flag(False, concrete=False).where(3.0, 4.0) == 4
```

littleredcomputer (approved):
I made one small unit test request

PR #1279, opened 2024-08-23
Annotate bare tuples

sritchie:
This PR fixes errors from `tuple`, `list` and `dict` appearing without type parameters. Fixing these leaves us with only the "missing type parameters" errors for the type params that we've added.


femtomc (approved):

littleredcomputer (approved):
lgtm

PR #1282, opened 2024-08-23
More generics

sritchie:
Generics provided for dimap, vmap, mask and repeat, with some other ones sprinkled around.


littleredcomputer (approved):
lgtm

PR #1288, opened 2024-08-27
Add more missing generics

sritchie:
This PR:

- adds all missing generics
- `tfp_distribution` objects now report their type as `Distribution[Array]`
- bumps pyright to `1.1.377`
- relaxes the `R` of `switch` and `mix` to `Any`, since we can't say anything the final return type thanks to the multiple return types that might come in
- modifies `Mask.maybe` to never return `None`
- `Mask.maybe_none` made more readable, and with clear types
- fixes bugs in `_DiscreteHMMLatentSequencePosterior`
- `update_constraint` in `Distribution` gains a bugfix on the `UpdateProblem` returned from one of the branches, and an assertion that the type of choicemap we're assuming made it in (`MaskChm`) was in fact a `MaskChm`.

### DIFF
--- a/src/genjax/_src/core/generative/core.py
+++ b/src/genjax/_src/core/generative/core.py
@@ -91,7 +91,7 @@ When used under type checking, `Retdiff` assumes that the argument values are `P
 
 
 Retdiff = Annotated[
-    Diff[R],
+    Any,
     Is[Diff.static_check_tree_diff],
 ]
 
@@ -377,7 +377,7 @@ class Trace(Generic[R], Pytree):
         key: PRNGKey,
         problem: GenericProblem | UpdateProblem,
         argdiffs: tuple[Any, ...] | None = None,
-    ) -> tuple["Trace[R]", Weight, Retdiff[R], UpdateProblem]:
+    ) -> tuple["Trace[R]", Weight, Retdiff, UpdateProblem]:
         """
         This method calls out to the underlying [`GenerativeFunction.update`][genjax.core.GenerativeFunction.update] method - see [`UpdateProblem`][genjax.core.UpdateProblem] and [`update`][genjax.core.GenerativeFunction.update] for more information.
         """
@@ -595,7 +595,7 @@ class GenerativeFunction(Generic[R], Pytree):
         key: PRNGKey,
         trace: Trace[R],
         update_problem: GenericProblem,
-    ) -> tuple[Trace[R], Weight, Retdiff[R], UpdateProblem]:
+    ) -> tuple[Trace[R], Weight, Retdiff, UpdateProblem]:
         """
         Update a trace in response to an [`UpdateProblem`][genjax.core.UpdateProblem], returning a new [`Trace`][genjax.core.Trace], an incremental [`Weight`][genjax.core.Weight] for the new target, a [`Retdiff`][genjax.core.Retdiff] return value tagged with change information, and a backward [`UpdateProblem`][genjax.core.UpdateProblem] which requests the reverse move (to go back to the original trace).
 
@@ -1269,9 +1269,7 @@ class GenerativeFunction(Generic[R], Pytree):
 
         return genjax.mask(self)
 
-    def or_else(
-        self, gen_fn: "GenerativeFunction[S]", /
-    ) -> "GenerativeFunction[R | S]":
+    def or_else(self, gen_fn: "GenerativeFunction[R]", /) -> "GenerativeFunction[R]":
         """
         Returns a [`GenerativeFunction`][genjax.GenerativeFunction] that accepts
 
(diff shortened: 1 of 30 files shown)


src/genjax/_src/core/generative/core.py
```diff
@@ -91,10 +91,11 @@
 
 
 Retdiff = Annotated[
-    R,
```
sritchie: This is wrong, this should be `R` - this one is weird, there's a "runtime" check that all leaves are diffs, but `R` can be either a diff or some pytree with diff leaves.

littleredcomputer: A Diff is a pytree too, but (say) a pair of Diffs is not a Diff, but a Pytree with Diff leaves: so the problem is that Pytree doesn't really have a presence in the type system. I guess this is the best we can do

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -63,28 +63,25 @@ class Mask(Generic[R], Pytree):
...
-        return (
-            None
-            if v is None
-            else v
-            if f.concrete_true()
-            else None
-            if f.concrete_false()
-            else Mask.maybe(f, v)
-        )
+    def maybe_none(cls, f: Flag, v: R) -> "R | Mask[R] | None":
```
littleredcomputer: might be interesting to consider `f.mabye(v)`

littleredcomputer (approved):
lots of excellent work in there @sritchie

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -63,36 +63,32 @@ class Mask(Generic[R], Pytree):
     value: R
 
     @classmethod
-    def maybe(cls, f: Flag, v: Any):
+    def maybe(cls, f: Flag, v: "R | Mask[R]") -> "Mask[R]":
```
littleredcomputer, line 67: I'm still unclear why this and maybe_none are @classmethods

src/genjax/_src/core/interpreters/forward.py
```diff
@@ -34,7 +34,7 @@
 
 
 def batch_fun(fun: lu.WrappedFun, in_dims):
-    fun, out_dims = batching.batch_subtrace(fun)
+    fun, out_dims = batching.batch_subtrace(fun)  # pyright: ignore
```
littleredcomputer, line 37: Are we giving up on this? It may be the right strategy, but I'd make a follow up ticket to see If this can be tightened up

src/genjax/_src/core/interpreters/staging.py
```diff
@@ -169,7 +169,7 @@ def stage(f):
     def wrapped(*args, **kwargs):
         fun = lu.wrap_init(f, kwargs)
         flat_args, in_tree = jtu.tree_flatten(args)
-        flat_fun, out_tree = flatten_fun_nokwargs(fun, in_tree)
+        flat_fun, out_tree = flatten_fun_nokwargs(fun, in_tree)  # pyright: ignore
```
littleredcomputer, line 179: Is there a type assertion we could make to narrow this case? That is preferable to ignore IMO

src/genjax/_src/generative_functions/combinators/scan.py
```diff
@@ -99,7 +104,7 @@ class CheckerboardProblem(UpdateProblem):
 
 
 @Pytree.dataclass
-class ScanCombinator(GenerativeFunction):
+class ScanCombinator(Generic[Carry, Y], GenerativeFunction[tuple[Carry, Y]]):
```
littleredcomputer, line 107: I pause again to note the great victory that is the successful typing of Carry

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -354,7 +356,11 @@ def _false_branch(key, new_value, old_value):
                         DistributionTrace(self, primals, new_value, score),
                         w,
                         Diff.tree_diff_unknown_change(new_value),
-                        MaskedProblem(flag, old_value),
```
littleredcomputer, line 357: 👍

src/genjax/_src/core/interpreters/staging.py
sritchie: yeah but let's do it in a ticket, I just don't know how to handle it yet

src/genjax/_src/generative_functions/combinators/scan.py
sritchie: yes!

src/genjax/_src/core/generative/functional_types.py
littleredcomputer: sorry about this, I thought they were members of Flag

PR #1293, opened 2024-08-30
Fix assess in switch combinator (GEN-433)

sritchie:
This PR:

- Removes the python version fork for `typing_extensions`, as it's available in 3.11 and up as well
- threads the `R` type parameter through so that values traced with `@` return `R` instead of `Any`
- fixes the `assess` implementation of `switch` to properly pass branch args around and not create lists vs tuples

### DIFF
--- a/src/genjax/_src/core/generative/core.py
+++ b/src/genjax/_src/core/generative/core.py
@@ -1532,7 +1532,7 @@ class GenerativeFunction(Generic[R], Pytree):
 GLOBAL_TRACE_OP_HANDLER_STACK: list[Callable[..., Any]] = []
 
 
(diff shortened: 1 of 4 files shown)


src/genjax/_src/core/generative/core.py
```diff
@@ -1588,7 +1588,7 @@ def get_gen_fn_with_kwargs(self):
         return self.gen_fn.handle_kwargs()
 
     # NOTE: Supports callee syntax, and the ability to overload it in callers.
-    def __matmul__(self, addr):
+    def __matmul__(self, addr) -> R:
```
littleredcomputer, line 1591: nice work here

littleredcomputer (approved):

PR #1296, opened 2024-08-30
Remove noqa, ignore cls from vulture

sritchie:
This PR reverts the `_cls` changes that were causing warnings from pyright, and removes so extra `noqa` annotations that were floating around.


littleredcomputer (approved):
do we need vulture if we have coverage requirements?

sritchie:
@littleredcomputer I didn't turn on coverage as a blocker for merging, maybe we should though...

PR #1298, opened 2024-08-30
Only allow scalar flags to `Mask` constructor (GEN-523)

sritchie:
This PR modifies `Mask` so that it only accepts flags built out of arrays with a single entry (or primitive bools).


MathieuHuot (approved):
Ok, looks good to me, but I'm curious about Colin's input before we merge.

littleredcomputer:
@mhuot what is the code that should have been written in this case? (from the user's perspective, not inside GenJAX)

MathieuHuot:
> @mhuot what is the code that should have been written in this case? (from the user's perspective, not inside GenJAX)

@littleredcomputer  I'm not sure I fully get the question so let me know if I'm not answering it: an example code that currently runs and shouldn't:
```
@genjax.gen
def model(): 
  x = genjax.normal(0., 1.) @ "x"
  return x

key = jax.random.PRNGKey(0)
masks = jnp.array([True, True, False]) 

tr = model.mask().simulate(key, (masks,))
```
Here I'm expecting something like a type error because the model expects a boolean and it received a boolean array of size 3.

sritchie:
@MathieuHuot what if `model` returned an array of size 3? Should it succeed then?

MathieuHuot:
> @MathieuHuot what if `model` returned an array of size 3? Should it succeed then?

That's a design question. I think right now we'd do that with a `vmap` combinator wrapping around the `mask` combinator. We could perhaps have a more ergonomic masked-vmap combinator. 

The problem with the design you suggest is that it seems to open the way to a lot of thinking (and possible headaches) about broadcasting. For example I may still want the previous behaviour with a single mask even though the return/traces type would be an array, but now I'd also accept a mask tensor with the same leading or ending axes.

TLDR: right now I'd say only accept a boolean, and then by starring at examples from GenBrain/ChiSight we should think about better ways to deal with masking in more general situations (think more carefully about masked-scan and maybe masked-vmap combinators) either through broadcasting or via combinators that make it easier and less confusing to do certain patterns.

littleredcomputer:
Ultimately (I have updated the bug with some more study of this case) the shape of the masks argument should be compatible with the shape of the score, not the shape of the retval. To have a score that is not scalar, I argue, `vmap` must have been involved, either inside GenJAX, using the combinator, or outside GenJAX, producing a batched trace. In the first case, the combinator logic handles the fanout. In the latter case, you must operate on the trace with vmap, to pair up the flag vector with the score vector.

MathieuHuot:
> Ultimately (I have updated the bug with some more study of this case) the shape of the masks argument should be compatible with the shape of the score, not the shape of the retval. To have a score that is not scalar, I argue, `vmap` must have been involved, either inside GenJAX, using the combinator, or outside GenJAX, producing a batched trace. In the first case, the combinator logic handles the fanout. In the latter case, you must operate on the trace with vmap, to pair up the flag vector with the score vector.

This is a nice insight and recap! I think we'll most likely want to see examples of people willing to use this masking logic of manually pairing up the flag vector with the score vector outside a combinator (if they exist) to see if we could wrap it nicely as a combinator: `masked-scan` is one, and maybe we need one for `masked-vmap`. 
I don't know if there are other cases that go beyond these 2 for now.

sritchie:
@littleredcomputer @MathieuHuot you guys take care of the final approval / merge if this looks good, I can't quite track what we've decided on this PR from the discussion above. I'm happy to add @littleredcomputer 's note as a comment somewhere in the codebase, just tell me where.

MathieuHuot:
> @littleredcomputer @MathieuHuot you guys take care of the final approval / merge if this looks good, I can't quite track what we've decided on this PR from the discussion above. I'm happy to add @littleredcomputer 's note as a comment somewhere in the codebase, just tell me where.

@littleredcomputer could you handle this?

sritchie:
@littleredcomputer do you mind making a call here, for if we want to include this for 0.6.0 or no?

PR #1301, opened 2024-09-03
Lock down types for SwitchCombinator, cast / throw in concrete idx case (GEN-486)

sritchie:
@littleredcomputer and I discovered that `SwitchCombinator` ALREADY casts / requires that all of the incoming generative functions have the same return type, thanks to `jnp.choose` (invoked when an index is not an `int`).

This behavior allowed the user to see different return types when they passed an `int` index vs an array.

This PR:
- removes this difference in behavior by _always_ calling `jnp.choose` (to benefit from its shape-and-type checking and errors), but then in the case of a concrete `int` index, selecting using `vs[idx]` instead of returning the result of `jnp.choose`.
- types `SwitchCombinator` as `GenerativeFunction[R | Sum[R]]` and pushes this through `mix` and `or_else`
- swaps out `Sum.maybe_none` calls throughout `SwitchCombinator` for `Sum.maybe` calls

### DIFF
--- a/src/genjax/_src/core/generative/functional_types.py
+++ b/src/genjax/_src/core/generative/functional_types.py
@@ -203,19 +203,20 @@ class Sum(Generic[R], Pytree):
             return Sum.maybe(idx, vs)
 
     def maybe_collapse(self) -> R | Self:
-        def choose(*vs):
-            # Computing `result` above the branch allows us to:
-            # - catch incompatible types / shapes in the result
-            # - in the case of compatible types requiring casts (like bool => int),
-            #   result's dtype tells us the final type.
-            result = jnp.choose(idx, vs, mode="wrap")
-            if isinstance(idx, Int):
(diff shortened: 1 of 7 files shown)


src/genjax/_src/core/generative/functional_types.py
```diff
@@ -182,20 +182,16 @@ def maybe(
...
+        return Sum[R](idx, vs).maybe_collapse()
 
     @classmethod
     def maybe_none(
         cls,
         idx: ArrayLike | Diff[Any],
         vs: list[R],
-    ):
-        possibles = []
+    ) -> "R | Mask[R] | Sum[R] | None":
```
sritchie, line 193: @femtomc notice this wacky type... you can get a `Mask[R]` out if you pass a singleton list with a concrete index. That can't be intentional, right?

femtomc (approved):

PR #1303, opened 2024-09-03
Remove Sum type from `SwitchCombinator`'s return (GEN-536)

sritchie:
This PR attempts to remove the `Sum` type that's showed up in the return value of `SwitchCombinator`. To do this, we:

- removed `Sum`
- converted `Sum.maybe` to `staged_choose`
- removed `Sum.maybe_none`, since the behavior of returning `None` for all `None` inputs works already with `staged_choose`
- adds return types to many functions in and related to `ScanCombinator`
- simplifies the index generation logic in `XorChm` and `OrChm`

This PR builds on top of #1301 .


src/genjax/_src/core/generative/functional_types.py
```diff
@@ -106,120 +104,42 @@ def unsafe_unmask(self) -> R:
...
-    idx: ArrayLike | Diff[Any]
-    """
-    The runtime index tag for which value in `Sum.values` is active.
-    """
-    values: list[R]
-    """
-    The possible values for the `Sum` instance.
-    """
-
+class Sum:
```
sritchie: wdyt of moving the two methods here out of `Sum`?

src/genjax/_src/generative_functions/combinators/scan.py
```diff
@@ -203,24 +203,30 @@ def simulate(
     ) -> ScanTrace[Carry, Y]:
         carry, scanned_in = args
 
-        def _inner_simulate(key, carry, scanned_in):
+        def _inner_simulate(
```
sritchie, line 206: the new types above caused a type error in `scan` that I fixed by adding types to all of the inner functions.

src/genjax/_src/core/generative/functional_types.py
femtomc: If you want, we could move them to `staging.py` -- and define a `staged_choose` function -- which does the checking (for staticness) on the integer argument, and otherwise uses `jnp.choose`

femtomc (approved):

src/genjax/_src/core/generative/functional_types.py
sritchie: I kept it here, and now we only have `staged_choose`.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -766,11 +766,15 @@ def get_value(self) -> Any:
         v1 = self.c1.get_value()
         v2 = self.c2.get_value()
 
-        def pair_bool_to_idx(bool1, bool2):
```
littleredcomputer, line 769: consider:
```
err_check = check1.xor_(check2).not_()
...
idx = check1.where(0, 1)
```
you will have to supply `xor_`

src/genjax/_src/generative_functions/combinators/scan.py
littleredcomputer: 👍

littleredcomputer (approved):
left an idea you might like in a comment

PR #1304, opened 2024-09-04
fix GEN-524 by equilibrating argdiff shape in `VmapTrace`

littleredcomputer:
- correct an imbalance of argdiffs vs primals in VmapTrace
- stop auto-wrapping Flag around MaskTrace flag argument
- add Rif's failing example as unit test

### DIFF
--- a/src/genjax/_src/generative_functions/combinators/mask.py
+++ b/src/genjax/_src/generative_functions/combinators/mask.py
@@ -121,6 +121,9 @@ class MaskCombinator(Generic[R], GenerativeFunction[Mask[R]]):
         args: tuple[Any, ...],
     ) -> MaskTrace[R]:
         check, inner_args = args[0], args[1:]
+        if not isinstance(check, Flag):
+            check = Flag(check)
+
         tr = self.gen_fn.simulate(key, inner_args)
         return MaskTrace(self, tr, check)
 
--- a/src/genjax/_src/generative_functions/combinators/mask.py
+++ b/src/genjax/_src/generative_functions/combinators/mask.py
@@ -121,8 +121,7 @@ class MaskCombinator(Generic[R], GenerativeFunction[Mask[R]]):
         args: tuple[Any, ...],
     ) -> MaskTrace[R]:
         check, inner_args = args[0], args[1:]
-        if not isinstance(check, Flag):
-            check = Flag(check)
+        check = Flag.as_flag(check)
 
         tr = self.gen_fn.simulate(key, inner_args)
         return MaskTrace(self, tr, check)
--- a/src/genjax/_src/generative_functions/combinators/mask.py
+++ b/src/genjax/_src/generative_functions/combinators/mask.py
@@ -122,7 +122,7 @@ class MaskCombinator(Generic[R], GenerativeFunction[Mask[R]]):
     ) -> MaskTrace[R]:
         check, inner_args = args[0], args[1:]
         tr = self.gen_fn.simulate(key, inner_args)
-        return MaskTrace(self, tr, Flag(check))
+        return MaskTrace(self, tr, check)
 
     def update_change_target(
         self,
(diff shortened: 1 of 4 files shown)


littleredcomputer:
@femtomc Please look at the one line change in `vmap.py`. The `Mask.update` does a `cond` between `update_change_target` and `update_change_target_from_false`. One of these is routed to `VmapCombinator.update_importance` which puts the argdiffs in the trace, and the other is routed to `VmapCombinator.update_choice_map`, which puts the **primals** in the trace.  `Cond` requires that this choice be made consistently. Perhaps the "opposite" change is better?

femtomc:
The argdiffs should not go in the trace -- whichever option follows that pattern is the correct one.

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -121,8 +121,11 @@ def simulate(
         args: tuple[Any, ...],
     ) -> MaskTrace[R]:
         check, inner_args = args[0], args[1:]
+        if not isinstance(check, Flag):
+            check = Flag(check)
```
sritchie: Is there some constructor function we could write that would automatically handle this flattening? We have something like it in `Mask.maybe`...

src/genjax/_src/generative_functions/combinators/vmap.py
```diff
@@ -231,7 +231,7 @@ def _update(key, idx, subtrace, argdiffs):
         w = jnp.sum(w)
         retval = new_subtraces.get_retval()
         scores = new_subtraces.get_score()
-        map_tr = VmapTrace(self, new_subtraces, primals, retval, jnp.sum(scores))
+        map_tr = VmapTrace(self, new_subtraces, argdiffs, retval, jnp.sum(scores))
```
sritchie: as @femtomc notes, `argdiffs` should not go in the trace, so we do in fact want `primals` here.

sritchie: (and some other argdiffs in the file probably needs to change to primals?)

sritchie (approved):
Nice, approve post-changes

sritchie:
@littleredcomputer can you also look at https://github.com/probcomp/genjax/pull/1261? I feel like @arijitnoobstar 's PR is covering a similar / adjacent problem, it'd be great to get these both in.

littleredcomputer:
> @littleredcomputer can you also look at #1261? I feel like @arijitnoobstar 's PR is covering a similar / adjacent problem, it'd be great to get these both in.

Agree, merging that PR would have saved us some time. Rather than try to rebase it I propose to add @arijitnoobstar as an author of this PR after incorporating the rest of his fix

src/genjax/_src/generative_functions/combinators/vmap.py
littleredcomputer: ...perhaps, but we'll want test cases etc.

tests/generative_functions/test_mask_combinator.py
```diff
@@ -116,3 +115,44 @@ def test_mask_update_weight_to_argdiffs_from_false(self, key):
         w = tr.update(key, argdiffs)[1]
         assert w == 0.0
         assert w == tr.get_score()
+
+    def test_mask_scan_update(self, key):
+        def masked_scan_combinator(step, **scan_kwargs):
+            def scan_step_pre(state, flag):
+                return flag, state
+
```
femtomc, line 123: worried about this -- but not in this PR (more generally, in `masked_scan`)

It would be easy to generate generative functions where the masked return value could be masked False -- which means unwrapping it and returning the value is invalid (invalid meaning, if its use affects generative computation later on, the scores and weights computed may be incorrect)

sritchie:
@littleredcomputer note that we want to make sure we solve the issue in https://probcomp.slack.com/archives/C06EV7GHCCW/p1724876582854649?thread_ts=1724873178.418879&cid=C06EV7GHCCW.

PR #1305, opened 2024-09-04
Overhaul ChoiceMap docs, instance and class methods (GEN-308, GEN-328, GEN-544)

sritchie:
This PR:

- Removes all of the bare `select_*` and `*_choice_map` functions in favor of instance methods on `ChoiceMap` and `Selection`, or `build` class methods on the specific instances where appropriate
- adds docs to all classes and user-facing methods
- switches the order of parameters to the `*Sel` and `*Chm` methods so that the selection or choice map comes first, mirroring the `(self, *)` calling structure
- pushes logic like the static/dynamic address switches back into the objects, vs forcing the caller to distinguish
- adds a type parameter to `ValueChm`... of course this becomes `Any` for most use cases, but it's nice if you happen to care!

- [ ] TODO before merge: fix example compilation.
- [ ] remove html true from example blocks
- [ ] put types closer together
- [ ] add `ChoiceMap.d({('y','x'):3.0,'z':4.0})` alex compatible test
- [ ] make a test case out of
```python
In [9]: @genjax.gen
   ...: def f():
   ...:     return genjax.normal(0.0,1.0) @ ('x', 'a')
   ...: 

In [10]: tr = f.simulate(jax.random.PRNGKey(0), ())
In [19]: cake = ChoiceMap.d({('x','a'): jnp.asarray(1000.0)})

In [20]: tr.update(jax.random.PRNGKey(1), cake, ())
```

- [ ] note that tuple values are allowed as keys of 
- [ ]


src/genjax/_src/core/generative/choice_map.py
```diff
@@ -84,56 +85,155 @@ class Selection(ProjectProblem):
...
+
+        Example:
+            ```python exec="yes" html="true" source="material-block" session="choicemap"
+            base_selection = Selection.all()
+            indexed_selection = base_selection.indexed("x")
+            assert indexed_selection["x"]["any_subaddress"] == True
+            assert indexed_selection["y"].check() == False
+            ```
+        """
+        if isinstance(addr, ExtendedStaticAddressComponent):
```
sritchie: this switch is now pushed inside Selection

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -161,26 +261,6 @@ def check(self) -> Flag:
...
-    #################################################
-    # Convenient syntax for constructing selections #
-    #################################################
-
-    @classmethod
-    def all(cls) -> "Selection":
-        return select_all()
-
-    @classmethod
-    def str(cls, comp: ExtendedStaticAddressComponent, sel: "Selection") -> "Selection":
```
sritchie, line 173: these are gone in favor of `indexed`

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -384,40 +570,150 @@ def __getitem__(
...
+            assert chm["x"].get_value() == 1
+
+            nested_chm = ChoiceMapBuilder.a(("x", "y"), 2)
+            assert nested_chm["x"]["y"].get_value() == 2
+            ```
+        """
         addr = addr if isinstance(addr, tuple) else (addr,)
         new = ChoiceMap.value(v) if not isinstance(v, ChoiceMap) else v
         for comp in reversed(addr):
-            if isinstance(comp, StaticAddressComponent):
```
sritchie, line 410: this logic is now pushed inside `ChoiceMap.idx`

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -384,40 +570,150 @@ def __getitem__(
...
+            new = ChoiceMap.idx(new, comp)
+
         return new
 
 
 ChoiceMapBuilder = _ChoiceMapBuilder()
 
 
-def check_none(v) -> Flag:
+def check_none(v: Any | Mask[Any] | None) -> Flag:
```
sritchie: types for days

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -521,11 +817,141 @@ def get_submap(
...
+        Example:
+            ```python exec="yes" html="true" source="material-block" session="choicemap"
+            value_chm = ChoiceMap.value(42)
+            assert value_chm.get_value() == 42
+            ```
+        """
+        return ValueChm(v)
+
+    @classmethod
+    def idx(cls, v: Any, addr: ExtendedAddressComponent) -> "ChoiceMap":
```
sritchie: this now handles static and dynamic and switches internally, vs having a separate `str`.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -601,97 +1124,105 @@ def __contains__(
...
-    @classmethod
-    def str(cls, addr: StaticAddressComponent, v: Any) -> "ChoiceMap":
-        return choice_map_static(
-            addr, ChoiceMap.value(v) if not isinstance(v, ChoiceMap) else v
-        )
-
-    @classmethod
-    def idx(cls, addr: DynamicAddressComponent, v: Any) -> "ChoiceMap":
-        return choice_map_idx(
-            addr, ChoiceMap.value(v) if not isinstance(v, ChoiceMap) else v
```
sritchie, line 629: this funky logic all goes into `build` methods

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -601,97 +1124,105 @@ def __contains__(
...
     def get_submap(self, addr: ExtendedAddressComponent) -> ChoiceMap:
-        return EmptyChm()
+        return self
 
     def static_is_empty(self) -> Bool:
         return True
 
 
-choice_map_empty = EmptyChm()
+_empty = EmptyChm()
```
sritchie, line 1092: `_` makes this private

src/genjax/_src/generative_functions/combinators/scan.py
```diff
@@ -65,7 +65,7 @@ def get_retval(self) -> tuple[Carry, Y]:
 
     def get_sample(self):
         return jax.vmap(
-            lambda idx, subtrace: ChoiceMap.idx(idx, subtrace.get_sample()),
+            lambda idx, subtrace: ChoiceMap.idx(subtrace.get_sample(), idx),
         )(jnp.arange(self.scan_gen_fn.length), self.inner)
```
femtomc: why change the order?

sritchie: @femtomc this sticks with the numpy and `jax.numpy` convention of placing the data structure first in the list, and the "things we're applying to the data structure" afterward: https://jax.readthedocs.io/en/latest/jax.numpy.html

femtomc: cool, makes sense!

sritchie:
@femtomc lmk if this looks good... I will get the examples fixed.

I do want to understand `OrChm` , as the example that I got working still feels weird to me and I don't understand how to explain what is going on.

femtomc:
> I do want to understand OrChm , as the example that I got working still feels weird to me and I don't understand how to explain what is going on.

OrChm is XorChm -- without the error behavior. It lets you collide two choice maps which share the same addressed leaves without issue (it's how we represent "mutation", functionally -- in the choice map DSL)

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -312,18 +479,34 @@ def check(self) -> Flag:
...
+        s1: The first selection to be combined.
+        s2: The second selection to be combined.
+
+    Examples:
+        ```python exec="yes" html="true" source="material-block" session="choicemap"
+        sel1 = Selection.all().indexed("x")
+        sel2 = Selection.all().indexed("y")
+        or_sel = sel1 | sel2
+
+        # TODO what is going on with the example here? why is this true?
```
sritchie: this is weird to me. this feels like it is saying yes to a sequential selection when that should not pass?

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -521,11 +823,145 @@ def get_submap(
...
+        Example:
+            ```python exec="yes" html="true" source="material-block" session="choicemap"
+            import jax.numpy as jnp
+
+            # Static address
+            static_chm = ChoiceMap.idx(42, "x")
+            assert static_chm["x"] == 42
+
+            # Dynamic address
+            # TODO this looks like a bug that this is possible? assert fails.
```
sritchie: the assert fails because we have mismatched shaped on 42 and [1,2,3]. Should it be possible to create what I've written? Should we validate shapes?

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -601,97 +1136,108 @@ def __contains__(
...
+        ```python exec="yes" html="true" source="material-block" session="choicemap"
+        import jax.numpy as jnp
+
+        base_chm = ChoiceMap.value(jnp.array([1, 2, 3]))
+        idx_chm = base_chm.indexed(jnp.array([0, 1, 2]))
+
+        assert idx_chm.get_submap(1).get_value().unmask() == 2
+
+        # TODO this seems like a bug too, we get submaps for all sorts of addresses
+        assert idx_chm.get_submap(3).static_is_empty() == True
```
sritchie: `.get_submap(3)`, or `.get_submap(10)` etc all return the same thing.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -601,97 +1136,108 @@ def __contains__(
...
+        ```python exec="yes" html="true" source="material-block" session="choicemap"
+        import jax.numpy as jnp
+
+        base_chm = ChoiceMap.value(jnp.array([1, 2, 3]))
+        idx_chm = base_chm.indexed(jnp.array([0, 1, 2]))
+
+        assert idx_chm.get_submap(1).get_value().unmask() == 2
+
+        # TODO this seems like a bug too, we get submaps for all sorts of addresses
+        # assert idx_chm.get_submap(3).static_is_empty() == True
```
sritchie: this actually returns false for `(10)`, `(3)` etc

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -291,18 +444,32 @@ def check_fn(v):
...
+    This selection is true only if both of its constituent selections are true.
+    It allows for the combination of multiple selection criteria.
+
+    Attributes:
+        s1: The first selection to be combined.
+        s2: The second selection to be combined.
+
+    Examples:
+        ```python exec="yes" html="true" source="material-block" session="choicemap"
+        sel1 = Selection.all().indexed("y").indexed("x")
```
sritchie: This one also seems almost vacuous. is there a better test case? I couldn't get anything to return `True` unless they were both exactly the same. Should this return `True` on `x` if the second element of the paths are different?

src/genjax/_src/core/generative/choice_map.py
sritchie: TODO (@sritchie) add an explanation here - passing `[...]` says "do i match a prefix of what you've given me"? Note that there is an implicit ellipsis.

femtomc (approved):

PR #1306, opened 2024-09-04
fix GEN-387, GEN-333

littleredcomputer:
If the length in the `ScanCombinator` is `None`, use the observed length of the scanned-over array from the trace in its place.

GEN-333 proposes to run importance over a scan with `jnp.arange(0)`, which produces an array of shape `(0,)`, which JAX doesn't want to deal with so we have to screen it out.


sritchie (approved):

sritchie:
Nice work!

sritchie:
Can we check GEN-333 against this one? https://linear.app/chi-fro/issue/GEN-333/[bug]-error-in-importance-on-0-length-scan I think that will require an additional fix...

PR #1318, opened 2024-09-06
Annotate ignored types, add specific pyright ignores (GEN-512)

sritchie:
This PR fixes our blanket `pyright: ignore` directives in favor of

- adding explicit directives to ignore
- for the `staging` and `forward` ignores, adding back types that were obscured by jax's use of `@partial`.

### DIFF
--- a/src/genjax/_src/core/generative/core.py
+++ b/src/genjax/_src/core/generative/core.py
@@ -343,13 +343,13 @@ class Trace(Generic[R], Pytree):
         This method calls out to the underlying [`GenerativeFunction.update`][genjax.core.GenerativeFunction.update] method - see [`UpdateProblem`][genjax.core.UpdateProblem] and [`update`][genjax.core.GenerativeFunction.update] for more information.
         """
         if isinstance(problem, GenericProblem) and argdiffs is None:
-            return self.get_gen_fn().update(key, self, problem)  # pyright: ignore
+            return self.get_gen_fn().update(key, self, problem)  # pyright: ignore[reportReturnType]
         else:
             return self.get_gen_fn().update(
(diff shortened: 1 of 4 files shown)


src/genjax/_src/core/generative/core.py
```diff
@@ -1600,7 +1600,7 @@ def __matmul__(self, addr) -> R:
 
     # This override returns `R`, while the superclass returns a `GenerativeFunctionClosure`; this is
     # a hint that subclassing may not be the right relationship here.
-    def __call__(self, key: PRNGKey, *args) -> R:  # pyright: ignore
+    def __call__(self, key: PRNGKey, *args) -> R:  # pyright: ignore[reportIncompatibleMethodOverride]
```
littleredcomputer, line 1603: does this work:
```
def __call__(self, *args) -> R:
    key, *args = args
    ...
```

littleredcomputer (approved):
pending one comment

src/genjax/_src/core/generative/core.py
sritchie: It's not that, it's the incompatible return type...

```
Method "__call__" overrides class "GenerativeFunction" in an incompatible manner
  Return type mismatch: base method returns type "GenerativeFunctionClosure[R@GenerativeFunctionClosure]", override returns type "R@GenerativeFunctionClosure"
    "object*" is not assignable to "GenerativeFunctionClosure[R@GenerativeFunctionClosure]"basedpyright[reportIncompatibleMethodOv](https://github.com/detachhead/basedpyright/blob/main/docs/configuration.md#reportIncompatibleMethodOverride)
```

PR #1320, opened 2024-09-07
RandomChoice for sampling without replacement (GEN-13)

### LINKED ISSUE #903: [GJEP] Distribution for sampling without replacement
From @midfield, a possible implementation, which requires testing:

```python
def log_binomial_coefficient(x, y):
  return -jnp.log(x + 1) - jsp.special.betaln(y + 1, x - y + 1)
 
class RandomChoice(genjax.distributions.ExactDensity):
  def sample(self, key: types.Key, range: types.Int, shape=()):
    return jr.choice(key, range, shape, replace=False)
 
  def logpdf(self, value: types.Int, range: types.Int, shape=()):
    count = 1
    for d in shape:
      count *= d
    return -log_binomial_coefficient(range, count)
```

sritchie:
@femtomc if we want this, boom, here it is with tests. I don't know if that's the right logpdf or not, but let me know.

### DIFF
--- a/src/genjax/_src/generative_functions/distributions/distribution.py
+++ b/src/genjax/_src/generative_functions/distributions/distribution.py
@@ -48,10 +48,12 @@ from genjax._src.core.interpreters.staging import Flag, staged_check
 from genjax._src.core.pytree import Closure, Pytree
 from genjax._src.core.typing import (
     Any,
+    Array,
     ArrayLike,
     Callable,
     Generic,
     PRNGKey,
+    Sequence,
 )
 
 #####
@@ -569,7 +571,7 @@ class ExactDensityFromCallables(Generic[R], ExactDensity[R]):
     def sample(self, key, *args) -> R:
         return self.sampler(key, *args)
 
-    def logpdf(self, v, *args) -> Score:
+    def logpdf(self, v: R, *args) -> Score:
         return self.logpdf_evaluator(v, *args)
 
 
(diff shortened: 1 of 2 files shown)


src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -584,3 +586,40 @@ def exact_density(
         logpdf = Pytree.partial()(logpdf)
 
     return ExactDensityFromCallables[R](sample, logpdf)
+
+
```
femtomc, line 590: I would recommend putting this somewhere else (outside of the `distribution.py` file -- which I'm preferring we keep to the abstract distribution definitions)

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -584,3 +586,40 @@ def exact_density(
...
+def log_binomial_coefficient(x: ArrayLike, y: ArrayLike) -> Array:
+    return -jnp.log(x + 1) - jax.scipy.special.betaln(y + 1, x - (y + 1))
+
+
+@Pytree.dataclass
+class RandomChoice(ExactDensity[Array]):
```
femtomc, line 596: I would recommend putting this somewhere else (outside of the distribution.py file -- which I'm preferring we keep to the abstract distribution definitions)

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -584,3 +586,40 @@ def exact_density(
         logpdf = Pytree.partial()(logpdf)
 
     return ExactDensityFromCallables[R](sample, logpdf)
+
+
+def log_binomial_coefficient(x: ArrayLike, y: ArrayLike) -> Array:
```
derifatives, line 591: I'm not sure this is correct as written.

Stating my assumptions in case they're wrong:
1. You are returning an *ordered* sample: if your elements are [0, ..., 10] and your sample is size 3, [1, 4, 7] and [1, 7, 4] are *different* outcomes. You're asking for the log prob of drawing these elements in this order, *not* this set of elements. (It would also be totally feasible to build a version that returned an (unordered) set, I'm just assuming that's not what you want to do.) This is probably worth stating more clearly in the documentation (it's possible I missed it).
2. Given this, the number of possible samples is x! / y!, and all possible samples are equally likely, so you're looking to compute log(y! / x!).
3. Given the above, I think the correct expression is just `jax.lax.lgamma(y+1) - jax.lax.lgamma(x+1)`

Other thoughts:
- I would suggest naming the arguments n and k; selecting k objects from a set of n feels more canonical than selecting y objects for a set of x.
- I would suggest a separate test for `log_binomial_coefficient`, grounded out against something like:
```
def factorial(n):
  if n == 1:
     return 1
  else:
     return n * factorial(n-1)

def num_samples_without_replacement(n, k):
  return factorial(n) / factorial(k)
  ```
(which is slow and not numerically stable, but that's fine.)

PR #1322, opened 2024-09-07
Remove Sample and Constraint inheritance for ChoiceMap / Selection. Add `edit` & `generate` interfaces. (1/N)

### LINKED ISSUE #1230: [BUG] fix math behind update_change_target_from_false in mask.py
Here was the broken, old implementation:

```python
    @typecheck
    def update_change_target_from_false(
        self,
        key: PRNGKey,
        trace: Trace,
        update_problem: UpdateProblem,
        argdiffs: Argdiffs,
    ) -> tuple[Trace, Weight, Retdiff, UpdateProblem]:
        check = Diff.tree_primal(argdiffs)[0]
        check_diff, inner_argdiffs = argdiffs[0], argdiffs[1:]

        inner_trace = EmptyTrace(self.gen_fn)

        assert isinstance(update_problem, Constraint)
        imp_update_problem = ImportanceProblem(update_problem)

        premasked_trace, w, _, _ = self.gen_fn.update(
            key, inner_trace, GenericProblem(tuple(inner_argdiffs), imp_update_problem)
        )

        _, _, retdiff, bwd_problem = self.gen_fn.update(
            key, premasked_trace, GenericProblem(tuple(inner_argdiffs), update_problem)
        )

        w = jax.lax.select(
            check,
            premasked_trace.get_score(),
            0.0,
        )

        return (
            MaskTrace(self, premasked_trace, check),
            w,
            Mask.maybe(check_diff, retdiff),
            MaskedProblem(check, bwd_problem),
        )
```

### LINKED ISSUE #1312: [BUG] backward problems are broken (more info coming)
Here is the smoking gun: https://github.com/probcomp/genjax/blob/main/src/genjax/_src/generative_functions/static.py#L551

More info coming soon once we finish up pairing. I think this affects every call to `ChoiceMap.entry` supplied with an `UpdateProblem` that is NOT a `ChoiceMap`.

femtomc:
Then, fix the holes, and define overloads so users don't notice the difference.

TODO:
- [ ] Can we change the signature of mask combinator in a follow up PR?
- [ ] Instead of a type assertion for checking equality between a trace's generative function and self, could we check for tr.get_gen_fn() is self ?

Here are the large scale changes which have occurred:
* `ChoiceMap` no longer inherits from both `Sample` and `Constraint` -- only from `Sample`
* `Selection` inherits from a new type called `Projection`, which is a new concept -- denoting a way to break down a `Sample` into smaller pieces (if allowed for that `Sample`).
* Introduced `ChoiceMapConstraint`, a type to cast a `ChoiceMap` into `Constraint` (in general, old GenJAX users won't really need to care about this type -- as we've added derived interfaces which retain the interfaces where behaviors can be specified just via `ChoiceMap`, without requiring wrapping in `ChoiceMapConstraint`)
* Added 3 new methods to generative functions: `generate`, `project`, and `edit`. `generate` and `project` are actually specializations of `edit`, but we keep them separate to make the implementation more concise and readable -- to avoid introducing extra concepts like `EmptyTrace`. These new methods are used to implement `importance` and `update` -- whose signatures are specialized for people working with generative functions whose sample space is `ChoiceMap`. The `generate`, `project`, and `edit` methods are _more general_ than the derived interfaces -- they are generic over the types `Constraint`, `Projection`, and `EditRequest`. In other words, the old behavior people are used to is retained -- but now as library authors, we can expand the capabilities of the system by adding to the vocabulary of `edit` without disrupting the derived interfaces.
* The interface `update` used to accept an `UpdateProblem` (this was like v0 of `edit`) -- I've restored it to accept as signature `(key, trace, choice_map, argdiffs)`.
* Given the above changes, I've propagated them throughout the code base -- I've focused just on getting things with `ChoiceMap` working with the above changes.
* I've made changes to the tests where signature changes caused errors -- mostly with `update`.

I've added small comments in several of the files to point out specific points of interest. There's a lot of spurious "I had to fix this, because I changed this signature or this type" noise in this PR -- but hopefully my summary above and the comments below allow a concise summarized picture.


src/genjax/_src/core/generative/core.py
```diff
@@ -213,37 +160,154 @@ class MaskedConstraint(Constraint):
     flag: Flag
     constraint: Constraint
 
+    idx: IntArray
+    constraint: list[Constraint]
```
femtomc, line 136: Removed this as unnused -- to focus on choice map constraint compatibility.

src/genjax/_src/core/generative/core.py
```diff
@@ -213,37 +160,154 @@ class MaskedConstraint(Constraint):
...
-    A `SumConstraint` encodes that one of a set of possible constraints is active _at runtime_, using a provided index.
 
-    Formally, `SumConstraint(idx: IntArray, cs: List[Constraint])` represents the constraint (`x` $\\mapsto$ `xs[idx]`, `ys[idx]`).
-    """
+###############
+# Projections #
+###############
 
-    idx: IntArray
-    constraint: list[Constraint]
```
femtomc, line 142: Introduced as a concept to represent the projection of a sample onto a subspace. Of course, with choice map samples -- projections are most conveniently specified via selections.

src/genjax/_src/core/generative/core.py
```diff
@@ -213,37 +160,154 @@ class MaskedConstraint(Constraint):
...
+        raise NotImplementedError
 
 
-class Sample(Pytree):
-    """A `Sample` is a value which can be sampled from generative functions. Samples can be scalar values, or map-like values ([`ChoiceMap`][genjax.core.ChoiceMap]). Different sample types can induce different interfaces: `ChoiceMap`, for instance, supports interfaces for accessing sub-maps and values."""
+#########################
+# Update specifications #
+#########################
+
+
```
femtomc, line 157: This PR doesn't introduce `edit` as a method on `EditRequest` (in lines with internal discussion on compositional usage & extension of new `EditRequest` types)

I'll handle that in a follow up PR.

src/genjax/_src/core/generative/core.py
```diff
@@ -213,37 +160,154 @@ class MaskedConstraint(Constraint):
...
+            case _:
+                return (
+                    problem
+                    if f.concrete_true()
+                    else EmptyRequest()
+                    if f.concrete_false()
+                    else MaskedRequest(f, problem)
+                )
+
+
```
femtomc: This `IncrementalGenericRequest` represents the edit which corresponds to the old `GenerativeFunction.update` interface.

As you'll see in `GenerativeFunction`, I've kept this method as a derived method on generative functions -- it just defers to the handling of this type of `EditRequest`

src/genjax/_src/core/generative/core.py
```diff
@@ -95,96 +99,39 @@
...
-            case _:
-                return (
-                    problem
-                    if f.concrete_true()
-                    else EmptyProblem()
-                    if f.concrete_false()
-                    else MaskedProblem(f, problem)
-                )
-
-
```
femtomc, line 141: I removed several of these types, to focus on just getting support up for `IncrementalGenericRequest` (the type of edit request which implements the old `update` interface move).

src/genjax/_src/core/generative/core.py
```diff
@@ -213,37 +160,129 @@ class MaskedConstraint(Constraint):
...
+class IncrementalGenericRequest(EditRequest):
+    argdiffs: Argdiffs
+    constraint: Constraint
+
+
+################################
+# Choice map specialized types #
+################################
+
+
```
femtomc: These are "cast" types to convert `ChoiceMap` into either the type `Constraint` or the type `Sample`.

src/genjax/_src/core/generative/core.py
```diff
@@ -213,37 +160,129 @@ class MaskedConstraint(Constraint):
...
+    ) -> ChoiceMap:
+        return ChoiceMapSample(self.choice_map.get_submap(addr))
+
+    def get_value(self) -> Any:
+        return self.choice_map.get_value()
+
+    def static_is_empty(self):
+        return self.choice_map.static_is_empty()
+
+
```
femtomc: This is a "cast" type to convert a `Selection` into a `Projection`.

src/genjax/_src/core/generative/core.py
```diff
@@ -322,46 +361,62 @@ def get_sample(self) -> Sample:
         """Return the [`Sample`][genjax.core.Sample] sampled from the distribution over samples by the generative function during the invocation which created the [`Trace`][genjax.core.Trace]."""
 
     # TODO: deprecated.
-
```
femtomc, line 325: Leaving `get_choices` for traces that wish to easily furnish `ChoiceMap`, without using the new types.

src/genjax/_src/core/generative/core.py
```diff
@@ -322,46 +361,62 @@ def get_sample(self) -> Sample:
...
+    def edit(
+        self,
+        key: PRNGKey,
+        problem: EditRequest,
+    ) -> tuple[Self, Weight, Retdiff[R], EditRequest]:
+        """
+        This method calls out to the underlying [`GenerativeFunction.edit`][genjax.core.GenerativeFunction.edit] method - see [`EditRequest`][genjax.core.EditRequest] and [`edit`][genjax.core.GenerativeFunction.edit] for more information.
+        """
+        return self.get_gen_fn().edit(key, self, problem)  # pyright: ignore[reportReturnType]
+
```
femtomc: The `Trace.update` method of course defers its implementation to `GenerativeFunction.update` -- but that further defers its implementation to `GenerativeFunction.edit` (which implements `update` via the `IncrementalGenericRequest` edit request).

src/genjax/_src/core/generative/core.py
```diff
@@ -322,46 +361,62 @@ def get_sample(self) -> Sample:
...
+                Diff.tree_diff_no_change(self.get_args()),
+            )  # pyright: ignore[reportReturnType]
         else:
             return self.get_gen_fn().update(
                 key,
                 self,
-                GenericProblem(Diff.tree_diff_no_change(self.get_args()), problem),
+                constraint,
+                argdiffs,
             )  # pyright: ignore[reportReturnType]
```
femtomc: `Trace.project` supports this type overload in `projection` to support people just using `Selection`, instead of the new `Projection` type (which works on more generic `Sample` types).

src/genjax/_src/core/generative/core.py
```diff
@@ -379,32 +434,6 @@ def batch_shape(self):
         return len(self.get_score())
```
femtomc: Yay -- I was able to eliminate `EmptyTrace` entirely, but had to move logic into a `GenerativeFunction` interface method to do it -- the interface method is `GenerativeFunction.generate` (named after Gen.jl's `generate`) -- this is the general `Constraint` version of `GenerativeFunction.importance` -- `GenerativeFunction.importance` stays and works natively with `ChoiceMap` (again, for those who don't care about `Sample`, `Constraint`, the new types, etc...)

src/genjax/_src/core/generative/core.py
```diff
@@ -541,16 +570,77 @@ def model():
...
+
+
+            sample = C.kw(v1=1.0, v2=0.0)
+            score, retval = model.assess(sample, ())
+            print((score, retval))
+            ```
+        """
+        raise NotImplementedError
+
+    @abstractmethod
```
femtomc: This is `generate` -- it's just like `importance`, but it accepts a generic `Constraint` (need not be specialized to `ChoiceMap`).

src/genjax/_src/core/generative/core.py
```diff
@@ -606,108 +695,82 @@ def model(var):
...
-            print((score, retval))
-            ```
-
-            But it also works with generative functions that sample from spaces with more structure:
-
-            ```python exec="yes" html="true" source="material-block" session="core"
-            from genjax import gen
-            from genjax import normal
-            from genjax import ChoiceMapBuilder as C
-
```
femtomc, line 685: Now, I've added `importance` and `update` as derived interfaces -- they work natively with choice map (but they rely on `generate` and `edit` internally, which are the more general methods which `GenerativeFunction` implementors actually implement)

src/genjax/_src/core/generative/core.py
```diff
@@ -606,108 +695,82 @@ def model(var):
...
-                v1 = normal(0.0, 1.0) @ "v1"
-                v2 = normal(v1, 1.0) @ "v2"
-
-
-            sample = C.kw(v1=1.0, v2=0.0)
-            score, retval = model.assess(sample, ())
-            print((score, retval))
-            ```
-        """
-        raise NotImplementedError
```
femtomc, line 698: Here is the derived `update` interface -- just defers to a call to `GenerativeFunction.edit` with an `IncrementalGenericRequest`.

src/genjax/_src/core/interpreters/staging.py
```diff
@@ -207,11 +207,21 @@ def get_trace_shape(gen_fn, args):
     return get_data_shape(gen_fn.simulate)(key, args)
```
femtomc, line 213: Added some of these staged ("trace and get the return type shape") methods for `generate` and `edit`.

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -126,103 +127,89 @@ def simulate(
...
+        #               w' = w
+        #
+        # In any case, we always apply the move... we're not avoiding
+        # that computation
+
+        check = trace.check
+        final_trace = jtu.tree_map(
+            lambda v1, v2: jnp.where(check_arg.f, v1, v2), edited_trace, inner_trace
+        )
+        inner_chm = bwd_move.constraint.choice_map
```
femtomc: The code here is disgusting -- I just wrote the math above directly into this big long computation for the weight.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -80,7 +80,7 @@ def __getitem__(
 """Deprecated! please use `Selection.at`."""
```
femtomc, line 82: ChoiceMap is a `Sample`, and `Selection` is a `Projection` that operates on `ChoiceMap`.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1549,3 +1555,25 @@ def get_submap(self, addr: ExtendedAddressComponent) -> ChoiceMap:
         submap = self.c.get_submap(addr)
         subselection = self.selection(addr)
         return submap.filter(subselection)
+
+
+################################
+# Choice map specialized types #
+################################
+
+
```
femtomc, line 1561: We need only add `ChoiceMapConstraint` to cast a `ChoiceMap` to the `Constraint` type.

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -0,0 +1,1554 @@
+# Copyright 2024 MIT Probabilistic Computing Project
```
femtomc, line 1: I moved traces and generative functions into this file to avoid cyclic dependencies in imports.

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -0,0 +1,1554 @@
...
+                v2 = normal(v1, 1.0) @ "v2"
+
+
+            sample = C.kw(v1=1.0, v2=0.0)
+            score, retval = model.assess(sample, ())
+            print((score, retval))
+            ```
+        """
+        raise NotImplementedError
+
```
femtomc, line 386: This is a new method (not a new name!) -- it's the `Constraint` generic version of `importance`

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -0,0 +1,1554 @@
...
+    @abstractmethod
+    def generate(
+        self,
+        key: PRNGKey,
+        constraint: Constraint,
+        args: Arguments,
+    ) -> tuple[Trace[R], Weight]:
+        raise NotImplementedError
+
```
femtomc, line 395: This is `project` -- it lives off generative functions now. Traces defer their implementation of `project` to the logic inside the generative function.

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -0,0 +1,1554 @@
...
+        **Additional notes on [`Argdiffs`][genjax.core.Argdiffs]**
+
+        Argument changes induce changes to the distribution over samples, internal K and L proposals, and (by virtue of changes to $P$) target distributions. The [`Argdiffs`][genjax.core.Argdiffs] type denotes the type of values attached with a _change type_, a piece of data which indicates how the value has changed from the arguments which created the trace. Generative functions can utilize change type information to inform efficient [`edit`][genjax.core.GenerativeFunction.edit] implementations.
+        """
+        raise NotImplementedError
+
+    ######################
+    # Derived interfaces #
+    ######################
+
```
femtomc, line 524: `update` is a derived interface -- it's `edit`, provided with an `IncrementalGenericRequest` (todo on naming)

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -0,0 +1,1554 @@
...
+            IncrementalGenericRequest(
+                argdiffs,
+                constraint
+                if isinstance(constraint, Constraint)
+                else ChoiceMapConstraint(constraint),
+            ),
+        )
+        assert isinstance(bwd, IncrementalGenericRequest), type(bwd)
+        return tr, w, rd, bwd.constraint
+
```
femtomc, line 544: `importance` is just `generate` with a `ChoiceMap` provided as the constraint. We then lift it into a `ChoiceMapConstraint` under the hood, and call `generate`.

src/genjax/_src/core/interpreters/staging.py
```diff
@@ -211,11 +211,21 @@ def get_trace_shape(gen_fn, args):
...
```
femtomc, line 211: In this file, I added "trace and get the output shape" methods for both `generate` and `update`.

src/genjax/_src/generative_functions/combinators/dimap.py
```diff
@@ -167,28 +188,26 @@ def closed_mapping(args: tuple[Any, ...], retval: R) -> S:
...
-    ) -> tuple[DimapTrace[R, S], Weight, Retdiff[S], UpdateProblem]:
-        match update_problem:
-            case GenericProblem(argdiffs, subproblem):
-                return self.update_change_target(key, trace, subproblem, argdiffs)
-            case _:
-                return self.update_change_target(
-                    key, trace, update_problem, Diff.no_change(trace.get_args())
-                )
+        edit_request: EditRequest,
+    ) -> tuple[DimapTrace[R, S], Weight, Retdiff[S], EditRequest]:
```
femtomc, line 199: Right now we only support `IncrementalGenericRequest` -- that's fine, we'll add more in the future.

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -207,32 +222,26 @@ def update_change_target(
             # that computation.
         )
```
femtomc, line 221: This (and elsewhere in the PR!) we are assuming that -- anytime you use an `IncrementalGenericRequest` with `edit`, the backward request (the request that would "put the trace back into its initial state") is also an `IncrementalGenericRequest`

I'd like to figure out a way for us to state these assumptions (and there will be others, for other edit requests) -- but saving for a future PR.

src/genjax/_src/generative_functions/combinators/vmap.py
```diff
@@ -151,7 +161,12 @@ def find_axis_size(axis, x):
                     return leaves[0].shape[axis]
             return ()
 
-        axis_sizes = jax.tree_util.tree_map(find_axis_size, self.in_axes, args)
```
femtomc, line 154: This change fixes a deprecation warning emitted by JAX.

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -118,17 +116,17 @@ def simulate(
         tr = DistributionTrace(self, args, v, w)
         return tr
 
-    def importance_choice_map(
```
femtomc, line 121: In several other places in this PR, I essentially replace the underlying logic of `importance` as a specialization of `edit` ... with `generate`.

This is nice -- because it cleans up the signature & implementor concerns.

Also! It lets us kill `EmptyTrace` entirely.

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -349,128 +279,68 @@ def _false_branch(key, _, old_value: R):
...
+                            Diff.tree_diff_unknown_change(trace.get_args()),
+                            ChoiceMapConstraint(old_choices.mask(flag)),
+                        ),
                     )
                 else:
                     raise Exception(
                         "Only `FilteredChm` is currently supported for dynamic flags."
                     )
 
             case _:
```
femtomc, line 291: We're still just raising trace time exceptions in this PR -- not doing anything fancy with typing.

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -0,0 +1,1554 @@
...
+    # Derived interfaces #
+    ######################
+
+    def update(
+        self,
+        key: PRNGKey,
+        trace: Trace[R],
+        constraint: ChoiceMap | Constraint,
+        argdiffs: Argdiffs,
+    ) -> tuple[Trace[R], Weight, Retdiff[R], Constraint]:
```
femtomc, line 531: Also note here how the signature of `update` changes -- doesn't involve `EditRequest` at all.

tests/core/generative/test_core.py
```diff
@@ -29,7 +30,9 @@ def f():
             return y
 
         tr = f.simulate(jax.random.PRNGKey(0), ())
```
femtomc, line 32: @sritchie In general, it's a bad idea to assert comparisons with raw numbers... our code is stochastic, so even an internal change to how the random seeds work in JAX could then break out tests.

tests/generative_functions/test_distributions.py
```diff
@@ -42,7 +41,7 @@ def test_importance(self):
         # Constraint, mask with True flag.
         (tr, w) = genjax.normal.importance(
             key,
-            MaskedConstraint(True, C.v(1.0)),
```
femtomc, line 45: I removed `MaskedConstraint` entirely -- in favor of just getting everything working with `ChoiceMap` (and the constraint wrapper `ChoiceMapConstraint`)

src/genjax/_src/core/generative/generative_function.py
femtomc: Later on, we'll also add `regenerate` (it's an `edit`, provided with an `IncrementalRegenerateRequest` -- using `Projection` to pick which random variables to re-propose)

notebooks/cookbook/expressivity/dynamic_computation.ipynb
```diff
@@ -451,8 +451,8 @@
     "    key, subkey = jax.random.split(key)\n",
     "    new_circle_mask = create_circle_mask(radius=10 * i)\n",
     "    arg_diff = (genjax.Diff(new_circle_mask, genjax.UnknownChange),)\n",
-    "    constraints = C.n()\n",
-    "    update_problem = genjax.GenericProblem(arg_diff, constraints)\n",
+    "    constraints = genjax.ChoiceMapConstraint(C.n())\n",
```
sritchie, line 466: thinking out loud about `.to_constraint()` on ChoiceMap... I appreciate that the type system will tell people if they get these things wrong

notebooks/cookbook/inference/mcmc.ipynb
```diff
@@ -91,7 +91,7 @@
     "    model_args = trace.get_args()\n",
     "\n",
     "    # The core computation is updating a trace, and for that we will call the model's update method.\n",
-    "    # The update method takes a random key, a trace, and a GenericProblem object.\n",
+    "    # The update method takes a random key, a trace, and a IncrementalGenericRequest object.\n",
```
sritchie: we should update this comment, because that's not in fact what we're passing, we're just passing argdiffs.

sritchie: AND the fwd_choices?

src/genjax/_src/core/generative/choice_map.py
sritchie: This is not a pipe

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -233,6 +233,12 @@ def mask(self, flag: Flag) -> "Selection":
         """
         return MaskSel.build(self, flag)
 
+    def project(self, sample: "ChoiceMap") -> "ChoiceMap":
+        return self.filter(sample)
+
+    def complement(self) -> "Selection":
```
sritchie, line 236: why add this one vs forcing ~? It's fine of course, but want to add a test that these are the same? Can't let me coverage go down

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -233,6 +233,12 @@ def mask(self, flag: Flag) -> "Selection":
         """
         return MaskSel.build(self, flag)
 
+    def project(self, sample: "ChoiceMap") -> "ChoiceMap":
+        return self.filter(sample)
```
sritchie: test please! we can add to the existing tests and show that these are the same. Perfect generative test...

sritchie: For when we have those

src/genjax/_src/core/generative/choice_map.py
sritchie: not cast, just wrap, yeah?

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1549,3 +1555,25 @@ def get_submap(self, addr: ExtendedAddressComponent) -> ChoiceMap:
...
         return submap.filter(subselection)
+
+
+################################
+# Choice map specialized types #
+################################
+
+
+@Pytree.dataclass
+class ChoiceMapConstraint(Constraint, ChoiceMap):
```
sritchie, line 1563: I still wonder if we can make this NOT a choicemap anymore, and force users to go get the choice_map out. that would make me happier

src/genjax/_src/core/generative/generative_function.py
sritchie: when I think of a good name for this returned object we're making a proper class vs a tuple

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -0,0 +1,1554 @@
...
+        args: Arguments,
+    ) -> tuple[Sample, Score, R]:
+        """
+        Samples a [`Sample`][genjax.core.Sample] and any untraced randomness $r$ from the generative function's distribution over samples ($P$), and returns the [`Score`][genjax.core.Score] of that sample under the distribution, and the `R` of the generative function's return value function $f(r, t, a)$ for the sample and untraced randomness.
+        """
+        tr = self.simulate(key, args)
+        sample = tr.get_sample()
+        score = tr.get_score()
+        retval = tr.get_retval()
+        return sample, score, retval
```
sritchie, line 606: oopsies, test please?

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -0,0 +1,1554 @@
...
+
+    def project(
+        self,
+        key: PRNGKey,
+        trace: Trace[Any],
+        projection: Projection[Any],
+    ):
+        raise NotImplementedError
+
+    def edit(
```
sritchie, line 1504: here we goooo

src/genjax/_src/core/generative/core.py
```diff
@@ -13,56 +13,35 @@
...
 from abc import abstractmethod
-from typing import TYPE_CHECKING
-
-import jax
-import jax.numpy as jnp
-from penzai.core import formatting_util
 
 from genjax._src.core.interpreters.incremental import Diff
-from genjax._src.core.interpreters.staging import FlagOp, get_trace_shape
+from genjax._src.core.interpreters.staging import Flag
```
sritchie, line 18: is there now a better name than `core` for this file?

src/genjax/_src/core/interpreters/staging.py
```diff
@@ -211,11 +211,21 @@ def get_trace_shape(gen_fn, args):
...
 
 
-def get_update_shape(gen_fn, tr, problem):
+def get_edit_shape(gen_fn, tr, problem):
     key = jax.random.PRNGKey(0)
-    return get_data_shape(gen_fn.update)(key, tr, problem)
+    return get_data_shape(gen_fn.edit)(key, tr, problem)
+
+
+def get_importance_shape(gen_fn, choice_map, args):
```
sritchie, line 224: add types please (input and output)

src/genjax/_src/core/interpreters/staging.py
```diff
@@ -211,11 +211,21 @@ def get_trace_shape(gen_fn, args):
...
 
-def get_importance_shape(gen_fn, constraint, args):
+def get_generate_shape(gen_fn, constraint, args):
     key = jax.random.PRNGKey(0)
-    return get_data_shape(gen_fn.importance)(key, constraint, args)
+    return get_data_shape(gen_fn.generate)(key, constraint, args)
 
 
-def get_update_shape(gen_fn, tr, problem):
+def get_edit_shape(gen_fn, tr, problem):
```
sritchie, line 219: missing types

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -52,18 +51,20 @@ class MaskTrace(Generic[R], Trace[Mask[R]]):
     inner: Trace[R]
     check: ScalarFlag
 
-    def get_args(self) -> tuple[Flag, Any]:
+    def get_args(self) -> tuple[Any, ...]:
```
sritchie, line 53: whoops, put Flag back as the first element

sritchie: I think this should return a 2-tuple.

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -52,18 +51,20 @@ class MaskTrace(Generic[R], Trace[Mask[R]]):
...
         return self.mask_combinator
 
-    def get_sample(self):
-        inner_sample = self.inner.get_sample()
-        if isinstance(inner_sample, ChoiceMap):
-            return inner_sample.mask(self.check)
-        else:
-            return MaskedSample(self.check, self.inner.get_sample())
+    def get_sample(self) -> ChoiceMap:
+        inner_chm_sample = self.inner.get_sample()
```
sritchie: just call get_choices() here for now?

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -207,32 +222,26 @@ def update_change_target(
...
-                return self.update_change_target(key, trace, subproblem, argdiffs)
-            case _:
-                return self.update_change_target(
-                    key, trace, update_problem, Diff.no_change(trace.get_args())
-                )
-
     def assess(
         self,
-        sample: ChoiceMap,
+        sample: Sample,
```
sritchie: for now, imo we should keep this as ChoiceMap.

src/genjax/_src/generative_functions/combinators/scan.py
```diff
@@ -63,7 +62,12 @@ def get_args(self) -> tuple[Any, ...]:
...
 
-    def get_sample(self):
+    def get_choices(self) -> ChoiceMap:
+        return jax.vmap(
+            lambda idx, subtrace: ChoiceMap.entry(subtrace.get_choices(), idx),
+        )(jnp.arange(self.inner.get_score().shape[0]), self.inner)
+
+    def get_sample(self) -> ChoiceMap:
         return jax.vmap(
             lambda idx, subtrace: ChoiceMap.entry(subtrace.get_sample(), idx),
```
sritchie: I think this is busted, the sample is going to be wrapped in a `ChoiceMap.value` if it's not a `ChoiceMap`, not what we want, right? with `Sample` having no structure I don't think there is a way to implement this for general samples.

src/genjax/_src/generative_functions/combinators/scan.py
```diff
@@ -226,49 +219,49 @@ def _inner(
...
+            carry: tuple[PRNGKey, IntArray, Carry],
+            scanned_over: Any,
         ) -> tuple[
             tuple[PRNGKey, IntArray, Carry],
-            tuple[Trace[tuple[Carry, Y]], Y, Score, Weight, UpdateProblem],
+            tuple[Trace[tuple[Carry, Y]], Y, Score, Weight],
         ]:
             key, idx, carried_value = carry
             key = jax.random.fold_in(key, idx)
             submap = constraint.get_submap(idx)
```
sritchie, line 252: this is where I want you to pull out the choicemap and call `get_submap` on it, vs having `ChoiceMapConstraint` be a ChoiceMap (breaking our current algebraic data type status!!)

src/genjax/_src/generative_functions/combinators/scan.py
```diff
@@ -226,49 +219,49 @@ def _inner(
...
-            (carried_out, score), (tr, scanned_out, w, inner_bwd_problem) = (
-                _inner_importance(key, submap, carried_value, scanned_over)
+            assert isinstance(submap, ChoiceMapConstraint)
```
sritchie, line 253: no need for this assertion if you get the choicemap out yourself and call get_submap

src/genjax/_src/generative_functions/combinators/switch.py
```diff
@@ -77,24 +76,25 @@ class SwitchTrace(Generic[R], Trace[R]):
...
         return self.args
 
-    def get_sample(self) -> Sample:
+    def get_choices(self) -> ChoiceMap:
+        (idx, *_) = self.get_args()
+        sub_chms = list(map(lambda v: v.get_choices(), self.subtraces))
+        chm = ChoiceMap.empty()
+        for _idx, _chm in enumerate(sub_chms):
+            assert isinstance(_chm, ChoiceMap)
+            masked_submap = _chm.mask(jnp.all(_idx == idx))
```
sritchie, line 85: take away the `jnp.all`; then if you DO have a concrete idx, not a jnp.array(idx), you can optimize away a bunch of the choicemap structure.

src/genjax/_src/generative_functions/combinators/switch.py
```diff
@@ -77,24 +76,25 @@ class SwitchTrace(Generic[R], Trace[R]):
...
+            chm = chm ^ masked_submap
+        return chm
+
+    def get_sample(self) -> ChoiceMap:
         subsamples = list(map(lambda v: v.get_sample(), self.subtraces))
```
sritchie: just call `get_choices()`, the assertion shows that this is all we support for now.

src/genjax/_src/generative_functions/combinators/switch.py
```diff
@@ -244,82 +244,250 @@ def _inner(idx: int):
...
+                constraint,
+                branch_args,
+            )
+            empty_trace = jtu.tree_map(
+                lambda v: jnp.zeros(v.shape, v.dtype), trace_shape
+            )
+            trace_leaf, trace_def = jtu.tree_flatten(empty_trace)
+            retval_leaf, retval_def = jtu.tree_flatten(empty_trace.get_retval())
+            retval_defs.append(retval_def)
+            retval_leaves.append(retval_leaf)
```
sritchie, line 313: gnarly

src/genjax/_src/generative_functions/combinators/switch.py
```diff
@@ -244,82 +244,250 @@ def _inner(idx: int):
...
+            constraint,
+            branch_args,
+        )
+        subtraces = list(
+            map(
+                lambda x: jtu.tree_unflatten(trace_defs[x], trace_leaves[x]),
+                range(len(trace_leaves)),
+            )
+        )
+        retvals = list(
```
sritchie, line 390: do you need these list wrappers?

src/genjax/_src/generative_functions/combinators/switch.py
```diff
@@ -402,23 +570,23 @@ def update_dispatch(static_idx: int):
...
-                    argdiffs: self._generic_update_idx_change(
+                    argdiffs: self._generic_edit_idx_change(
                         key, static_idx, trace, problem, idx, argdiffs
                     )
                 )
 
         primals = Diff.tree_primal(argdiffs)
         idx = primals[0]
-        branch_functions = list(map(update_dispatch, range(len(self.branches))))
+        branch_functions = list(map(edit_dispatch, range(len(self.branches))))
```
sritchie, line 572: my eyes are glazing over, tapping @littleredcomputer in for careful reading

src/genjax/_src/generative_functions/combinators/vmap.py
```diff
@@ -63,14 +65,22 @@ def get_retval(self):
     def get_gen_fn(self):
         return self.gen_fn
 
-    def get_sample(self):
+    def get_sample(self) -> ChoiceMap:
```
sritchie, line 67: how does this pass typechecking? also, call `get_choices() as this is busted and will wrap a non-choicemap sample in `ChoiceMap.value`

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -85,6 +80,9 @@ def get_score(self) -> Score:
     def get_sample(self) -> ChoiceMap:
```
sritchie, line 79: call get_choices?

src/genjax/_src/generative_functions/distributions/distribution.py
sritchie: Nice!!

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -349,128 +279,68 @@ def _false_branch(key, _, old_value: R):
...
-
-    def update_change_target(
+        projection: Projection[Any],
+    ) -> Weight:
+        assert isinstance(projection, Selection)
+        return jnp.where(
+            projection.check(),
+            trace.get_score(),
+            jnp.array(0.0),
+        )  # pyright: ignore
```
sritchie: woah buddy what's this ignore about??

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -537,9 +407,10 @@ def estimate_logpdf(
 
     def assess(
         self,
-        sample: ChoiceMap,
+        sample: Sample,
```
sritchie: re-add ChoiceMap

tests/core/generative/test_core.py
sritchie: thanks for fixing

tests/core/test_staging.py
```diff
@@ -41,9 +43,9 @@ def model():
 
         key = PRNGKey(0)
         trace = model.simulate(key, ())
-        new_trace, _w, _rd, bwd_problem = get_update_shape(model, trace, U.g((), C.n()))
+        new_trace, _w, _rd, bwd_request = get_update_shape(model, trace, C.n(), ())
         assert isinstance(new_trace.get_sample(), Chm)
-        assert isinstance(bwd_problem, Chm)
+        assert isinstance(bwd_request, Chm)
```
sritchie, line 48: great character savings

tests/generative_functions/test_distributions.py
sritchie: `.mask(True)`  is a no-op, this won't do anything (since we know `True` concretely)

tests/generative_functions/test_distributions.py
```diff
@@ -52,7 +51,7 @@ def test_importance(self):
         # Constraint, mask with False flag.
         (tr, w) = genjax.normal.importance(
             key,
-            MaskedConstraint(False, C.v(1.0)),
+            C.v(1.0).mask(False),
```
sritchie: This is equal to `C.n()`, masking to concrete false returns `ChoiceMap.empty()`

tests/generative_functions/test_distributions.py
```diff
@@ -132,10 +131,8 @@ def test_update(self):
         (new_tr, w, _, _) = genjax.normal.update(
             sub_key,
             tr,
-            U.g(
-                (Diff(0.0, NoChange), Diff(1.0, NoChange)),
-                MaskedConstraint(True, C.v(1.0)),
-            ),
+            C.v(1.0).mask(True),
```
sritchie: no-op

tests/generative_functions/test_mask_combinator.py
```diff
@@ -61,49 +60,37 @@ def test_mask_update_weight_to_argdiffs_from_true(self, key):
...
-        pre_weight = tr.update(key, argdiffs)[1]
-        post_weight = tr.inner.update(
-            key, U.g(Diff.tree_diff_unknown_change((3.0,)), C.n())
-        )[1]
-        assert pre_weight == post_weight
-
+        argdiffs = (Diff.unknown_change(True), Diff.no_change(tr.get_args()[1]))
+        w = tr.update(key, C.n(), argdiffs)[1]
+        assert w == tr.inner.update(key, C.n())[1]
+        assert w == 0.0
```
sritchie, line 66: this is the old test, please re-add the one we wrote with @MathieuHuot

sritchie (changes requested):

tests/generative_functions/test_mask_combinator.py
femtomc: wait -- I kept both tests from main and my branch...

did i clobber something?

notebooks/cookbook/expressivity/dynamic_computation.ipynb
femtomc: As of this PR, people using the old interfaces don't have to worry about doing this, it only occurs here because Mathieu explicitly used the new interface -- but could be beneficial as we build towards the future.

src/genjax/_src/core/generative/choice_map.py
femtomc: it's not users who suffer from this choice, it's us -- in our internal implementation of generative functions -- you'd then need to (a) unwrap the ChoiceMapConstraint -- (b) call `get_submap` on the inside choice map and (c) wrap it back up.

femtomc: But that sounds like it could perfectly fit in `get_submap`...

src/genjax/_src/core/interpreters/staging.py
femtomc: we can't add types here, even with `TYPE_CHECKING`, because then `beartype` will complain about forward references.

src/genjax/_src/core/interpreters/staging.py
femtomc: we can't add types here, even with TYPE_CHECKING, because then beartype will complain about forward references.

src/genjax/_src/generative_functions/combinators/mask.py
femtomc: Is the suggestion here to remove `get_sample`? I'm fine to do that for now, if that's what you mean

tests/generative_functions/test_distributions.py
femtomc: Is the correct thing `.mask(Flag(True))`? or `.mask(jnp.array(True))`?

src/genjax/_src/core/generative/choice_map.py
femtomc: missed joke -- or explain further if you want changes

src/genjax/_src/core/generative/core.py
femtomc: concepts?

src/genjax/_src/generative_functions/combinators/mask.py
femtomc: Can we change the signature of mask combinator in a follow up PR?

src/genjax/_src/generative_functions/combinators/scan.py
femtomc: not until we make choice map a generic collection ... fixing now

src/genjax/_src/generative_functions/distributions/distribution.py
femtomc: can remove now -- had to add when we had all the `Flag` stuff

src/genjax/_src/core/generative/choice_map.py
femtomc: Do you mean -- why not force ~ in projection? versus naming the method `Projection.complement`?

src/genjax/_src/core/generative/choice_map.py
femtomc: I see now that you're upset about the assertions

src/genjax/_src/core/generative/choice_map.py
femtomc: renamed `Projection.project` to `Projection.filter` -- therefore, only "redundant" method on `Selection` is `Selection.complement` -- which I can add a test for coverage to resolve this conversation.

src/genjax/_src/core/generative/generative_function.py
femtomc: now you're doing real PL research work

src/genjax/_src/core/generative/choice_map.py
femtomc: @sritchie How about this -- remove `ChoiceMap` inheritance on `ChoiceMapConstraint`, add `to_constraint` to `ChoiceMap` -- do the unpacking, remove the asserts.

src/genjax/_src/generative_functions/combinators/switch.py
femtomc: switch is possibly the most impenetrable part of the codebase -- i'd recommend we dissect in a separate PR.

src/genjax/_src/core/generative/core.py
femtomc: This is handled in #1337.

tests/generative_functions/test_mask_combinator.py
sritchie: oh cool I saw this and didn't see the other, sounds good if they're both there

tests/generative_functions/test_distributions.py
sritchie: no it's good, just remove it (it's concretely true so there's no reason to call `.mask(True)`.

src/genjax/_src/core/generative/choice_map.py
sritchie: image

src/genjax/_src/core/generative/choice_map.py
sritchie: I meant that we can already call `~` on a selection, so why do we need to add a `.complement(..)` method, since I don't see it used anywhere (could be wrong)

tests/generative_functions/test_distributions.py
femtomc: Well I want to test the code paths where it’s not concretely true or concretely false

notebooks/cookbook/inference/mcmc.ipynb
femtomc: Changed the comment to reflect this.

src/genjax/_src/core/generative/choice_map.py
femtomc: It’s part of the specification for a Projection — for Projection, one needs to specify a partitioning of the sample space, and that’s what complement is there for

femtomc: We could remove complement as the name of the method and just assert that subtypes must implement the ~ dunder?

tests/generative_functions/test_distributions.py
femtomc: I added jnp.array(…) versions of these lines — hopefully this hits the not concrete code (must check)

sritchie: yup that will hit it!

notebooks/cookbook/expressivity/dynamic_computation.ipynb
```diff
@@ -451,9 +451,9 @@
     "    key, subkey = jax.random.split(key)\n",
     "    new_circle_mask = create_circle_mask(radius=10 * i)\n",
     "    arg_diff = (genjax.Diff(new_circle_mask, genjax.UnknownChange),)\n",
-    "    constraints = C.n()\n",
-    "    update_problem = genjax.GenericProblem(arg_diff, constraints)\n",
-    "    tr, _, _, _ = tr.update(key, update_problem)\n",
+    "    constraints = genjax.ChoiceMapConstraint(C.n())\n",
+    "    update_problem = genjax.IncrementalGenericRequest(arg_diff, constraints)\n",
```
littleredcomputer, line 467: Suppose I have never heard of argdiffs and don't care. Is there a one-step class I can use to just use update to get the weight of a change to traced values and nothing else?

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1549,3 +1552,25 @@ def get_submap(self, addr: ExtendedAddressComponent) -> ChoiceMap:
...
+    choice_map: ChoiceMap
```
littleredcomputer, line 1564: I would say that a ChoiceMapConstraint both "is-a" ChoiceMap (inheritance) and "has-a" ChoiceMap (composition) so I'd say pick one. The implementation of the choice map interface would naturally be delegated to the member variable, so just expose it as a method

littleredcomputer: In Golang, if you mention a struct namelessly in another struct's definition, you get all the member variables and methods of the included struct in the hosting struct for free, which appears to be what we want here, so you're not alone in wanting this pattern: this is supposed to be an enriched choice map. But it's not enriched by having any additional data, it's enriched in what way? I'm not opposed to using inheritance for this

littleredcomputer: If _any_ choice map can serve as a constraint, why did we remove Constraint from ChoiceMap's subtypes? It still leaves open the possibility of a Constraint that is _not_ a ChoiceMap which I understand is desirable.

littleredcomputer: I'm more convinced than ever. Please consider putting Constraint back in to the base classes of ChoiceMap. If there are methods of ChoiceMap that _every constraint should satisfy_ then make those methods of Constraint.  Then, those functions that can do their job on any constraint, whether or not it's a choice map, can delcare the parameter of that type.

src/genjax/_src/core/generative/core.py
littleredcomputer: Screenshot 2024-09-12 at 15 59 38

src/genjax/_src/core/generative/core.py
```diff
@@ -211,1435 +131,44 @@ class MaskedConstraint(Constraint):
...
-        if self.kwargs:
-            maybe_kwarged_gen_fn = self.get_gen_fn_with_kwargs()
-            return maybe_kwarged_gen_fn.assess(
-                sample,
-                (full_args, self.kwargs),
-            )
-        else:
-            return self.gen_fn.assess(sample, full_args)
+@Pytree.dataclass(match_args=True)
+class IncrementalGenericRequest(EditRequest):
```
littleredcomputer, line 172: IIUC, "Incremental" means "I want to tamper with the arguments, creating the possibility that you will have to re-execute parts of the model" and "Generic" means "I want to change the values of traced variables with a ChoiceMap." The name isn't wrong, but conveys little to the neophyte. Everything else seems dumb, though, like "ChoiceMapWithArgDiffs", which is less appetizing.

src/genjax/_src/core/generative/generative_function.py
littleredcomputer: If we agree that ChoiceMap is-a Constraint, then the signatures of `importance` and `generate` are the same (and I suspect the implementation is as well).

src/genjax/_src/core/generative/generative_function.py
littleredcomputer: I understand how `regenerate` compares to `importance`:  with importance, I have cooked up some new values and want to measure a weight. For regenerate, I want to _indicate_ a subset of variables, roll the dice again, and get the weight as if I had asked for importance with those values.

But does it make sense to also allow argdiffs? I would think that, argdiffs could cause different branches to be taken in the model, so the Projection would have to account for that. Maybe so. At any rate, using the original projection on the returned trace should reveal exactly those values which were recomputed?

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -0,0 +1,1554 @@
...
+        self,
+        key: PRNGKey,
+        trace: Trace[R],
+        edit_request: EditRequest,
+    ) -> tuple[Trace[R], Weight, Retdiff[R], EditRequest]:
+        match edit_request:
+            case IncrementalGenericRequest(argdiffs, subproblem):
+                full_argdiffs = (*self.args, *argdiffs)
+                if self.kwargs:
+                    maybe_kwarged_gen_fn = self.get_gen_fn_with_kwargs()
```
littleredcomputer, line 1514: should we kwarg gen fns when we create them if `self.kwargs` ?

littleredcomputer: I'm wondering if there's a way to sweep this under the rug

src/genjax/_src/generative_functions/combinators/mask.py
littleredcomputer: Random thought: what if the backward request were a thunk? Just return `lambda tr: whatever it takes` and then there's no type

src/genjax/_src/generative_functions/combinators/scan.py
littleredcomputer: A ChoiceMap is a kind of Constraint. Simple as. Not every Constraint will allow `get_submap`, so we need a ChoiceMap here, and we have one. Which is nice!

src/genjax/_src/generative_functions/combinators/scan.py
```diff
@@ -282,58 +272,81 @@ def _importance(
...
+
+        def _project(
+            carry: tuple[PRNGKey, IntArray],
+            subtrace: Trace[Any],
+        ) -> tuple[tuple[PRNGKey, IntArray], Weight]:
+            key, idx = carry
+            key = jax.random.fold_in(key, idx)
+            subprojection = projection(idx)
+            assert isinstance(subprojection, Selection)
+            w = _inner_project(key, subtrace, subprojection)
```
littleredcomputer, line 305: I'd just inline the body of _inner_project here, it's easier to read that way

src/genjax/_src/generative_functions/combinators/switch.py
```diff
@@ -244,82 +244,249 @@ def _inner(idx: int):
...
+        args: tuple[Any, ...],
+    ) -> tuple[Score, R]:
+        idx, branch_args = args[0], args[1:]
+        self.static_check_num_arguments_equals_num_branches(branch_args)
+
+        def _inner(static_idx: int):
+            return lambda sample, args: self._assess(static_idx, sample, args)
+
+        branch_functions = list(map(_inner, range(len(self.branches))))
+
```
littleredcomputer, line 275:
```suggestion
branch_functions = [lambda sample, args: self._assess(k, sample, args) for k in range(len(self.branches))]
```

src/genjax/_src/generative_functions/combinators/switch.py
```diff
@@ -244,82 +244,249 @@ def _inner(idx: int):
...
+        retval_leaves, score = jax.lax.switch(
+            idx, branch_functions, sample, branch_args
+        )
+        (_, retval_defs) = self._empty_assess_defs(sample, branch_args)
+        retvals = list(
+            map(
+                lambda x: jtu.tree_unflatten(retval_defs[x], retval_leaves[x]),
```
littleredcomputer, line 282: A comment in this function would be handy since it's not obvious why we're unflattening things

src/genjax/_src/generative_functions/combinators/switch.py
littleredcomputer: If you do, a list comprehension is preferred here

src/genjax/_src/generative_functions/combinators/switch.py
```diff
@@ -244,82 +244,249 @@ def _inner(idx: int):
...
-        bwd_problem_leaves = []
+        bwd_request_defs = []
+        bwd_request_leaves = []
         retdiff_defs = []
         retdiff_leaves = []
         for static_idx in range(len(self.branches)):
             subtrace = trace.subtraces[static_idx]
             gen_fn = self.branches[static_idx]
             branch_argdiffs = argdiffs[static_idx]
             key = jax.random.PRNGKey(0)
```
littleredcomputer, line 423: Since key is constant, these computations feel pro-forma. Can these shapes change from one invocation of _empty_edit_defs to the next? Could these shapes be computed once and then saved?  If we stop using penzai for data classing, these objects are not necessarily frozen and so some precomputation could be done. However trace is a parameter, leaving me to wonder how the results are sensitive to the trace. 

Whenever we resort to tree_flatten and tree_unflatten, it would be cool to have a comment explaining why this is necessary.

littleredcomputer: This would be less gnarly if we had a function that would do the job for one branch, and have the for loop drive that function

src/genjax/_src/generative_functions/combinators/vmap.py
```diff
@@ -173,103 +182,110 @@ def simulate(
...
         broadcast_dim_length = self._static_broadcast_dim_length(primals)
         idx_array = jnp.arange(0, broadcast_dim_length)
         sub_keys = jax.random.split(key, broadcast_dim_length)
 
         def _update(key, idx, subtrace, argdiffs):
-            subproblem = update_problem(idx)
-            new_subtrace, w, retdiff, bwd_problem = self.gen_fn.update(
-                key, subtrace, GenericProblem(argdiffs, subproblem)
+            subconstraint = constraint(idx)
+            assert isinstance(subconstraint, ChoiceMapConstraint), type(subconstraint)
```
littleredcomputer, line 238: Granting that there are other kinds of constraints beside ChoiceMapConstraints, can't we just agree that the ChoiceMap is a model for every constraint, so that we can avoid this kerfuffle? Concretely: it doesn't seem awful to require that every constraint be callable with an address segment even if it has no substructure (it may raise an exception in that case)

littleredcomputer: Like ValueChm, which is an _equality constraint_. It has get_submap which always returns empty.

src/genjax/_src/generative_functions/combinators/vmap.py
```diff
@@ -173,103 +182,110 @@ def simulate(
...
-    ) -> tuple[VmapTrace[R], Weight, Retdiff[R], UpdateProblem]:
-        match update_problem:
-            case GenericProblem(argdiffs, subproblem):
-                return self.update_change_target(key, trace, subproblem, argdiffs)
-            case _:
-                return self.update_change_target(
-                    key, trace, update_problem, Diff.no_change(trace.get_args())
-                )
+        edit_request: EditRequest,
+    ) -> tuple[VmapTrace[R], Weight, Retdiff[R], EditRequest]:
```
littleredcomputer, line 277: I'm glad that GFs and not Traces own these implementations. One feature of the trace API is that you can call update on it and it supplies self.get_gen_fn() for the GF. Instead of a type assertion, could we check for `tr.get_gen_fn() is self `? And if not, emit the error message "you must supply a trace generated by _this_ generative function in order to use this interface"?

src/genjax/_src/generative_functions/static.py
```diff
@@ -389,7 +395,92 @@ def wrapper(key, previous_trace, constraints, diffs: tuple[Any, ...]):
...
+
+    def handle_trace(
+        self,
+        addr: StaticAddress,
+        gen_fn: GenerativeFunction[Any],
+        args: tuple[Any, ...],
+    ):
+        self.visit(addr)
+        subconstraint = self.get_subconstraint(addr)
+        self.key, sub_key = jax.random.split(self.key)
```
littleredcomputer, line 451: Yikes! Do we really want this object to acquire time-varying state in this way?

src/genjax/_src/generative_functions/static.py
```diff
@@ -389,7 +395,92 @@ def wrapper(key, previous_trace, constraints, diffs: tuple[Any, ...]):
...
+# Importance request handler #
+##############################
+
+
+@dataclass
+class GenerateHandler(StaticHandler):
+    key: PRNGKey
+    choice_map_constraint: ChoiceMapConstraint
+    address_visitor: AddressVisitor = Pytree.field(default_factory=AddressVisitor)
+    score: Score = Pytree.field(default_factory=lambda: jnp.zeros(()))
```
littleredcomputer, line 415: since numpy arrays are immutable, I do not think we need a factory here, just a default value

src/genjax/_src/generative_functions/static.py
littleredcomputer: ...well, it's a handler object, so probably only used once and then thrown away... or is it? Maybe I am confusing generate and regenerate. Is the idea to produce a stream of values?

littleredcomputer: ...if we really want this, it would be way more efficient as a scan over keys, than one at a time, though you would have to say up front how many you wanted

This transform seems un-JAXy

littleredcomputer (dismissed):
I like it.

notebooks/cookbook/expressivity/dynamic_computation.ipynb
femtomc: We define a convenient way to do this via Trace.update (just accepts a key, a choice map, and optionally — argdiffs)

src/genjax/_src/core/generative/choice_map.py
femtomc: Unfortunately not any choice map can serve as a constraint — we have to think forwards to using choice map as a generic container, which may contain EditRequest (!) values at its addresses (this is not a bridge too far, it’s going to be a perfectly consistent thing that people will want to do when we unlock the edit functionality)

I started by removing Sample and Constraint with this in mind … is a Constraint a Sample? Well, sometimes yes — but sometimes no. If I use a Constraint type inside my ChoiceMap which says “constrain the generation to be within some range on the reals” — that’s not a sample, that constraint doesn’t correspond to a value that the generative function may actually sample at that address, it’s asking for the sample to satisfy some property.

and then we go up the hierarchy … what if we have an EditRequest at an address — asking a callee of the toplevel generative function to modify its subtrace in some way? Well, that’s not a constraint (doesn’t have to be — c.f. discussion about regenerate for an example of how an edit request need not contain a constraint)

For exposing the way that things used to work (backwards compatibility), there’s no cost to adding an enriched type like ChoiceMapConstraint (or, in the future, ChoiceMapEditRequest) — and based on some of my reasoning in this discussion, I think it is better to not try and pile everything into choice map … but to eventually move towards choice map as a generic collection type.

src/genjax/_src/core/generative/core.py
femtomc: The name is not good, and I change it in #1337

src/genjax/_src/core/generative/generative_function.py
femtomc: I quickly put together the kwargs functionality when I contributed it — I’m happy to work on another PR with you that really understands the design

src/genjax/_src/generative_functions/static.py
femtomc: It’s used once and then thrown away, and we can’t setup a scan because user code which consumes the keys in static gen fns is heterogeneous in computation shape.

femtomc: This will make more sense after we discuss staged JAX interpreters next week.

src/genjax/_src/generative_functions/combinators/mask.py
femtomc: like the type is a `Callable`?

notebooks/cookbook/expressivity/dynamic_computation.ipynb
femtomc: The syntax is `Trace.update(key, chm)`

src/genjax/_src/core/generative/choice_map.py
femtomc: @littleredcomputer a simple (trivial) example of this -- not relevant to this PR, but relevant to a (soon to be incoming) PR:

```python
def outer_model(v):
    return inner.scan(n=20)(v, None) @ "inner"
```

Say you want to update a specific index in the scan via an `IndexEditRequest` (denoting a specialized update -- which only looks at the specified index of the scan) -- you need some way to create an `EditRequest` which _addresses_ the inner scan.

So the natural way to do that would be to say:
```
ChoiceMapEditRequest(ChoiceMap.kw("inner" = IndexEditRequest(5, ...))
```

This is what I mean when I say "we should be moving towards a world where `ChoiceMap` is more like a generic collection type"

We may not have to fully embrace that world, but it's going to be very natural to want to do things like the above.

littleredcomputer: ... and that's a Constraint, right? To win this game, you must produce a counterexample: a ChoiceMap which is not a constraint.

My position is, that since the implementation of `Constraint` is `pass`, any object can be a constraint, merely by mentioning its name among its base classes and changing nothing else--certainly every ChoiceMap can be a Constraint too. For the example you supply, it's a Constraint, like any other object.

If you are planning to add a method to Constraint which a ChoiceMap cannot satisfy, even in principle, please name it.

I think you're saying "it's possible to build a ChoiceMap which is inappropriate for use in some contexts where Constraint is required," or "the description Constraint is not appropriate for certain ChoiceMaps". If either of those is true, and you want to enforce this discipline at the type level, you are going to have to say something about what a Constraint _is_. 

As it stands, the class of ChoiceMaps is a subclass of Constraint. Admit that, and all the assertions go away.

littleredcomputer: The fact that there are Constraints that are not ChoiceMaps--such as an inequality of a continuous variable--is arguing the opposite direction, which I agree with.

femtomc: ```ChoiceMapEditRequest(ChoiceMap.kw("inner" = IndexEditRequest(5, ...))```
is not a `Constraint`

femtomc: We don't want to confuse `Constraint` and `EditRequest` in the signature of our interfaces -- or in the logic.

We continue to have these conversations about "what the type has to satisfy" -- but we keep forgetting that we're essentially doing something which amounts to something like concepts in C++20 -- combined with template metaprogramming...

In other words, in our library, a type can correspond _to specialized forms of code generation_ -- not just its usage or interfaces. That's essentially what we use the type `Constraint` and the type `EditRequest` now -- they don't have interfaces, they denote extendable customized code generation paths which denote particular types of things (in the case of `Constraint` -- specialized codegen for importance sampling for generative functions -- in the case of `EditRequest` -- specialized codegen for SMC moves to move from trace to trace)

Is this reasoning satisfactory? We haven't talked about this clearly -- but this is the clear reason why we have `pass` types like `Constraint` and `EditRequest` -- they're specialized branches in our compiler for different types of inference logic.

We separate them at the type level because they correspond to code generation for inference logic _with different mathematical requirements / guarantees_

src/genjax/_src/core/generative/generative_function.py
femtomc: Argdiffs are required -- exactly because of the example you're thinking about: say we regenerate a random choice which is an index for a switch -- the `edit` interface for `IncrementalRegenerateRequest` has to account for this, so when we visit the switch, we need to know that we changed the branch.

The projection doesn't need to provide further information about what happens when you do this (it only needs a way to provide a subprojection -- for any branch of the switch which we go into) -- but switch itself (if it implements the `IncrementalRegenerateRequest` move) needs to provide a way to sample new values for the switched-to branch.

src/genjax/_src/generative_functions/combinators/mask.py
femtomc: @littleredcomputer a few more intuitions: the “bwd request” is not just there to reset the trace data — it represents the entire SMC move which puts the trace back.

so, in the signature of the `Callable` -- you'd need to essentially duplicate the signature of `edit` (because the bwd move needs to account for the weight, the retdiff, and then the bwd bwd move ...)

_it may be possible to make this compositional_ -- but it seems more complicated than our current setup

src/genjax/_src/generative_functions/combinators/vmap.py
femtomc: That's a good idea.

src/genjax/_src/core/generative/choice_map.py
sritchie: no, that's fine, I didn't realize it was on projection.

src/genjax/_src/core/generative/choice_map.py
sritchie: I defer to @littleredcomputer and you here as you're discussing this in other comments.

src/genjax/_src/generative_functions/combinators/mask.py
sritchie: yes, can you leave a TODO in your PR description reminding us?

src/genjax/_src/generative_functions/combinators/switch.py
sritchie: Ping on this, cc @littleredcomputer for viz and in case I'm wrong here

sritchie (commented):
Boom, just a few follow-up comments

src/genjax/_src/generative_functions/static.py
femtomc: turns out we actually need this or else XLA yells at us:
```python
ValueError: mutable default <class 'jaxlib.xla_extension.ArrayImpl'> for field score is not allowed: use default_factory
```

src/genjax/_src/generative_functions/combinators/vmap.py
femtomc: I've pulled this into a TODO for a next PR, okay?

src/genjax/_src/generative_functions/combinators/switch.py
femtomc: I'm not yet convinced that this implementation of switch is the way to go -- versus a programmatically simpler (but perhaps, less optimized) one which just uses MaskCombinator (and perhaps composes it with DimapCombinator).

femtomc: > Whenever we resort to tree_flatten and tree_unflatten, it would be cool to have a comment explaining why this is necessary.

This is very hard to explain in a way which will be clear -- sometimes you want to return Pytrees with different "static parts" but the same "dynamic parts" from a jax.lax.cond -- that's why all this annoying stuff is happening here.

sritchie (approved):

femtomc:
LOVE IT

PR #1323, opened 2024-09-08
ChoiceMap and Selection constructor optimizations (GEN-548, GEN-558)

sritchie:
In this PR:

- `ChoiceMap.{d, from_mapping, kw, entry}` will all call `ChoiceMap.d` on dict-shaped values, allowing calls like ` C["x"].set({"y": {"z": 3.0}})` to work. Same thing applies to the methods on `ChoiceMapBuilder`.
- `switch_combinator` no longer uses `jnp.all` when constructing the flags it uses for building choicemaps. In the case of a concrete index, this lets the `ChoiceMap` constructor optimizations come into play and simplify the returned choicemap.
- disallows `Ellipsis` from coming in to `ChoiceMapBuilder` or `SelectionBuilder`, as we don't have semantics for this locked down yet
- introduces `NoneSel()` so we can recognize this case explicitly vs `~Selection.all()`
- `ChoiceMap.__add__` now defers to `__or__` directly, so that `__or__` overrides for `ValueChm` work there too
- `XorChm`'s staged validation moves up to the constructor from `get_value`
- `MaskChm` is gone, in favor of `FilteredChm` with a `MaskSel` internally.
  - The check in `Distribution` for `MaskChm` now allows any `FilteredChm`
- adds `build` static methods to all selector and choicemap subclasses, so that we can introduce the following optimizations:
  - `MaskSel.build`: Returns the original selection if the flag is concretely true, or `Selection.none()` if concretely false
  - `ComplementSel.build`: Returns `Selection.none()` for `AllSel`, `Selection.all()` for `NoneSel`, and unwraps double complements
  - `StaticSel.build`: Returns the original selection if it's `NoneSel`
  - `AndSel.build`: 
    - Returns the non-`AllSel` operand if one operand is `AllSel`
    - Returns either operand if the other is `NoneSel`
    - Combines `MaskSel` operands by AND-ing their flags
  - `OrSel.build`: 
    - Returns `AllSel` if either operand is `AllSel`
    - Returns the non-`NoneSel` operand if one operand is `NoneSel`
    - Combines `MaskSel` operands by OR-ing their flags
  - `ChmSel.build`: Returns `Selection.none()` for `EmptyChm`
  - `XorChm.build`: Performs staged validation in the constructor instead of `get_value`
  - `OrChm.build`: Returns the non-empty operand if one operand is `EmptyChm`
  - `FilteredChm.build`: Returns the original ChoiceMap if the selection is `AllSel`, or `EmptyChm` if the selection is `NoneSel`
  - `StaticChm.build`: Returns `EmptyChm` if the submap is `EmptyChm`
  - `IdxChm.build`: Returns `EmptyChm` if the submap is `EmptyChm`

### DIFF
--- a/src/genjax/_src/core/generative/choice_map.py
+++ b/src/genjax/_src/core/generative/choice_map.py
@@ -1514,6 +1514,8 @@ class FilteredChm(ChoiceMap):
 
 
 # TODO clamp switch inputs?
+# TODO handle ellipsis in StaticChm correctly
+# TODO make the isinstance check first, so we can get a concrete true if possible
 
 # In [6]: ChoiceMap.value("x").mask(Flag(jnp.all(1 == 1))).mask(Flag(jnp.all(1==1)))
 # Out[6]:
--- a/src/genjax/_src/core/generative/choice_map.py
+++ b/src/genjax/_src/core/generative/choice_map.py
@@ -649,11 +649,10 @@ class ChmSel(Selection):
 
     @staticmethod
     def build(chm: "ChoiceMap") -> Selection:
-        match chm:
-            case EmptyChm():
-                return Selection.none()
-            case _:
-                return ChmSel(chm)
+        if chm.static_is_empty():
+            return Selection.none()
+        else:
+            return ChmSel(chm)
 
     def check(self) -> Flag:
         return self.c.has_value()
@@ -1289,11 +1288,10 @@ class IdxChm(ChoiceMap):
 
     @staticmethod
     def build(chm: ChoiceMap, addr: DynamicAddressComponent) -> ChoiceMap:
-        match chm:
-            case EmptyChm():
-                return chm
-            case _:
-                return IdxChm(chm, addr)
+        if chm.static_is_empty():
+            return chm
+        else:
+            return IdxChm(chm, addr)
 
     def get_value(self) -> Any:
         return None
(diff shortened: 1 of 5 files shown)


femtomc:
Love that we’re tickling 85% coverage

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -397,6 +450,18 @@ class ComplementSel(Selection):
 
     s: Selection
 
+    @staticmethod
+    def build(s: Selection) -> Selection:
```
littleredcomputer, line 454: this is pretty cool. If only there were a way to make the `__init__` private in cases like this. But this is a good design

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -850,14 +971,17 @@ def kw(**kwargs) -> "ChoiceMap":
...
 
+        Dict-shaped values are recursively converted to ChoiceMap instances with calls to `ChoiceMap.d`.
+
         Returns:
             A ChoiceMap containing the key-value pairs from the input keyword arguments.
 
         Example:
             ```python
-            kw_chm = ChoiceMap.kw(x=42, y=[1, 2, 3])
+            kw_chm = ChoiceMap.kw(x=42, y=[1, 2, 3], z={"w": 10.0})
```
littleredcomputer, line 980: nice

tests/core/test_choice_maps.py
```diff
@@ -42,3 +37,481 @@ def test_selection(self):
...
+        assert chm["x"] == 1
+        with pytest.raises(ChoiceMapNoValueAtAddress, match="y"):
+            chm["y"]
+
+    def test_contains(self):
+        chm = ChoiceMap.kw(x={"y": 1})
+        assert "x" not in chm
+        assert "y" in chm("x")
+        assert ("x", "y") in chm
+        assert "z" not in chm
```
littleredcomputer, line 517: what a sweet addition!

littleredcomputer (approved):

Issue #1314, opened 2024-09-06
implement Flag.xor_


sritchie:
Sketch:

```python
def xor_(self, f: "Flag") -> "Flag":
    # True xor X => ~X. False xor X => X.
    if self.f is True:
        return f.not_()
    if self.f is False:
        return f
    if f.f is True:
        return self.not_()
    if f.f is False:
        return self
    return Flag(jnp.logical_xor(self.f, f.f))
```

sritchie:
This will be closed by #1325.

PR #1325, opened 2024-09-09
hide Flag, by making it a class of static operations (GEN-554)

littleredcomputer:


src/genjax/_src/core/generative/choice_map.py
```diff
@@ -153,7 +157,7 @@ def all() -> "Selection":
             from genjax import Selection
 
             all_selection = Selection.all()
-            assert all_selection["any_address"].f == True
+            assert all_selection["any_address"] == True
```
sritchie, line 185: I like it so far!

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -321,12 +325,12 @@ class AllSel(Selection):
...
         ```python exec="yes" html="true" source="material-block" session="choicemap"
         all_sel = Selection.all()
-        assert all_sel["any_address"].f == True
+        assert all_sel["any_address"] == True
         ```
     """
 
     def check(self) -> Flag:
-        return Flag(True)
+        return True
```
sritchie, line 352: I can pull this into my PR easily, so good

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1268,16 +1266,16 @@ def build(
...
         staged_err(
             err_check,
             f"The disjoint union of two choice maps have a value collision:\nc1 = {self.c1}\nc2 = {self.c2}",
         )
         v1 = self.c1.get_value()
         v2 = self.c2.get_value()
 
-        def pair_flag_to_idx(first: Flag, second: Flag):
-            return first.f + 2 * second.f - 1
+        def pair_flag_to_idx(first, second):
```
sritchie: hey, where my types at

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1389,9 +1387,9 @@ def build(
         c: ChoiceMap,
         flag: Flag,
     ) -> ChoiceMap:
-        if c.static_is_empty() | flag.concrete_true():
+        if c.static_is_empty() | FlagOp.concrete_true(flag):
             return c
-        elif flag.concrete_false():
+        elif FlagOp.concrete_false(flag):
```
sritchie: nice, these are the right checks for concrete, I like it. I had wanted `is_concrete` as well...

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -57,22 +59,23 @@ class Mask(Generic[R], Pytree):
     If you use invalid `Mask(False, data)` data in inference computations, you may encounter silently incorrect results.
     """
 
-    flag: Flag
+    flag: Flag | Diff[Flag]
     value: R
 
     @staticmethod
-    def maybe(f: Flag, v: "R | Mask[R]") -> "Mask[R]":
+    def maybe(f: Flag | Diff[Flag], v: "R | Mask[R]") -> "Mask[R]":
```
sritchie, line 76: you could provide an `@overload` that shows that if a Mask comes in we'd better not see a diff?

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -112,10 +115,17 @@ def _check():
...
 
             def inner(true_v: ArrayLike, false_v: ArrayLike) -> Array:
-                return jnp.where(self.flag.f, true_v, false_v)
+                return jnp.where(self.primal_flag(), true_v, false_v)
 
             return jtu.tree_map(inner, self.value, default)
 
+    def primal_flag(self) -> Flag:
+        match self.flag:
+            case Diff(primal, _):
```
sritchie, line 134: maybe `@property` here?

src/genjax/_src/core/typing.py
```diff
@@ -52,6 +52,7 @@
 Int = int
 Float = float
 Bool = bool
+Flag = Bool | BoolArray
```
sritchie, line 55: woohoo, look at that!

src/genjax/_src/generative_functions/combinators/switch.py
```diff
@@ -84,7 +84,7 @@ def get_sample(self) -> Sample:
             chm = ChoiceMap.empty()
             for _idx, _chm in enumerate(subsamples):
                 assert isinstance(_chm, ChoiceMap)
-                masked_submap = _chm.mask(Flag(jnp.all(_idx == idx)))
+                masked_submap = _chm.mask(jnp.all(_idx == idx))
```
sritchie: my PR kills this jnp.all, I don't think we want this as it prevents `_idx == idx` from staying concrete.

src/genjax/_src/core/interpreters/staging.py
```diff
@@ -39,89 +39,80 @@
 # Concrete Boolean arithmetic #
 ###############################
 
+R = TypeVar("R")
 
-@Pytree.dataclass
-class Flag(Pytree):
+
+class FlagOp:
```
sritchie, line 45: cool, I like it. The name is still a little weird. JAXBool or something might be clearer, but everyone is used to Flag so no need to change.

sritchie (approved):
Nice work, so good to see the types here get cleaned up. This simplifies all of the assertion in choice_map, such a great sign.

src/genjax/_src/core/generative/choice_map.py
sritchie: do we want this to return `jnp.asarray(True)`? should `check` return `BoolArray`, to match the jax "ArrayLike in, Array out"?

src/genjax/_src/core/interpreters/staging.py
sritchie: What about `BoolArrayLike`? that matches the pattern... I suspect you are not going to like this suggestion.

sritchie:
I'll close #1314 ... can you merge that into here?

src/genjax/_src/core/generative/choice_map.py
littleredcomputer: My gut says that we should return a concrete value when we know we can (and this is the point of the type AllSel, a selection that is known to match at compile time)

src/genjax/_src/core/generative/choice_map.py
littleredcomputer: not sure! they're back now

src/genjax/_src/core/generative/functional_types.py
littleredcomputer: I tried this:
```
    @overload
    @staticmethod
    def maybe(f: Flag, v: "Mask[R]") -> "Mask[R]":
        ...

    @overload
    @staticmethod
    def maybe(f: Flag | Diff[Flag], v: R) -> "Mask[R]":
        ...
```
but it did not redline the following test invocation:
```
Mask.maybe(Diff.no_change(True), Mask(True, 99))
```
Because the Mask type matches R in the second  overload :(

src/genjax/_src/core/generative/functional_types.py
littleredcomputer: Hm. Maybe the thing to do is call the property `.flag` and rename the actual inner value? The argument would be that nobody is ever interested in the Diff form of a mask: it has to be that way to satisfy the diff-leaf property but has no meaning beyond that

src/genjax/_src/generative_functions/combinators/switch.py
littleredcomputer: Good point, removing it also

src/genjax/_src/core/interpreters/staging.py
littleredcomputer: The alternatives are not bad but none seems better that FlagOp to me atm

sritchie:
Brilliant!!

sritchie:
@littleredcomputer I went ahead and linked this to the Linear issue. nice work!

PR #1330, opened 2024-09-10
Add LeafSel, AndThenSel for composing selections

sritchie:
This PR:

- adds a `LeafSel` type, allowing for exact matches
- adds `AndThenSel`, allowing for chained matches
- modifies `SelectionBuilder` to extend `Selection.leaf()` by default instead of `Selection.all()`.

Users can recover the old behavior with `S["x", "y"].and_then(Selection.all())`.


sritchie:
Just wanted to note that `LeafSel` is now merged, and `AndThenSel` is not yet merged.

sritchie:
Closing since we don't need this for now and it's easy to recover.

PR #1331, opened 2024-09-10
Fix mask combinator move from false (GEN-365)

sritchie:
We also added a test for the case from True => True.

### DIFF
--- a/src/genjax/_src/generative_functions/combinators/mask.py
+++ b/src/genjax/_src/generative_functions/combinators/mask.py
@@ -182,9 +182,15 @@ class MaskCombinator(Generic[R], GenerativeFunction[Mask[R]]):
         #
         # In any case, we always apply the move... we're not avoiding
         # that computation
-        final_trace = jtu.tree_map(
-            lambda v1, v2: jnp.where(check_arg, v1, v2), premasked_trace, inner_trace
-        )
+        if isinstance(inner_trace, EmptyTrace):
+            final_trace = premasked_trace
+        else:
+            final_trace: Trace[R] = jtu.tree_map(
+                lambda v1, v2: jnp.where(check_arg, v1, v2),
+                premasked_trace,
+                inner_trace,
+            )
+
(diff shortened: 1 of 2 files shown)


src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -134,62 +133,76 @@ def update_change_target(
...
+            final_trace = premasked_trace
+        else:
+            final_trace: Trace[R] = jtu.tree_map(
+                lambda v1, v2: jnp.where(check_arg, v1, v2),
+                premasked_trace,
+                inner_trace,
+            )
+
+        weight = (
+            FlagOp.and_(check, check_arg) * weight
```
sritchie: commenting this out doesn't affect the tests, since the only test we have has `0.0` for weight.

src/genjax/_src/generative_functions/combinators/mask.py
```diff
@@ -134,62 +133,76 @@ def update_change_target(
...
+            + (FlagOp.and_(check, FlagOp.not_(check_arg))) * (-inner_trace.get_score())
+            + FlagOp.and_(FlagOp.not_(check), FlagOp.not_(check_arg)) * 0.0
```
sritchie: why do we have this line in here? Can we leave the math comment and delete, given that we multiply by 0.0?

src/genjax/_src/generative_functions/combinators/mask.py
MathieuHuot: Why do we need to distinguish the case of the inner_trace being an empty trace?

sritchie: @MathieuHuot if we don't, there's a test that fails when it attempts to `tree_map` a `MaskTrace` and an `EmptyTrace` together.

femtomc (approved):

MathieuHuot (approved):

PR #1334, opened 2024-09-11
Upgrade to new treescope dependency, remove custom formatter (GEN-571)

sritchie:
Treescope is now a standalone package: https://github.com/google-deepmind/treescope

This PR pulls `penzai.struct.Struct` into our repo and removes the penzai dependency.


littleredcomputer:
If we don't have any tests for the pytree struct idiom that's a problem.

src/genjax/_src/core/generative/core.py
```diff
@@ -364,13 +363,6 @@ def project(
         )
         return -w
 
-    ###################
-    # Pretty printing #
-    ###################
-
-    def treescope_color(self):
```
sritchie, line 371: I modified the penzai base class to always color structs. Previously it was only coloring structs with `__call__` implemented.

src/genjax/_src/core/pytree.py
```diff
@@ -212,103 +204,8 @@ def static_check_tree_structure_equivalence(trees: list[Any]):
             check = all(map(lambda v: treedef == jtu.tree_structure(v), rest))
             return check
 
-    #############
-    # Utilities #
-    #############
-
     def render_html(self):
-        def _pytree_handler(node, subtree_renderer):
```
sritchie, line 220: I am okay with NOT doing the ellipsis thing on rendering. once we get pretty-printing figured out we'll be able to print a dict representation of choicemaps. In the meantime I actually like seeing things expanded. I think it's okay that it overflows to the right on the website.

sritchie: It will overflow anyway if you expand...

sritchie:
@littleredcomputer this definitely tests pytree / struct idioms... coverage is high on this file (the only misses are for validation errors) and everything fails if we remove inheritance from Struct. JAX doesn't recognize anything as a pytree without this base class.

sritchie:
@littleredcomputer I took back the penzai dependency, as it did NOT have a bunch of transitive dependencies. I had misunderstood this.

femtomc (approved):

PR #1335, opened 2024-09-12
Pretty printing options for ChoiceMap (GEN-162, GEN-422)

sritchie:
Here are some things I tried. `_repr_` is the problem, this is causing the arrays at the leaves not to be foldable:

```python
    def __repr__(self):
        """Renders this object with treescope, on a single line."""
        # Defer to Treescope.
        # import treescope  # pylint: disable=g-import-not-at-top

        print("called")
        return "cake"
        # with treescope.using_expansion_strategy(max_height=1):
        #     return treescope.render_to_text(self.pretty(), ignore_exceptions=True)

    def _repr_pretty_(self, p, cycle):
        """Pretty-prints this object for an IPython pretty-printer."""
        del cycle
        # Defer to Treescope.
        import treescope  # pylint: disable=g-import-not-at-top

        rendering = treescope.render_to_text(self.pretty(), ignore_exceptions=True)
        for i, line in enumerate(rendering.split("\n")):
            if i:
                p.break_()
            p.text(line)

    def __treescope_repr__(self, path, subtree_renderer):
        # Convert the tree to a dictionary
        dict_repr = self.pretty()

        # Use repr_lib to render the object with a custom label
        return treescope.repr_lib.render_dictionary_wrapper(
            object_type=ChoiceMap,
            wrapped_dict=dict_repr,
            path=path,
            subtree_renderer=subtree_renderer,
            roundtrippable=False,
        )
```


src/genjax/pretty.py
```diff
@@ -12,116 +12,12 @@
...
-)
-from penzai.treescope.handlers import builtin_structure_handler
-from penzai.treescope.handlers.penzai import struct_handler
 
-from genjax._src.core.pytree import Pytree
+"""Implementation of an autovisualizer, visualizing arrays."""
 
 
 def pretty():
-    def _pytree_handler(node, subtree_renderer):
```
sritchie: @femtomc it seems like all this did was put in ellipses for the guts of the choicemaps, like

```
In [3]: chm = C[jnp.arange(10), "b"].set(jnp.ones(10)) ^ C["x", "y"].d({"z": 1.0, "q": jnp.arange(100)}).filter(
   ...:     S[..., "y", "q"]
   ...: ) # .mask(jnp.asarray(True))
   ...:
   ...: chm
Out[3]: XorChm(...)
```

This is fine when you're in jupyter and can expand, but it's not good in ipython at the REPL, since there's no way to do the expansion. I don't think we need all of this, and removing it makes the upgrade to treescope easier.

sritchie:
let's also look at https://linear.app/chi-fro/issue/GEN-305/long-rendering-time-for-large-traces-in-jupyter-cell-rendering-both-in for an example of an `IdxChm` that does not print well with `to_dict`. str is not the right move!

sritchie:
Closing in favor of #1359 .

PR #1338, opened 2024-09-12
Fix docs page installation formatting, mix example (GEN-563)

sritchie:
Just a few casual fixes, no big deal...


femtomc (approved):
LOVE IT

PR #1241, opened 2024-07-23
Adding a function to compactly print choicemaps (GEN-377)

MathieuHuot:
Currently added as a cookbook entry with tests showcasing each printing option.
Let me know if there's a better way to integrate this to the library.

PR #1341, opened 2024-09-12
Add in `.T` calls for mask flags' binary operations

sritchie:
This need a proper test.

@MathieuHuot , can you comment with an example on where we need to be careful with indices and why this works?

NOTE that I think we should use https://docs.kidger.site/jaxtyping/ here to type the matching axes.


MathieuHuot:
> This need a proper test.
> 
> @MathieuHuot , can you comment with an example on where we need to be careful with indices and why this works?

@sritchie re tests: I think we should look at all the testing in my compact_printing PR: https://github.com/probcomp/genjax/pull/1241/files#diff-64e7702469415285491f651797da96dd8dce44dd44e10e69f68eb4faa3068fb5

lines 125-152.
and lines 324-349 in
notebooks/cookbook/basics/compact_printing.ipynb

MathieuHuot:
@sritchie  re explanation:
- doing `mask * value_array` will do a point-wise multiplication of the boolean array converted to {0,1}-real valued with the `value_array`, and broadcasting the dimension of the mask appropriately. For instance:
```
import jax.numpy as jnp

array = jnp.arange(1, 10).reshape(3, 3)

mask1 = jnp.array([True, False, True])

array, mask1*array
```
will return 
```
(Array([[1, 2, 3],
        [4, 5, 6],
        [7, 8, 9]], dtype=int32),
 Array([[1, 0, 3],
        [4, 0, 6],
        [7, 0, 9]], dtype=int32))
```
which correctly masks `array[1] (= Array([4, 5, 6], dtype=int32)))`.

MathieuHuot:
> @sritchie re explanation:
> 
> * doing `mask * value_array` will do a point-wise multiplication of the boolean array converted to {0,1}-real valued with the `value_array`, and broadcasting the dimension of the mask appropriately. For instance:
> 
> ```
> import jax.numpy as jnp
> 
> array = jnp.arange(1, 10).reshape(3, 3)
> 
> mask1 = jnp.array([True, False, True])
> 
> array, mask1*array```
> will return 
> ```
> 
> (Array([[1, 2, 3], [4, 5, 6], [7, 8, 9]], dtype=int32), Array([[1, 0, 3], [4, 0, 6], [7, 0, 9]], dtype=int32))
> 
> ```
> which correctly masks `array[1] (= Array([4, 5, 6], dtype=int32)))`.
> ```

Now what happens when using 2 masks? Let's try.
```
mask2 = jnp.array([True, False, True, False, True, False,True, False, True]).reshape(3,3) # looks like an X of True
mask2 * array #removes 2,4,6,8.
mask1 *(mask2 * array) # further removes 5, the last remaining digit in column 1 (the middle one).
```
so we want `mask = f(mask1,mask2)` such that `mask * array` does the above computation.

MathieuHuot:
> > @sritchie re explanation:
> > 
> > * doing `mask * value_array` will do a point-wise multiplication of the boolean array converted to {0,1}-real valued with the `value_array`, and broadcasting the dimension of the mask appropriately. For instance:
> > 
> > ```
> > import jax.numpy as jnp
> > 
> > array = jnp.arange(1, 10).reshape(3, 3)
> > 
> > mask1 = jnp.array([True, False, True])
> > 
> > array, mask1*array```
> > will return 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (Array([[1, 2, 3], [4, 5, 6], [7, 8, 9]], dtype=int32), Array([[1, 0, 3], [4, 0, 6], [7, 0, 9]], dtype=int32))
> > ```
> > which correctly masks `array[1] (= Array([4, 5, 6], dtype=int32)))`.
> > ```
> 
> Now what happens when using 2 masks? Let's try.
> 
> ```
> mask2 = jnp.array([True, False, True, False, True, False,True, False, True]).reshape(3,3) # looks like an X of True
> mask2 * array #removes 2,4,6,8.
> mask1 *(mask2 * array) # further removes 5, the last remaining digit in column 1 (the middle one).
> ```
> 
> so we want `mask = f(mask1,mask2)` such that `mask * array` does the above computation.



> > @sritchie re explanation:
> > 
> > * doing `mask * value_array` will do a point-wise multiplication of the boolean array converted to {0,1}-real valued with the `value_array`, and broadcasting the dimension of the mask appropriately. For instance:
> > 
> > ```
> > import jax.numpy as jnp
> > 
> > array = jnp.arange(1, 10).reshape(3, 3)
> > 
> > mask1 = jnp.array([True, False, True])
> > 
> > array, mask1*array```
> > will return 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (Array([[1, 2, 3], [4, 5, 6], [7, 8, 9]], dtype=int32), Array([[1, 0, 3], [4, 0, 6], [7, 0, 9]], dtype=int32))
> > ```
> > which correctly masks `array[1] (= Array([4, 5, 6], dtype=int32)))`.
> > ```
> 
> Now what happens when using 2 masks? Let's try.
> 
> ```
> mask2 = jnp.array([True, False, True, False, True, False,True, False, True]).reshape(3,3) # looks like an X of True
> mask2 * array #removes 2,4,6,8.
> mask1 *(mask2 * array) # further removes 5, the last remaining digit in column 1 (the middle one).
> ```
> 
> so we want `mask = f(mask1,mask2)` such that `mask * array` does the above computation.

In this nice symmetric case, you could just do `mask1 * mask2` and life seems great. But now let's look at the case where the value array is non-square.
```
array = jnp.arange(1, 13).reshape(3, 4)
mask1 = jnp.array([True, False, True])
mask1 * array
```
suddenly that doesn't work anymore. The broadcast looks at the right and compares 3 to 4 which are incompatible.
So the first fix (@sritchie and maybe we didn't include that one yet!) is that we shouldn't do `mask1 * array`, but `(mask1.T * array.T).T`.

For the same reason that we can't multiply `mask1` and `array`, if we create a new `mask2` of shape (3,4) we will run into trouble.

littleredcomputer:
Is it possible that the mask and target should always be the same shape so
that point wise multiplication is always effective? Then when vmapping
masked things, mask and target will grow in the same way

I wonder if the idea of masking on initial axis was wrong in the first
place

MathieuHuot:
> Is it possible that the mask and target should always be the same shape so that point wise multiplication is always effective? Then when vmapping masked things, mask and target will grow in the same way I wonder if the idea of masking on initial axis was wrong in the first place

It doesn't feel right to me? For instance it seems perfectly fine to sample from a multivariate gaussian within a generative function `model` and have a single boolean mask using `model.mask()`. The former would give an array value, and more generally it could be any tensor with a single mask. When you `vmap` that whole thing, we quickly get into the situation I was describing I think.

MathieuHuot:
@littleredcomputer re "I wonder if the idea of masking on initial axis was wrong in the first place" I don't know to be honest. Not sure why broadcasting is in this direction. It felt natural to me at some point the way we did it, but if it's the exact opposite to how everyone uses it in numpy (which we should check) then yeah we might've done something wrong.

littleredcomputer:
> It doesn't feel right to me? For instance it seems perfectly fine to sample from a multivariate gaussian 

True. So I think the expression is right. Suppose the generative function produces an object *b* of shape (N, k1, k2, ...kn) and we have a mask *a* of shape (N,). Then a.T has shape (N,), b.T has shape (kn, ..., k1, N), so the objects are now broadcast-compatible since the shapes agree from right-to-left; pointwise multiplication will produce an object of shape (kn, ..., k1, N) which transposes back to (N, k1, ..., kn) as desired.

littleredcomputer:
TIL:
- [numpy has a masked-array library](https://numpy.org/doc/stable/reference/maskedarray.html) (jnp does not reimplement it so it's no help to us, but it's fascinating to see another take on this)
- SO has an [implementation that I slightly prefer](https://stackoverflow.com/a/18522281): a * b[:, jnp.newaxis]

MathieuHuot:
> TIL:
> 
> * [numpy has a masked-array library](https://numpy.org/doc/stable/reference/maskedarray.html) (jnp does not reimplement it so it's no help to us, but it's fascinating to see another take on this)
> * SO has an [implementation that I slightly prefer](https://stackoverflow.com/a/18522281): a * b[:, jnp.newaxis]

I'm not sure how the last one behaves in general, e.g. when there are multiple flag tensors being multiplied, and the difference of ranks between the mask array and the value array is more than 1, etc. Also there's a comment "With new axis method it possible to multiply two 1D arrays and generate a 2D array" that seems like something we don't want?

MathieuHuot:
@sritchie 
- the and flag looks good
- I see what you mean now for the or flag and this seems right
- I don't see the required change when multiplying the mask by the value though. I think I explained above why we still need to do the same trick an extra time: (mask.T * value.T).T is required for printing the value tensor with masked values replaced by 0. Something similar needs to happen when computing scores or returned values that depend on the traces values, and I don't know if we do it right yet.

The 2 tricks don't cancel out, but maybe we can change the invariant on the flag to propagate the flag transposed instead of the flag: that way when we have two masks (representing the transposed masks) mask1 and mask2, the updated one would just be mask1 * mask2 (and max(mask1,mask2) for the or case). But we'd still need (mask * value.T).T somewhere at the leaves.

sritchie:
Closing this in favor of the correct fix in #1422 (I'll update the description there)

PR #1348, opened 2024-09-16
type-check inference package (GEN-484)

sritchie:
These warnings weren't too tough to get rid of.


src/genjax/_src/inference/sp.py
```diff
@@ -87,37 +90,14 @@ def __getitem__(self, addr):
...
-    @abstractmethod
-    def estimate_logpdf(
-        self,
-        key: PRNGKey,
-        v: ChoiceMap,
-        *args: Any,
-    ) -> FloatArray:
-        raise NotImplementedError
-
+SampleDistribution = Distribution[ChoiceMap]
```
MathieuHuot, line 93: It would be good to keep a comment here about why this thing exists, especially if we're gonna refactor the SMC library later. Could you e.g. add the previous docstring as a comment above?

sritchie: done! I added it as a docstring on `SampleDistribution`, good call.

PR #1349, opened 2024-09-16
Make @genjax.gen work on methods (GEN-541)

sritchie:
This PR:

- modifies `@genjax.gen` so that multiple levels of `Closure` will flatten into a single `Closure`, keeping one level of `dyn_args` vs multiple levels
- adds the ability to use `@genjax.gen` on instance methods, partially applying the `self` argument automatically
- adds `partial_args` to `StaticGenerativeFunction`, for recovering the partially-applied args.

```python
In [5]: import jax
   ...: import genjax
   ...: from genjax._src.core.pytree import Pytree
   ...: from genjax._src.core.typing import Array, ArrayLike
   ...:
   ...:
   ...: @genjax.gen
   ...: def outside(x):
   ...:     return genjax.normal(x, 1.0) @ "x"
   ...:
   ...: @Pytree.dataclass
   ...: class Model(Pytree):
   ...:     foo: ArrayLike
   ...:     bar: ArrayLike
   ...:
   ...:     @genjax.gen
   ...:     def run(self, x):
   ...:         y = genjax.normal(self.foo, self.bar) @ "y"
   ...:         z = genjax.normal(x, 1.0) @ "z"
   ...:         return y + z
   ...:
   ...: key = jax.random.PRNGKey(0)
   ...: m = Model(foo=4.0, bar=6.0)
   ...: m.run.simulate(key, (1.0,)).get_choices()
Out[5]:
Static({
  'y': Choice(v=<jax.Array(-4.405305, dtype=float32)>),
  'z': Choice(v=<jax.Array(1.3905154, dtype=float32)>),
})
```


src/genjax/_src/generative_functions/static.py
```diff
@@ -89,15 +89,15 @@ def get_visited(self):
...
+    gen_fn: "StaticGenerativeFunction[R]"
     args: tuple[Any, ...]
     retval: R
     addresses: AddressVisitor
     subtraces: list[Trace[Any]]
     score: Score
 
     def get_args(self) -> tuple[Any, ...]:
-        return self.args
+        return self.gen_fn.source.dyn_args + self.args
```
femtomc: This won't work with `edit` without serious thinking.

tests/generative_functions/test_static_gen_fn.py
```diff
@@ -773,3 +773,48 @@ def higher_higher_model():
...
+
+        double_curry = model.partial_apply(1.0).partial_apply(1.0)
+        key = jax.random.PRNGKey(0)
+
+        tr = double_curry.simulate(key, (2.0,))
+        assert tr.get_args() == (
+            2.0,
+        ), "both curried args are not present alongside the final arg"
+
+        assert tr.gen_fn.source.dyn_args == (
```
sritchie: @littleredcomputer we could add a `.get_partial_args()` to `StaticGenerativeFunction` to get the `dyn_args` out?

src/genjax/_src/generative_functions/static.py
```diff
@@ -631,7 +639,8 @@ def my_model(x, y):
             # Now `partially_applied_model` is equivalent to a model that only takes 'y' as an argument
             ```
         """
-        return gen(Pytree.partial(*args)(self.inline))
+        all_args = self.source.dyn_args + args
```
sritchie, line 1147: this flattens out nested `Closure` instances, so we have a single level, a single `dyn_args` and access to the original wrapped fn.

src/genjax/_src/generative_functions/static.py
```diff
@@ -643,8 +652,8 @@ def gen(f: Closure[R] | Callable[..., R]) -> StaticGenerativeFunction[R]:
     if isinstance(f, Closure):
         return StaticGenerativeFunction[R](f)
     else:
-        closure = Pytree.partial()(f)
-        return StaticGenerativeFunction[R](closure)
+        closure = Closure[R]((), f)
```
sritchie, line 1160: I thought it was less opaque to create a Closure directly where we could.

femtomc:
@sritchie Can you add one more test case, if you don't mind:
> A call to `update` — where `Model` is instantiated as `m` inside a generative function (with arguments that flow into the construction), and then `m.run` is invoked inside that same generative function

this will test if the incremental interpreter works correctly with this idiom

sritchie:
@femtomc test added!

src/genjax/_src/generative_functions/static.py
sritchie: Removed

tests/generative_functions/test_static_gen_fn.py
sritchie: I added this as a `partial_args` property

PR #1354, opened 2024-09-19
modify StaticChm to wrap a dict

sritchie:
This PR modifies `StaticChm` so that it wraps a dictionary. This allows many addresses at the same level of tracing to show up in a single dict, vs a nested series of `XorChm`s:

image


femtomc (approved):
love it, way easier to read

PR #1355, opened 2024-09-20
IdxChm slice support, jnp.array lookup support

sritchie:


sritchie:
This whole thing is making me nervous. I am feeling more and more that `Indexed` should go away. More thoughts coming, but I've converted this to a draft for now so we don't merge it...

PR #1357, opened 2024-09-21
Add deterministic (& incremental aware) caching to the static language.

femtomc:
Adds the ability to store deterministic data in `StaticTrace` via a new caching primitive.

The caching primitive is aware of incremental computation, and will only recompute the cached function in `edit` (and `update`) if changes to the arguments are indicated by `UnknownChange`.

When holding a `StaticTrace`, accessing the cache can be performed via `StaticTrace.get_cache()` (it's stored as a choice map, for now).

TODOs:

- [ ] Caching is not compositional currently -- meaning, there's not a compositional trace interface for it, so you can only use it if you have a `StaticGenerativeFunction` at the toplevel (so e.g. you get to hold a `StaticTrace`, and not some other trace (e.g. calling a `StaticGenerativeFunction` with caching _inside_ of a `VmapCombinator` -- gives you a `VmapTrace`, and there's not yet a `Trace` interface which allows you to fetch the cache inside of the `VmapTrace`).


tests/generative_functions/test_static_gen_fn.py
```diff
@@ -777,3 +777,37 @@ def higher_higher_model():
             == genjax.normal.assess(choice.get_submap("y1"), (0.0, 1.0))[0]
             + genjax.normal.assess(choice.get_submap("y2"), (0.0, 1.0))[0]
         )
+
+
+class TestCaching:
```
femtomc, line 780: @georgematheos See these tests for usage.

georgematheos: @femtomc this looks great!

Is there a way to add to this test something that checks if `f` is run multiple times?  Would it work to have `f` increment a global counter variable, and check that this isn't incremented after `update` calls that shouldn't cause re-computation?

femtomc: Unfortunately JAX may run `f` abstractly multiple times during tracing -- that doesn't say much about the generated code.

For this one, we might just have to inspect the Jaxprs...

femtomc: Just checked the Jaxprs, it works

femtomc: Let's try and see to get it working in your codebase

femtomc: image

src/genjax/_src/generative_functions/static.py
```diff
@@ -181,6 +194,45 @@ def trace(
...
+    addr = Pytree.tree_const(addr)
+    fn = Pytree.partial()(fn)
+    return initial_style_bind(cache_p)(_abstract_call)(
+        addr,
+        fn,
+        args,
+    )
+
+
+def cache(
```
sritchie: bound it on `TypeVar("T", bound=Callable)`, then it'll pin the full type sig

src/genjax/_src/generative_functions/static.py
```diff
@@ -114,6 +116,14 @@ def get_choices(self) -> ChoiceMap:
         sub_chms = (tr.get_choices() for tr in self.subtraces)
         return ChoiceMap.from_mapping(zip(addresses, sub_chms))
 
+    def get_cache(self) -> ChoiceMap:
+        addresses = self.cached_addresses.get_visited()
+        return ChoiceMap.from_mapping(zip(addresses, self.cached_values))
+
+    def get_cached_state(self, addr) -> Any:
```
sritchie: type this addr, please

sritchie: and docs for these new public methods (use Cursor to generate with an example!)

src/genjax/_src/generative_functions/static.py
```diff
@@ -143,6 +153,9 @@ class AddressReuse(Exception):
 # Generative function trace intrinsic.
 trace_p = InitialStylePrimitive("trace")
 
+# Deterministic caching intrinsic.
```
sritchie, line 163: can you say more?

src/genjax/_src/generative_functions/static.py
```diff
@@ -181,6 +194,45 @@ def trace(
     )
 
 
+def _abstract_call(
```
sritchie: can you inline this? does it need to be a def?

src/genjax/_src/generative_functions/static.py
```diff
@@ -181,6 +194,45 @@ def trace(
...
+    return fn(*args)
+
+
+def _cache(
+    addr: StaticAddress,
+    fn: Callable[[Any], R],
+    args: Any,
+) -> R:
+    addr = Pytree.tree_const(addr)
+    fn = Pytree.partial()(fn)
```
sritchie: construct a Closure directly to make the types more clear

src/genjax/_src/generative_functions/static.py
```diff
@@ -205,14 +257,22 @@ def handle_trace(
...
+        args: Any,
+    ) -> R:
+        return fn(*args)
+
     def handle_retval(self, v):
         return jtu.tree_leaves(v)
 
     # By default, the interpreter handlers for this language
     # handle the two primitives we defined above
     # (`trace_p`, for random choices)
```
sritchie: can you move this comment down into a docstring?

MathieuHuot:
@georgematheos could you give some more context by providing the original example from ChiSight that motivated this (or a simplified version)?

My understanding from one of our discussions was that you wanted to visualise intermediate results within inference for debugging. If that's the main purpose, we want to tune for that purpose, or at least think about this use case some more.

femtomc:
@MathieuHuot / @sritchie Let's not merge this in now -- I need to continue working on it to develop an example from George's Notion: https://www.notion.so/chi-mit/Incremental-computation-example-for-McCoy-10701b4b8cde803f9919d308cb0e331c?pvs=4

femtomc:
Caching a little message I talked over with Mathieu:
```python
# In the static language...
@genjax.gen
def model(x):
    v = submodel.vmap(in_axes=(0, ))(x) @ "v"
    x = cache("x", jnp.sum)(v) # this is new: incremental aware caching
    return some_other_model(x) @ "y"

tr # model trace in hand
# if you change "y", but don't change "v"
# jnp.sum @ "x" doesn't get recomputed.
#
# If we change "v", but don't change the argument x
# then v gets Diff(new_v, UnknownChange)
# the cache primitive sees this, and recomputes jnp.sum @ "x"
#
# Let's say we change "v" using IndexEditRequest(idx = 3, ...)
# NOTE: this is a TODO in the library (we have to implement IndexEditRequest for vmap).
# Now, the value for v -- what is it tagged with?
# 
# We should have: v gets Diff(new_v, IndexChange(idx, old_v_at_idx))
# And then what does cache do?
# 
# That's where we might want to allow users to specify rules.
@genjax.gen
def model(x):
    v = submodel.vmap(in_axes=(0, ))(x) @ "v"
    v = some_other_fn(v)
    x = cache("x", jnp.sum, custom_rule= lambda v: ...)(v) # this is new: incremental aware caching
    return some_other_model(x) @ "y"

```

src/genjax/_src/generative_functions/static.py
femtomc: When I do this, I get an error -- that `Callable` is a variable and it's not allowed in type expressions.

src/genjax/_src/generative_functions/static.py
femtomc: Done.

PR #1360, opened 2024-09-23
Add test showing that project works (GEN-316)

sritchie:


MathieuHuot (approved):
Great that it works. I'm uneasy on the coverage for a more complex usage of `project` but it seems writing down the math is a prerequisite to writing a more complex test which seems outside the scope of this PR.

PR #1363, opened 2024-09-24
Convert nested `Static` choicemaps into a single `Static` with nested dicts (GEN-617)

sritchie:
This PR modifies `Static` so that it will

- unwrap `Static` values and store the bare dictionary
- re-wrap dict-shaped values as `Static` on the way out.

For example:

```python
@genjax.gen
def g(x, y):
    a = genjax.normal(x, y) @ "a"
    b = genjax.normal(x, y) @ "b"
    c = genjax.normal(x, y) @ "c"
    return a + b + c

@genjax.gen
def f(x, y):

    x = g(x, y) @ "x"
    y = genjax.normal(x, y) @ "y"
    z = genjax.normal(x, y) @ "z"
    return x + y + z

key = jax.random.PRNGKey(0)

chm = f.simulate(key, (0.0, 1.0)).get_choices()
chm
```

image

```python
chm.filter(S["x", "a"] | S["y"]).simplify()
```

image


eightysteele (approved):
LGTM. Had a couple of thoughts on testing:

1. Does `Static` handle empty mappings correctly (e.g., it doesn’t fail when no values are present)
2. Do changes in `Static` handle vectorized ops (e.g., wrapping/un-wrapping doesn't mess with vmap computations)

sritchie:
@eightysteele 

- empty mapping is fine, in fact we rely on this for our "empty choice map" implementation. `Static({})` is returned by `ChoiceMap.empty()`.
- dicts are pytrees, so JAX doesn't send their structure to the GPU, making dropping a layer internally like this a no-op. I have a test in there showing that you can't get a bare dict OUT via `get_submap`, so nothing changes in the interfaces, this is internal only.

PR #1365, opened 2024-09-24
Add a switch static method to ChoiceMap

sritchie:
This PR adds a `switch` static method for switched chm creation, and uses it in `SwitchTrace`.


MathieuHuot (approved):
Lgtm. This is in line with the idea we discussed of having one choice map creator per combinator to make it easier for users to create choice maps that resemble the structure of their generative model.

We'll need to update the cookbook at some point, but maybe we can do in a batch once we have more of these, e.g. for map or scan.

femtomc:
@MathieuHuot
> Lgtm. This is in line with the idea we discussed of having one choice map creator per combinator to make it easier for users to create choice maps that resemble the structure of their generative model.

This is in line with the idea -- but users _should not_ be making this type of choice map to specify constraints to `SwitchCombinator` -- they should make a normal choice map (to constrain a branch).

While technically possible to specify constraints on branches using this one, it's quite unwieldy -- I think this is mostly about moving an internal idiom into our big choice map file (putting things in one place).

littleredcomputer:
Are we confident that the latent tests are sufficient for this change? I wondered, while reading this code, what the observable difference would be if we changed the implementation of ChoiceMap.switch to the following:
```
        return list(chms)[idx.astype(int)]
```
To my surprise, all unit tests but one _passed_ with this implementation, and the one failing test involved a trace constructed with vmap but operated on directly. Can we say more in the doc about why the more elaborate design is necessary?

My guess is one could want to see the choices made in other branches either by inspection or by updating the choice argument. There should be a test for these two activities and the change above proves that there aren't any

sritchie:
@littleredcomputer , good idea... we talked about this today but I'll hold off on merging until I get an explanation of what's going on and some tests that pass in a dynamic index.

PR #1366, opened 2024-09-24
Use `pass` vs `UnimplementedError` for abstract methods

sritchie:
This lets us find the actual unimplemented methods that we need to fill in.


femtomc:
> This lets us find the actual unimplemented methods that we need to fill in.

As in -- `pyright`? Or, as an error?

femtomc (approved):

PR #1367, opened 2024-09-24
Implement `__and__` dunder for choicemaps for intersection of 2 chms

sritchie:


femtomc:
What does `__and__` mean? Does it mean "xor"? If it means xor, should we use `__xor__`?

... edit: spoke too quickly

```
    def __and__(self, other: "ChoiceMap") -> "ChoiceMap":
        return other.filter(self.get_selection())
```

this seems like a really bizarre operation. Do we have people clamoring for this?

femtomc (approved):

PR #1369, opened 2024-09-26
Static ChoiceMap shape validation, `vmap` and `scan` validation upgrades (GEN-418, GEN-83)

sritchie:
This PR:

- adds an `invalid_subset` method to `ChoiceMap`. This takes a generative function and its args tuple and returns the statically-invalid subset of the choicemap — i.e., the subset of entries that could NEVER affect the generative function's execution — or None if the choicemap is fully valid.
- adds `Selection.leaf()`, for building selections that match a specific spot without matching the rest of the tail
- runs side-effecting `vmap` and `scan` calls while computing the dimension size — this has the lovely effect of triggering the very detailed validations provided by JAX on mismatched types, mismatched dimensions, invalid in_axes and all the rest.
- modifies `ScanTrace` and `VmapTrace` to statically compute the size of their scanned or vmapped dimensions for storage in the trace
- modifies `VmapTrace` to pre-compute the `score` and `chm` fields
- modifies `ScanTrace` to pre-compute the `chm` field

The latter two need to happen because an outer `jax.vmap` call adds a new dimension at all trace leaves — if we delay creating choicemaps in `get_choices` to do it on demand, and someone has added a new dimension, then we will vmap across the WRONG dimension here.

This PR does NOT wire in any warnings for the user in `update` or any of those methods. I'm going to propose that we do this in another issue.

Examples from the new choicemap tests:

```python
    def test_choicemap_validation(self):
        @genjax.gen
        def model(x):
            y = genjax.normal(x, 1.0) @ "y"
            z = genjax.bernoulli(0.5) @ "z"
            return y + z

        # Valid ChoiceMap
        valid_chm = ChoiceMap.kw(y=1.0, z=1)
        assert valid_chm.invalid_subset(model, (0.0,)) is None

        # Invalid ChoiceMap - missing 'z'
        invalid_chm1 = ChoiceMap.kw(x=1.0)
        assert invalid_chm1.invalid_subset(model, (0.0,)) == invalid_chm1

        # Invalid ChoiceMap - extra address
        invalid_chm2 = ChoiceMap.kw(y=1.0, z=1, extra=0.5)
        assert invalid_chm2.invalid_subset(model, (0.0,)) == ChoiceMap.kw(extra=0.5)

    def test_choicemap_nested_validation(self):
        @genjax.gen
        def inner_model():
            a = genjax.normal(0.0, 1.0) @ "a"
            b = genjax.bernoulli(0.5) @ "b"
            return a + b

        @genjax.gen
        def outer_model():
            x = genjax.normal(0.0, 1.0) @ "x"
            y = inner_model() @ "y"
            return x + y

        # Valid nested ChoiceMap
        valid_nested_chm = ChoiceMap.kw(x=1.0, y=ChoiceMap.kw(a=0.5, b=1))
        assert valid_nested_chm.invalid_subset(outer_model, ()) is None

        # Invalid nested ChoiceMap - missing inner 'b'
        invalid_nested_chm1 = ChoiceMap.kw(x=1.0, y=ChoiceMap.kw(a=0.5))
        assert (
            invalid_nested_chm1.invalid_subset(outer_model, ()) is None
        ), "missing address is fine"

        # Invalid nested ChoiceMap - extra address in inner model
        invalid_nested_chm2 = ChoiceMap.kw(x=1.0, y=ChoiceMap.kw(a=0.5, b=1, c=2.0))
        assert invalid_nested_chm2.invalid_subset(outer_model, ()) == ChoiceMap.kw(
            y=ChoiceMap.kw(c=2.0)
        )

        # Invalid nested ChoiceMap - extra address in outer model
        invalid_nested_chm3 = ChoiceMap.kw(x=1.0, y=ChoiceMap.kw(a=0.5, b=1), z=3.0)
        assert invalid_nested_chm3.invalid_subset(outer_model, ()) == ChoiceMap.kw(
            z=3.0
        )

    def test_choicemap_nested_vmap(self):
        @genjax.gen
        def inner_model(x):
            a = genjax.normal(x, 1.0) @ "a"
            b = genjax.bernoulli(0.5) @ "b"
            return a + b

        @genjax.gen
        def outer_model():
            x = genjax.normal(0.0, 1.0) @ "x"
            y = inner_model.vmap(in_axes=(0,))(jnp.array([1.0, 2.0, 3.0])) @ "y"
            return x + jnp.sum(y)

        # Valid nested ChoiceMap with vmap
        valid_vmap_chm = ChoiceMap.kw(
            x=1.0,
            y=C[jnp.arange(3)].set(
                ChoiceMap.kw(a=jnp.array([0.5, 1.5, 2.5]), b=jnp.array([1, 0, 1]))
            ),
        )
        assert valid_vmap_chm.invalid_subset(outer_model, ()) is None

        # Invalid nested ChoiceMap - wrong shape for vmapped inner model
        inner_chm = ChoiceMap.kw(a=jnp.array([0.5, 1.5, 2.5]), b=jnp.array([1, 0, 1]))
        invalid_vmap_chm1 = ChoiceMap.kw(
            x=1.0,
            # missing the index nesting
            y=inner_chm,
        )
        assert invalid_vmap_chm1.invalid_subset(outer_model, ()) == C["y"].set(
            inner_chm
        )

        # Invalid nested ChoiceMap - extra address in vmapped inner model

        invalid_vmap_chm2 = ChoiceMap.kw(
            x=1.0,
            y=C[jnp.arange(3)].set(
                ChoiceMap.kw(
                    a=jnp.array([0.5, 1.5, 2.5]),
                    b=jnp.array([1, 0, 1]),
                    c=jnp.array([0.1, 0.2, 0.3]),  # Extra address
                )
            ),
        )
        expected_result = C["y", jnp.arange(3), "c"].set(jnp.array([0.1, 0.2, 0.3]))
        actual_result = invalid_vmap_chm2.invalid_subset(outer_model, ())
        assert jax.tree_util.tree_structure(
            actual_result
        ) == jax.tree_util.tree_structure(expected_result)
        assert jax.tree_util.tree_all(
            jax.tree_map(
                lambda x, y: jnp.allclose(x, y), actual_result, expected_result
            )
        )

    def test_choicemap_switch(self):
        @genjax.gen
        def model1():
            x = genjax.normal(0.0, 1.0) @ "x"
            return x

        @genjax.gen
        def model2():
            y = genjax.uniform(0.0, 1.0) @ "y"
            return y

        @genjax.gen
        def model3():
            z = genjax.normal(0.0, 1.0) @ "z"
            return z

        switch_model = genjax.switch(model1, model2, model3)

        @genjax.gen
        def outer_model():
            choice = genjax.categorical([0.3, 0.3, 0.4]) @ "choice"
            return switch_model(choice, (), (), ()) @ "out"

        # Valid ChoiceMap for model1
        valid_chm1 = ChoiceMap.kw(choice=0, out={"x": 0.5})
        assert valid_chm1.invalid_subset(outer_model, ()) is None

        # Valid ChoiceMap for model2
        valid_chm2 = ChoiceMap.kw(choice=1, out={"y": 0.7})
        assert valid_chm2.invalid_subset(outer_model, ()) is None

        # Valid ChoiceMap for model3
        valid_chm3 = ChoiceMap.kw(choice=2, out={"z": 1.2})
        assert valid_chm3.invalid_subset(outer_model, ()) is None

        # Valid ChoiceMap with entries for all models
        valid_chm_all = ChoiceMap.kw(choice=0, out={"x": 0.5, "y": 0.7, "z": 1.2})
        assert valid_chm_all.invalid_subset(outer_model, ()) is None

        # Invalid ChoiceMap - extra address
        invalid_chm2 = ChoiceMap.kw(choice=1, out={"q": 0.5})
        assert invalid_chm2.invalid_subset(outer_model, ()) == C["out", "q"].set(0.5)
        pass

    def test_choicemap_scan(self):
        @genjax.gen
        def inner_model(mean):
            return genjax.normal(mean, 1.0) @ "x"

        outer_model = inner_model.iterate(n=4)

        # Test valid ChoiceMap
        valid_chm = C[jnp.arange(4), "x"].set(jnp.array([0.5, 1.2, 0.8, 0.9]))
        assert valid_chm.invalid_subset(outer_model, (1.0,)) is None

        # forgot the index layer
        invalid_chm2 = C["x"].set(jnp.array([0.5, 1.2, 0.8, 0.9]))
        assert invalid_chm2.invalid_subset(outer_model, (1.0,)) == invalid_chm2

        xs = jnp.array([0.5, 1.2, 0.8, 0.9])
        zs = jnp.array([0.5, 1.2, 0.8, 0.9])
        invalid_chm3 = C[jnp.arange(4)].set({"x": xs, "z": zs})
        invalid_subset = invalid_chm3.invalid_subset(outer_model, (1.0,))
        expected_invalid = C[jnp.arange(4), "z"].set(zs)
        assert jtu.tree_structure(invalid_subset) == jtu.tree_structure(
            expected_invalid
        )
        assert jtu.tree_all(
            jtu.tree_map(
                lambda x, y: jnp.allclose(x, y), invalid_subset, expected_invalid
            )
        )

```

### DIFF
--- a/src/genjax/_src/core/generative/choice_map.py
+++ b/src/genjax/_src/core/generative/choice_map.py
@@ -200,6 +200,26 @@ class Selection(Projection["ChoiceMap"]):
         """
         return NoneSel()
 
+    @staticmethod
+    def leaf() -> "Selection":
+        """
+        Returns a Selection that selects only leaf addresses.
+
+        A leaf address is an address that doesn't have any sub-addresses.
+        This selection is useful when you want to target only the final elements in a nested structure.
+
+        Returns:
+            A Selection that selects only leaf addresses.
+
+        Example:
+            ```python exec="yes" html="true" source="material-block" session="choicemap"
+            leaf_selection = Selection.leaf().extend("a", "b")
+            assert leaf_selection["a", "b"]
+            assert not leaf_selection["a", "b", "anything"]
+            ```
+        """
+        return LeafSel()
+
     ######################
     # Combinator methods #
     ######################
@@ -1330,20 +1350,6 @@ class Choice(Generic[T], ChoiceMap):
 
     v: T
 
-    # def __xor__(self, other: "ChoiceMap") -> "ChoiceMap":
-    #     if isinstance(other, Choice):
-    #         raise Exception(
-    #             f"The disjoint union of two choice maps have a value collision:\nc1 = {self}\nc2 = {other}"
-    #         )
-    #     else:
-    #         return Xor.build(self, other)
-
-    def __or__(self, other: "ChoiceMap") -> "ChoiceMap":
-        if isinstance(other, Choice):
-            return self
-        else:
-            return Or.build(self, other)
-
     def get_value(self) -> T:
         return self.v
 
@@ -1405,7 +1411,6 @@ class Indexed(ChoiceMap):
             )
 
             check_array = jnp.asarray(check, copy=False)
-
             if check_array.shape:
                 if check_array.shape[0] == 0:
                     # this is an obscure case which can arise when doing an importance
@@ -1557,7 +1562,7 @@ class Xor(ChoiceMap):
         v2 = self.c2.get_value()
 
         def pair_flag_to_idx(first: Flag, second: Flag):
-            return first + (2 * second) - 1
+            return first + 2 * second - 1
 
         idx = pair_flag_to_idx(check1, check2)
 
@@ -1690,9 +1695,7 @@ class Filtered(ChoiceMap):
 
     def get_submap(self, addr: ExtendedAddressComponent) -> ChoiceMap:
         submap = self.c.get_submap(addr)
-
-        # this is one spot where
-        subselection = self.selection.get_subselection(addr)
+        subselection = self.selection(addr)
         return submap.filter(subselection)
 
 
@@ -1706,8 +1709,7 @@ def _pushdown_filters(chm: ChoiceMap) -> ChoiceMap:
                 })
 
             case Indexed(c, addr):
-                sel = selection(addr)
-                return loop(c, sel).extend(addr)
+                return loop(c, selection(addr)).extend(addr)
 
             case Choice(v):
                 if v is None:
--- a/src/genjax/_src/core/generative/choice_map.py
+++ b/src/genjax/_src/core/generative/choice_map.py
@@ -27,7 +27,10 @@ from deprecated import deprecated
 from genjax._src.core.generative.core import Constraint, Projection, Sample
 from genjax._src.core.generative.functional_types import Mask, staged_choose
 from genjax._src.core.interpreters import staging
-from genjax._src.core.interpreters.staging import FlagOp, staged_err
+from genjax._src.core.interpreters.staging import (
+    FlagOp,
+    staged_err,
+)
 from genjax._src.core.pytree import Pytree
 from genjax._src.core.typing import (
     Any,
@@ -1021,7 +1024,9 @@ class ChoiceMap(Sample):
         return ChoiceMap.d(kwargs)
 
     @staticmethod
-    def switch(idx: ArrayLike, chms: Iterable["ChoiceMap"]) -> "ChoiceMap":
+    def switch(
+        idx: ArrayLike | jax.ShapeDtypeStruct, chms: Iterable["ChoiceMap"]
+    ) -> "ChoiceMap":
         """
         Creates a ChoiceMap that switches between multiple ChoiceMaps based on an index.
 
@@ -1051,7 +1056,7 @@ class ChoiceMap(Sample):
         acc = ChoiceMap.empty()
         for _idx, _chm in enumerate(chms):
             assert isinstance(_chm, ChoiceMap)
-            masked = _chm.mask(_idx == idx)
+            masked = _chm.mask(jnp.all(_idx == idx))
             acc ^= masked
         return acc
 
--- a/src/genjax/_src/core/generative/choice_map.py
+++ b/src/genjax/_src/core/generative/choice_map.py
@@ -15,6 +15,7 @@
 from abc import abstractmethod
 from dataclasses import dataclass
 from operator import or_
+from typing import TYPE_CHECKING
 
 import jax
 import jax.numpy as jnp
@@ -25,6 +26,7 @@ from deprecated import deprecated
 
 from genjax._src.core.generative.core import Constraint, Projection, Sample
 from genjax._src.core.generative.functional_types import Mask, staged_choose
+from genjax._src.core.interpreters import staging
 from genjax._src.core.interpreters.staging import (
     FlagOp,
     staged_err,
@@ -42,6 +44,9 @@ from genjax._src.core.typing import (
     String,
     TypeVar,
 )
+
+if TYPE_CHECKING:
+    import genjax
 
 #################
 # Address types #
@@ -375,6 +380,29 @@ class NoneSel(Selection):
         return self
 
 
+@Pytree.dataclass
+class LeafSel(Selection):
+    """Represents a selection that matches only at the current address level.
+
+    This selection returns True for a check at the current level but returns an
+    empty selection (`Selection.none()`) for any subselection, effectively representing a
+    leaf node in the selection hierarchy.
+
+    Examples:
+        ```python exec="yes" html="true" source="material-block" session="choicemap"
+        leaf_sel = LeafSel()
+        assert leaf_sel.check()
+        assert isinstance(leaf_sel.get_subselection("any_address"), NoneSel)
+        ```
+    """
+
+    def check(self) -> Flag:
+        return True
+
+    def get_subselection(self, addr: ExtendedAddressComponent) -> Selection:
+        return Selection.none()
+
+
 @Pytree.dataclass(match_args=True)
 class MaskSel(Selection):
     """Represents a selection that is conditionally applied based on a flag.
@@ -541,10 +569,10 @@ class AndSel(Selection):
                 return a
             case (_, NoneSel()):
                 return b
-
             case (MaskSel(), MaskSel()):
                 return MaskSel.build(FlagOp.and_(a.flag, b.flag))
-
+            case (a, b) if a == b:
+                return a
             case _:
                 return AndSel(a, b)
 
@@ -597,7 +625,8 @@ class OrSel(Selection):
                 return a
             case (MaskSel(), MaskSel()):
                 return MaskSel.build(FlagOp.or_(a.flag, b.flag))
-
+            case (a, b) if a == b:
+                return a
             case _:
                 return OrSel(a, b)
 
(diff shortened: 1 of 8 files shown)


src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1263,13 +1330,13 @@ class Choice(Generic[T], ChoiceMap):
 
     v: T
 
-    def __xor__(self, other: "ChoiceMap") -> "ChoiceMap":
```
sritchie, line 1263: I realized that this was NOT correct in the face of `Choice(Mask(...))`, which we special-case in `has_value()`:

```python
    def has_value(self) -> Flag:
        match self.get_value():
            case None:
                return False
            case Mask() as m:
                return m.primal_flag()
            case _:
                return True
```

this is making me uneasy...

sritchie:
@femtomc before merging I think we need to stare closely at this error: https://github.com/probcomp/genjax/actions/runs/11045407654/job/30682981247#step:10:312

```
ValueError: vmap got inconsistent sizes for array axes to be mapped:
  * most axes (16 of them) had size 9, e.g. axis 0 of argument subtrace.args[0] of type float32[9,20];
  * one axis had size 20: axis 0 of argument idx of type int32[20]
```

I "fixed" it with this change: 

https://github.com/probcomp/genjax/pull/1369/files#diff-a867f858d1c9da58fba86cb8333753e4e4194ecdda0d929145ed7b04f176fc7fR70

but thinking about it now I think that this is NOT a fix, because the vmap forces us to check that the indices all match internally, and skipping it clearly is skipping an error from mismatched dimensions.

I think maybe this is exposing a problem with either
- the pattern of vmapping keys, and how it adds indices to internal arrays?
- the caching we're doing on the dimension

We've cached the dimension, but then an OUTER vmap call inflated the dimension. I think without our change, if we vmap over 20 keys, we'll get a score of leading dimension 20.

Anyway I like what we are doing here with validation, that's clearly correct. But is caching the dimension the right move?

sritchie:
@femtomc yeah, this would also happen if we vmapped keys over a scan. I think any nested vmap will hit this, actually.

We CAN fix this by not caching the size at all. Once we have better indexing support then we wouldn't materialize the indices, we'd use `slice(None, None, None)` for our new address...

BUT actually this is weird. this is a case where jax.vmap has added an extra level inside of our choicemaps, BUT it has not added a wrapping `Indexed` choicemap. Once again we've hit the batched handling debate.

sritchie:
I fixed the problem by pre-computing and caching the choicemap AND the score.

src/genjax/_src/generative_functions/combinators/scan.py
```diff
@@ -173,6 +189,12 @@ def _inner(carry: Carry, scanned_in: Any):
 
         return v, scanned_out
 
+    @staticmethod
+    def _static_scan_length(xs: Any, length: int | None) -> int:
+        # We start by triggering a scan to force all JAX validations to run.
+        jax.lax.scan(lambda c, x: (c, None), None, xs, length=length)
```
littleredcomputer, line 195: I'm concerned about these pro-forma checks of shape getting dragged into jitted code. If I jit `simulate` because I want to then vmap it across an array of keys, will this happen once per key? If not, why not?

littleredcomputer: Another point: while looking at ShapeDtypeStruct I ran into the function [jax.eval_shape](https://jax.readthedocs.io/en/latest/_autosummary/jax.eval_shape.html) ... possibly useful? It can't avoid side effects but can avoid math

littleredcomputer: the question would be: does that also provoke the scan shape-checking (you'd have to eval_shape on a lambda with scan within it)

src/genjax/_src/generative_functions/combinators/vmap.py
```diff
@@ -52,14 +52,25 @@ class VmapTrace(Generic[R], Trace[R]):
...
+
+    @staticmethod
+    def build(
+        gen_fn: GenerativeFunction[R], tr: Trace[R], args: tuple[Any, ...], length: int
+    ) -> "VmapTrace[R]":
+        score = jnp.sum(tr.get_score())
+        chm = jax.vmap(
+            lambda idx, subtrace: ChoiceMap.entry(subtrace.get_choices(), idx),
+        )(jnp.arange(length), tr)
+
```
littleredcomputer, line 66: I do like the pre-computation of chm. It does seem that by doing it here we are avoiding having it happen after arrays are stacked further onto the trace. One hopes that it is not duplicating the array space

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1239,6 +1295,43 @@ def simplify(self) -> "ChoiceMap":
...
+                y = bernoulli(0.5) @ "y"
+                return x + y
+
+
+            chm = ChoiceMap.d({"y": 1, "z": 2})
+            extras = chm.invalid_subset(model, (1,))
+            assert "z" in extras  # "z" is an extra choice not in the model
+            ```
+        """
+        shape_chm = staging.get_trace_shape(gen_fn, args).get_choices()
```
littleredcomputer, line 1329: this turned out beautifully

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -372,6 +400,29 @@ def get_subselection(self, addr: ExtendedAddressComponent) -> Selection:
         return self
 
 
+@Pytree.dataclass
+class LeafSel(Selection):
```
littleredcomputer, line 404: Is there a case where this would screen out subtracts from a score? Or is this currently used for the validation code?

src/genjax/_src/generative_functions/combinators/vmap.py
```diff
@@ -137,49 +143,42 @@ def inner(*args):
...
-            if not len(args) == axes_length:
-                raise Exception(
-                    f"VmapCombinator requires that length of the provided in_axes kwarg match the number of arguments provided to the invocation.\nA mismatch occured with len(args) = {len(args)} and len(self.in_axes) = {axes_length}"
-                )
-
-    def _static_broadcast_dim_length(self, args):
-        def find_axis_size(axis, x):
+    @staticmethod
+    def _static_broadcast_dim_length(in_axes: InAxes, args: tuple[Any, ...]) -> int:
+        # We start by triggering a vmap to force all JAX validations to run. If we get past this line we know we have compatible dimensions.
```
littleredcomputer, line 148: Again it looks like a jit of simulate will wind up doing this? But this time the function does no work, so maybe that's ok.

src/genjax/_src/generative_functions/combinators/vmap.py
```diff
@@ -189,12 +188,11 @@ def generate(
         args: tuple[Any, ...],
     ) -> tuple[VmapTrace[R], Weight]:
         assert isinstance(constraint, ChoiceMapConstraint)
-        self._static_check_broadcastable(args)
-        broadcast_dim_length = self._static_broadcast_dim_length(args)
+        broadcast_dim_length = self._static_broadcast_dim_length(self.in_axes, args)
         idx_array = jnp.arange(0, broadcast_dim_length)
```
littleredcomputer, line 192: I'm kind of curious as to what's going on here so a comment might help. We're trying to figure out how many keys to feed to vmap...which has happened before, and so we want to regenerate it a second time?

littleredcomputer (approved):

PR #1371, opened 2024-09-27
convert smc_custom_proposal to genstudio (GEN-233)

mhuebert:
@MathieuHuot 

- I converted the doc to a `.py` file (LMK if that's ok, I can also convert back to `.ipynb`).
- For performance reasons I am saving the logo as an image, and using that in the studio plot using a new `Plot.img` mark. (Turns out the Observable Plot raster mark is much slower, takes 500ms to render pixels from data.)
- I converted the logo to an outline so that dots inside/outside are equally visible.
- requires the latest genstudio version (so be sure to `poetry install`)

https://github.com/user-attachments/assets/3fb98bf6-c4f9-4f4b-8052-1e74bf54308b


MathieuHuot (approved):
This is great @mhuebert, and I think we can merge this version on my branch. 
I will continue to develop the project to get to the point I wanted with rejuvenation and different strategies, and then we'll be able to try interactive inference!

MathieuHuot:
@mhuebert I notice you still haven't merged this.
I the mean time I rewrote most of the code to make it cleaner and actually work (there were a few math bugs, now it's fixed but SMC looks worse for some reason). I rewrote the jupyter notebook version though, it's what I'm more used to for running cells and debugging.

mhuebert:
@MathieuHuot I've revised this PR:
- changes are applied directly to the `.ipynb` file
- I introduced a `dots_plot` helper fn
The only thing I'm not sure about are the correct image paths.

PR #1372, opened 2024-09-27
(2 / N) Add `Regenerate` request to support a "regenerate" edit for `Distributions` and `StaticGenerativeFunction`

femtomc:


femtomc:
CC @sritchie & @MathieuHuot I'm going to introduce some redundancy (in terms of "hierarchical specification") in this PR which we'll stare at after I get it down.

It points towards a resolution which involves -- either:

(a) Merging choice map and selection, as we've informally discussed (treating a selection as a projection paired with a value, the value could be a constraint, or an edit request)
(b) Treating choice map as an abstract collection and re-visiting the separation between values and submaps.

femtomc:
@MathieuHuot: https://github.com/probcomp/genjax/pull/1372/files#diff-cf05070a53a8ff04406eda8ff4d429024ccba01460279145af0a3ff56a6679e0

Will add HMC after I get back from pickle ball.

MathieuHuot (commented):
@femtomc  so what is this doing exactly? 
My guess is that is that you want to regenerate `y1` and to do that you use the selection on `y1`. But then I'm not fully sure. 
- Is `genjax.Regenerate()` applying a default regeneration strategy, e.g. sample from the default proposal with `y2` fixed?
- Why is `SelectApply` a type of `EditRequest`? Why does it take an `EditRequest` as argument? One guess (that I think is wrong given what you explained) would be that a selection is a projection which is a type of `SMC` move as it'd be marginalisation in the sense of GenSP, here used with a syntax close to the compositional SMC library in Gen.

MathieuHuot:
> @femtomc so what is this doing exactly? My guess is that is that you want to regenerate `y1` and to do that you use the selection on `y1`. But then I'm not fully sure.
> 
> * Is `genjax.Regenerate()` applying a default regeneration strategy, e.g. sample from the default proposal with `y2` fixed?
> * Why is `SelectApply` a type of `EditRequest`? Why does it take an `EditRequest` as argument? One guess (that I think is wrong given what you explained) would be that a selection is a projection which is a type of `SMC` move as it'd be marginalisation in the sense of GenSP, here used with a syntax close to the compositional SMC library in Gen.

What I'd be expecting as syntax for an `EditRequest` would be something like
```
request(selection: Optional?, strategy : chm -> chm, weight, in_place: bool,  trace_translator: Optional)
```
where:
- `selection` is just an argument, potentially with a default to all
- `strategy`: the function taking a choice map and returning the changed choicemap as well as the incremental weight
- `in_place`: a parameter indicating whether to do the in-place update on the trace
- `trace_translator`: an optional function on choicemaps mimicking the behaviour of the trace translator DSL. Maybe these should just be separate `EditRequest` but I'm not sure.
 
Then you can apply a request to a trace and a key as you show, maybe with the argdiffs too, not sure about the general case.

There's still a heavy limitation in what I just wrote actually, as you want the strategy to possibly depend on many things, e.g. the generative function to do get gradient of densities for HMC.

femtomc:
Not sure how to debug what you're thinking about, but maybe we should talk next week in person.

regenerate: https://www.gen.dev/docs/stable/api/model/gfi/#Gen.regenerate

Can you ask Alex to send you his write up on `Gen.update(update)`? I think that will help.

It's possible your picture is an accurate picture of a lower-level programmable thing, but I can't tell.

MathieuHuot:
> Not sure how to debug what you're thinking about, but maybe we should talk next week in person.
> 
> regenerate: https://www.gen.dev/docs/stable/api/model/gfi/#Gen.regenerate
> 
> Can you ask Alex to send you his write up on `Gen.update(update)`? I think that will help.
> 
> It's possible your picture is an accurate picture of a lower-level programmable thing, but I can't tell.

Ahh sorry I got thrown off when you mentioned HMC and I confused myself. 
This `regenerate` is consistent with Gen's regenerate and is the first point I described above. 

My other points that still seem relevant given your answer:
- why is selection an editRequest?
- what about adding a parameter for in place update? This regenerate seems like a good test to try this out, or do you it's really a bigger story and we should attempt to look at that?

femtomc:
> why is selection an EditRequest?

Selection is not an EditRequest -- it's a way to indicate _what addressed random choices_ the (edit request passed as an argument) should apply to.

To be clear, we have `Regenerate()` -- it's a sort of indicator edit request. Distributions implement edit logic which respond to `Regenerate` (exactly the implementation of `regenerate` for `Distribution` in Gen.jl). Now, to support the application of a "regenerate" like functionality (see the interface I linked) -- we provide another `EditRequest` (`SelectApply(selection, primitive_edit_request)`) _which deals explicitly with indicating how `Regenerate` should be applied across a hierarchical structure_.

The benefit here is that -- we get to add other indicator logic (`MALA(eps)`, `HMC(eps, ...)` -- HMC is a maybe because I still need to study the hyperparameters -- and then you can immediately extend the move to work with generative functions which work with addressed random choices by using `SelectApply`.

Think of `SelectApply` as a simple (and, likely not the only) way to write a larger "edit" program which extends the logic of the primitives to work with e.g. structured generative functions.

> What about adding a parameter for in place update? This regenerate seems like a good test to try this out, or do you it's really a bigger story and we should attempt to look at that?

I don't know what you mean by in place?

femtomc:
I can show you with MALA, if you'd like.

Edit: ah -- so right, MALA and HMC can't follow this pattern, because the gradient computations involved in those moves require that the gradient contributions _from downstream random choices_ be accumulated.

MathieuHuot:
> Selection is not an EditRequest -- it's a way to indicate what addressed random choices the (edit request passed as an argument) should apply to.

Oh, I was super unclear. I meant `SelectApply`: why is that its own thing instead of having the Regenerate request take a Selection argument as in the Gen.jl version?

Ah, and reading above, it's a way to combine an edit strategy with a specific selection to apply it to, if I understand correctly.

MathieuHuot:
> I don't know what you mean by in place?

E.g. the imperative update of an array where you don't create a copy but modify this version "in place", using the `at`/`set` syntax from JAX.

MathieuHuot:
> Edit: ah -- so right, MALA and HMC can't follow this pattern, because the gradient computations involved in those moves require that the gradient contributions from downstream random choices be accumulated.

I don't get this yet. But do you still want to merge that first version before moving to more complex edit requests?

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -299,13 +304,62 @@ def project(
...
+            return request.edit(key, trace, argdiffs)
+        elif FlagOp.concrete_false(check):
+            return (
+                trace,
+                jnp.array(0.0),
+                Diff.no_change(trace.get_retval()),
+                EmptyRequest(),
+            )
+        else:
+            raise NotImplementedError
```
MathieuHuot, line 337: Why are there more than 2 cases here? isn't `check` always True or False?

MathieuHuot: I also don't get the logic of why this works and does anything correctly for the examples.

femtomc:
> I don't get this yet. But do you still want to merge that first version before moving to more complex edit requests?

Let's just merge `Regenerate` for `Distribution` and `StaticGenerativeFunction` ... that's a good minimal PR.

I removed `SelectApply`, pushed `Selection` into `Regenerate` -- let's merge that version _now_ -- and then we can author `MALA` and `HMC` following the implementations in Gen.jl

I have more thoughts on composition of multiple requests which we should discuss, I'll setup a PR for that too.

src/genjax/_src/generative_functions/distributions/distribution.py
femtomc: `concrete` means "known at compile time"

It's possible that the value there is not actually known at compile time -- in that case, we currently error ... but we can extend the logic to use a `jax.lax.cond`

src/genjax/_src/core/generative/requests.py
```diff
@@ -0,0 +1,76 @@
...
+class EmptyRequest(EditRequest):
+    def edit(
+        self,
+        key: PRNGKey,
+        tr: Trace[R],
+        argdiffs: Argdiffs,
+    ) -> tuple[Trace[R], Weight, Retdiff[R], "EditRequest"]:
+        return tr, jnp.array(0.0), Diff.no_change(tr.get_retval()), EmptyRequest()
+
+
```
femtomc, line 50: @MathieuHuot See below for the specification of just `Regenerate` -- I've removed `SelectApply`, and instead just focused on "low risk" implementations from Gen.jl

src/genjax/_src/generative_functions/distributions/distribution.py
MathieuHuot: ah I see, makes sense! This deals with the efficiency of the produced code, and often we expect this pattern to be known at compile time, and avoiding whenever possible `jax.lax.cond` is nice.

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -299,13 +306,53 @@ def project(
             jnp.array(0.0),
         )
 
-    def edit_incremental_generic_request(
+    def edit_regenerate(
```
MathieuHuot, line 309: There's still something I don't get if one would want to apply it to a static choice map coming  from the model
```
def simple_normal():
            y1 = genjax.normal(0.0, 1.0) @ "y1"
            y2 = genjax.normal(0.0, 1.0) @ "y2"
            return y1 + y2
```
Here, if I want to regenerate `y2` but not `y1`, I need to call the edit request with the selection on the model, which should call edit on each submodel and accumulate things.
Arriving at edit for `y1`, it should see that `y1` is not in the selection and go to case 2 in the function below, while when arriving at edit for `y2` it should trigger the first case of the function below. 
Is that happening? What exactly is the spec of `check = () in selection`?

femtomc: Yes, let me explain -- the syntax `() in selection` is checking if the "leaf" (the distribution) is in the selection.

It's a bit of a peculiarity based on the fact that so much of the automation is based around "addressed random choices" -- the point being: we choose `()` to be the "address of a value" -- the process you are describing occurs, but then we want to check "am I in the selection?" for a distribution.

@MathieuHuot It would be nice if we unified the "submap" vs. "value" distinction for addressed random choices -- and I think if we do so, we'll be able to simplify the logic here.

tests/inference/test_requests.py
```diff
@@ -0,0 +1,70 @@
...
+        new_v = new_tr.get_choices()["y2"]
+        assert old_v != new_v
+        old_tr, bwd_w, _, bwd_request = bwd_request.edit(key, new_tr, ())
+        assert bwd_w != 0.0
+        assert (fwd_w + bwd_w) == 0.0
+        old_old_v = old_tr.get_choices()["y2"]
+        assert old_old_v == old_v
+
+        # What about both?
+        old_v = tr.get_choices()["y2"]
```
femtomc, line 60: @MathieuHuot Here's the change -- removed `SelectApply`, added `Regenerate` (with a selection).

MathieuHuot: ah yeah, that's nice, cool!

MathieuHuot (approved):

src/genjax/_src/generative_functions/distributions/distribution.py
MathieuHuot: Makes sense re the explanation of `() in selection`.
Re the simplification, I don't fully get the details of how much simplification we could further have but I'd love that to happen yes.

femtomc: @MathieuHuot I'll send another PR on that, soon

PR #1374, opened 2024-09-29
(3 / N) Add ProposalUpdate edit request.

femtomc:


src/genjax/_src/core/generative/requests.py
```diff
@@ -62,6 +68,27 @@ def edit(
         return gen_fn.edit(key, tr, self, argdiffs)
```
femtomc, line 70: CC @MathieuHuot we gotta eyeball the weight math on this one, I'm 99% sure it's not correct.

femtomc: (This is an `EditRequest` which bundles a proposal -- first uses the proposal to propose changes, and then updates the trace with them, but then we need a contribution in the weight from the proposal for the discarded random choices)

I think it's `model_weight_from_edit + (proposal_tr.get_score() - proposal.assess(discarded_choices_under_model))` -- but in the code I used `proposal.edit` to try and compute the second contribution directly using `edit`!

MathieuHuot: On top of checking the math, we also need to add tests for this.

femtomc:
Handled in #1411.

PR #1382, opened 2024-10-07
Simplify SwitchCombinator implementation, implement `project` for `switch` (GEN-640)

sritchie:
This PR:

- adds a `multi_switch` to `staging`, for switching between methods with different output shapes — this is used by `SwitchCombinator` to implement all methods
- re-implements `SwitchCombinator`'s methods using `multi_switch`
- implements `project` for `SwitchCombinator`
- renames `get_trace_shape` on `GenerativeFunction` to `get_zero_trace`, and modifies it to return a trace filled with zeros instead of JAX structs
- deletes the now-unneeded-by-switch `get_*_shape` functions from `staging`
- moves `staged_choose` to `genjax._src.core.interpreters.staging` and renames it to `tree_choose`
- fixes two warnings that had been showing up in the tests
- updates the jupyterlab dev dependency to fix the `safety` warning


src/genjax/_src/core/generative/core.py
```diff
@@ -174,7 +174,7 @@ def edit(
         key: PRNGKey,
         tr: "genjax.Trace[R]",
         argdiffs: Argdiffs,
-    ) -> tuple["genjax.Trace[R]", Weight, Retdiff[R], "EditRequest"]:
+    ) -> "tuple[genjax.Trace[R], Weight, Retdiff[R], EditRequest]":
```
sritchie, line 177: making the whole type a string fixes the beartype warning we've been getting in testing.

pyproject.toml
```diff
@@ -65,7 +65,7 @@ ruff = "^0.1.3"
 safety = "^2.3.5"
 seaborn = "^0.12.1"
 xdoctest = "^1.1.0"
-jupyterlab = "^4.1.6"
+jupyterlab = "^4.2.5"
```
sritchie, line 68: this fixes the new safety warning that popped up in CI.

femtomc (approved):
Love the simplifications — let’s fix the edit math + bwd request in follow up PR.

PR #1383, opened 2024-10-07
feat: add tfp Dirichlet

limarta:
This commit exports tfp's [Dirichlet distribution](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Dirichlet).


MathieuHuot (approved):
ok, straightforward enough.

femtomc:
@limarta Also add the `dirchlet` to the `__all__` list in the module (where all the imports come in and leave in the public API)

PR #1385, opened 2024-10-08
fix Diff methods, add tests (GEN-242)

sritchie:
This PR:

- removes `Diff.tree_diff_no_change` and `Diff.tree_diff_unknown_change` in favor of the identical `Diff.no_change` and `Diff.unknown_change` methods
- fixes the type on the `tangent` field of `Diff` to be `ChangeTangent`
- adds docs and tests for all methods in `Diff`
- fixes `Diff.tree_tangent` to return `NoChange` for any non-Diff value
- `Diff.static_check_no_change` will now fail if any of the leaves are not explicitly `NoChange`


femtomc:
@sritchie:
> fixes Diff.tree_tangent to return UnknownChange for any non-Diff value

this is wrong -- we should assume that a non-Diff value is assumed to be a constant (in other words, `NoChange`) (like in forward mode AD, if you encounter a value which is not a tangent participating in a forward mode computation, you assume it doesn't have a tangent -- a dual number component of 0).

femtomc (approved):

PR #1387, opened 2024-10-08
Tidy `scan`, `vmap` combinators (3/N)

sritchie:
This PR:
- removes a `vmap` during choicemap creation in both combinators, preparing us for a simpler choicemap world
- simplifies away some of the unnecessary internal functions

This is prep-work for removing `Indexed`...


MathieuHuot (changes requested):
@sritchie 
- I think the comments to McCoy make sense: we need to redesign the thing anyway to consider batched traces (where an external vmap could've been applied)
- The simplifications look good. Haven't checked every detail but if the tests pass seems good.
- I'm pretty sure `VmapTrace` doesn't do anything. Can we just get rid of it now? The only non-obvious part to me is whether it is somehow providing some useful debugging info for users. But in this case, I'm sure there's a better error message we can give that doesn't require us creating a class. Same thing with `ScanTrace`.

sritchie:
@MathieuHuot great, I made an issue to remove `ScanTrace` and `VmapTrace` in a separate PR: https://linear.app/chi-fro/issue/GEN-669/remove-vmaptrace-and-scantrace-in-favor-of-using-statictrace we'll see if we run into trouble but I agree that we should not, as long as no one is doing an instance check.

If so, they can check the type of the wrapped `gen_fn`.

PR #1389, opened 2024-10-08
Restrict Indexed lookups in ChoiceMap to working cases (5/N)

sritchie:
This PR:

- restricts `DynamicAddressComponent` to `int | IntArray`, the only two values we can actually handle via choicemap
- restricts the dynamic addresses passed to `get_submap` of `Indexed` to be scalar arrays or `int`s. The "pass a vector" case was broken and best handled by some sort of slicing support, which we don't have
- fix the case where `self.addr` in `Indexed` is an out-of-order array / not the output of `np.arange`.


src/genjax/_src/core/generative/choice_map.py
```diff
@@ -53,14 +52,14 @@
 #################
 
 StaticAddressComponent = String
-DynamicAddressComponent = ArrayLike
+DynamicAddressComponent = Int | IntArray
 AddressComponent = StaticAddressComponent | DynamicAddressComponent
-Address = tuple[()] | tuple[AddressComponent, ...]
```
sritchie, line 56: the empty tuple is covered by the `...`

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1389,6 +1388,8 @@ class Indexed(ChoiceMap):
     def build(chm: ChoiceMap, addr: DynamicAddressComponent) -> ChoiceMap:
         if chm.static_is_empty():
             return chm
+        elif isinstance(addr, Array) and addr.shape == (0,):
```
sritchie, line 1390: this guards against the case of scans or vmaps over an empty array.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1399,32 +1400,18 @@ def get_submap(self, addr: ExtendedAddressComponent) -> ChoiceMap:
...
-            check = (
-                jax.vmap(check_fn, in_axes=(None, 0))(addr, self.addr)
-                if jnp.array(self.addr, copy=False).shape
-                else check_fn(addr, self.addr)
-            )
-
-            check_array = jnp.asarray(check, copy=False)
-            if check_array.shape:
-                if check_array.shape[0] == 0:
-                    # this is an obscure case which can arise when doing an importance
```
sritchie, line 1417: this is now handled by the `build` check.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1399,32 +1400,18 @@ def get_submap(self, addr: ExtendedAddressComponent) -> ChoiceMap:
...
         else:
+            assert not jnp.asarray(
+                addr, copy=False
+            ).shape, "Only scalar dynamic addresses are supported by get_submap."
 
-            def check_fn(idx, addr) -> Flag:
-                return idx == addr
-
-            check = (
-                jax.vmap(check_fn, in_axes=(None, 0))(addr, self.addr)
```
sritchie, line 1409: this branch was always busted in the case of a non-scalar `addr`... non-scalar `addr` compared via `==` with `self.addr` was a recipe for trouble.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1399,32 +1400,18 @@ def get_submap(self, addr: ExtendedAddressComponent) -> ChoiceMap:
...
-
-            check_array = jnp.asarray(check, copy=False)
-            if check_array.shape:
-                if check_array.shape[0] == 0:
-                    # this is an obscure case which can arise when doing an importance
-                    # update of a scan GF with an array of shape (0,) or (0, ...)
-                    return ChoiceMap.empty()
-                else:
-                    return jtu.tree_map(lambda v: v[addr], self.c).mask(
-                        check_array[addr]
```
sritchie, line 1422: we don't need the mask - `addr` can only be a scalar, and `__getitem__` on the JAX arrays can handle traced inputs.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1399,32 +1400,18 @@ def get_submap(self, addr: ExtendedAddressComponent) -> ChoiceMap:
...
-                    )
+            if isinstance(self.addr, Array) and self.addr.shape:
+                return jtu.tree_map(lambda v: v[addr], self.c)
             else:
-                return self.c.mask(check)
```
sritchie, line 1425: we DO need the mask here in the scalar <-> scalar case, because we should have a scalar at the leaves.

tests/core/generative/test_core.py
```diff
@@ -81,11 +81,6 @@ def model(x):
         assert jnp.array_equal(chm[..., "v"], varr)
         assert jnp.array_equal(chm[..., "q"], qarr)
 
-        # check alternate access route:
-        assert jnp.array_equal(chm(jnp.arange(3))["v"].unmask(), varr)
```
sritchie, line 83: no longer allowed!

tests/core/test_choice_maps.py
```diff
@@ -528,9 +528,9 @@ def test_extend_dynamic(self):
         extended = chm.extend(jnp.array([0, 1, 2]))
         assert extended.get_value() is None
         assert extended.get_submap("x").static_is_empty()
-        assert extended[0].unmask() == 2.3
```
sritchie, line 531: phew, much better without the unmask everywhere

tests/generative_functions/test_vmap_combinator.py
```diff
@@ -131,27 +131,29 @@ def foo(loc: float, scale: float):
             ValueError,
             match="vmap was requested to map its argument along axis 0, which implies that its rank should be at least 1, but is only 0",
         ):
-            foo.vmap(in_axes=(0, None)).simulate(key, (10.0, jnp.arange(3.0)))
+            jax.jit(foo.vmap(in_axes=(0, None)).simulate)(key, (10.0, jnp.arange(3.0)))
```
sritchie, line 160: @littleredcomputer in his review had requested that we make sure these validations survive jit.

tests/generative_functions/test_vmap_combinator.py
```diff
@@ -174,3 +176,17 @@ def model(x):
 
         # the inner vmap has vmap'd over the y's
         assert chm[..., "y"].shape == (10, 5)
+
+    def test_zero_length_vmap(self):
```
sritchie, line 206: this tests the analogous case to the `scan` test that @littleredcomputer introduced.

MathieuHuot:
@sritchie Could you quickly comment regarding "removes the un-needed mask call when we have array-shaped leaves". I don't understand what in the code changed that enables this simplification. I do see that a bunch of tests are simplified but what happened?

MathieuHuot (approved):

sritchie:
> @sritchie Could you quickly comment regarding "removes the un-needed mask call when we have array-shaped leaves". I don't understand what in the code changed that enables this simplification. I do see that a bunch of tests are simplified but what happened?

@MathieuHuot when we were pairing you and I realized that removing this `mask` call introduces a bug if the user runs off the end of the array (since `jnp` clamps the indices). I've added the `mask` call back, AND I added tests and a fix for a case where the `Indexed` array's addresses are not the output of `arange`.

We will remove `Indexed` eventually, but we will still need the logic from this bug fix in whatever sparse array solution we come up with, so I figured it was worth adding here now.

PR #1390, opened 2024-10-08
remove ChoiceMapConstraint inheritance from ChoiceMap (6/N)

sritchie:
This PR:

- Modifies `ChoiceMapConstraint` to NOT inherit from `ChoiceMap`, preventing accidental wrapping layers
- modifies `Update` to take a `ChoiceMap`
- This PR removes the explicit `Sample` type in favor of using `ChoiceMap` in all places `Sample` was used before. (merged in from #1400)


femtomc:
Strongly disagree, and I've argued why in several threads. Even inheriting from `Sample` is a stretch -- by my same line of argument: a `ChoiceMap` can have _`EditRequest`_ leaves ... such a thing is _not a `Constraint`_ (it's not even a `Sample` -- so we're already "violating the types" by allowing that!). I don't care how simple it makes the code, it's confusing specific mathematical constructs.

Users aren't reaching for `edit` -- they'll be reaching for `update` or `regenerate` first -- derived interfaces on generative function which hide the "advanced types" from them... If they're ready for `edit`, they'll be ready to understand what a `Constraint` is -- and we'll be ready to explain it to them.

My philosophy: we need to keep the mathematical core correct -- we can layer whatever fancy sugar we want to make that core opaque to the user on top of that.

sritchie:
Okay, excellent, let's use this PR to collect the discussion. I will write more tomorrow - is it really true that the current choicemap can have editrequest leaves? What would
The indexing mean there?

When Indexed is gone ChoiceMap will end up being a wrapper around a dictionary with arraylike leaves. We can make other wrappers around dictionaries for EditRequest leaves, yeah? No need to overload the one type with no way for the type system to distinguish.

MathieuHuot (changes requested):
I'm still thinking whether we want that to happen or not, but what we all agreed on is that we should remove `Sample`.

It should be easier in this PR as you already made ChoiceMap inherits Constraint so `Sample` should be plain dead code at this point.

sritchie:
Okay, @MathieuHuot I will remove `Sample`!

sritchie:
Verbal +1 from @femtomc

PR #1392, opened 2024-10-08
Add slicing support to ChoiceMap and ChoiceMapBuilder (GEN-424, GEN-587 )

sritchie:
This PR:

- replaces remaining `PRNGKey` uses with `jax.random.key`
- adds `:` support to `ChoiceMap`, both in the `C[:, "x", :]` builder notation and in lookups, like `chm[:, "x", :]`
- adds a validation for making sure that users don't provide invalid selection sequences.

The validation applies to the address supplied in `ChoiceMap` creation via `C[<address>]`, and selection via `chm[<address>]`.

The rules are, after filtering out string addresses:
- you are allowed any number of `int` or `ScalarInt` address components, because drop a dimension at each `get_submap` call.
- afterward, you are allowed a single non-scalar array. (On lookup, this single entry could also be a partial slice, like `1:3`)
- the tail of the address must be all full slices, i.e. `:`

that is not FULL slice support, but it's pretty general and should help our users figure out how to create 2d slices!

```python
C[:, "x", :].set(<2d array>)
```

### DIFF
--- a/notebooks/cookbook/basics/choicemap_creation_selection.ipynb
+++ b/notebooks/cookbook/basics/choicemap_creation_selection.ipynb
@@ -40,7 +40,7 @@
     ")\n",
     "\n",
     "pretty()\n",
-    "key = random.PRNGKey(0)"
+    "key = random.key(0)"
    ]
   },
   {
@@ -78,7 +78,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "key = jax.random.PRNGKey(0)\n",
+    "key = jax.random.key(0)\n",
     "trace = jax.jit(beta_bernoulli_process.simulate)(key, (0.5,))"
    ]
   },
--- a/notebooks/cookbook/basics/choicemap_creation_selection.ipynb
+++ b/notebooks/cookbook/basics/choicemap_creation_selection.ipynb
@@ -272,34 +272,6 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "But because of the nested `vmap`, the address hierarchy can sometimes lead to unintuitive results. "
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "chm = chm ^ C[1, 3, \"new_pixel\"].set(1.0)  # seemingly adding a new constraint\n",
-    "tr, w = jax.jit(sample_image.importance)(key, chm, (image,))\n",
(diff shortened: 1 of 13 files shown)


src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1404,12 +1504,18 @@ def get_submap(self, addr: ExtendedAddressComponent) -> ChoiceMap:
...
-            ).shape, "Only scalar dynamic addresses are supported by get_submap."
+            if not isinstance(addr, slice):
+                assert not jnp.asarray(
+                    addr, copy=False
+                ).shape, "Only scalar dynamic addresses are supported by get_submap."
 
-            if isinstance(self.addr, Array) and self.addr.shape:
+            if self.addr is None:
                 return jtu.tree_map(lambda v: v[addr], self.c)
+                # TODO PUT THE MASK BACK ON!!!
```
sritchie: I don't think we need to do this, actually, more soon

notebooks/cookbook/basics/choicemap_creation_selection.ipynb
```diff
@@ -292,13 +292,8 @@
...
    "outputs": [],
    "source": [
-    "N = 4\n",
-    "F = 4\n",
-    "\n",
-    "\n",
-    "chm3 = C[\n",
-    "    jnp.arange(F), jnp.repeat(jnp.arange(N)[jnp.newaxis], F, axis=0), \"new_pixel\"\n",
-    "].set(1.0)"
+    "# TODO this is broken, since we need to set a proper matrix at the bottom.\n",
```
sritchie: @MathieuHuot , thoughts here?

MathieuHuot: @sritchie did you fix the lines right below that are right above the lines you're commenting about:
```
chm = chm ^ C[1, 3, "new_pixel"].set(1.0)  # seemingly adding a new constraint
tr, w = jax.jit(sample_image.importance)(key, chm, (image,))
w  # Yet we obtain the same weight as before
```
which were there to illustrate "But because of the nested `vmap`, the address hierarchy can sometimes lead to unintuitive results."
If so, then I suggest we remove these 2 examples showing that choicemaps for `vmap` are busted as it won't be true anymore :).

MathieuHuot (changes requested):
Looks great overall, after you answer my comment we can merge and onto the next one!

MathieuHuot (approved):

PR #1394, opened 2024-10-09
Replace `jax.random.split` in `StaticGenerativeFunction` with `jax.random.fold_in` - a faster primitive

femtomc:
https://jax.readthedocs.io/en/latest/_autosummary/jax.random.fold_in.html
https://github.com/jax-ml/jax/discussions/12395#discussioncomment-3703291

The number of key splits which occur in a static generative function -- for any interface -- is constant. Right now, we're splitting and producing _1 extra key_ each time we run the interface (which we're hoping JAX DCEs for us).

Instead, we can replace the key splitting with `fold_in` to ensure that we're generating exactly the number of keys we need.

Also, refactor tests which depend on hardcoded constants to instead test trace invariants.


src/genjax/_src/generative_functions/static.py
```diff
@@ -575,7 +583,7 @@ def handle_trace(
         self.visit(addr)
         subtrace = self.get_subtrace(addr)
         subrequest = self.get_subrequest(addr)
-        self.key, sub_key = jax.random.split(self.key)
+        sub_key = jax.random.fold_in(self.key, self.key_counter)
```
sritchie: can you make a method that both

- performs fold_in
- does the increment?

tests/generative_functions/test_dimap_combinator.py
```diff
@@ -43,7 +43,7 @@ def model(x, y, _):
         key = jax.random.PRNGKey(0)
         trace = dimap_model.simulate(key, (2.0, 3.0))
         assert (
-            -2.5092335 == trace.get_retval()
+            trace.get_retval() == trace.get_choices()["z"] + 2.0
```
sritchie, line 46: nice!

sritchie (commented):
One suggestion...

sritchie (approved):

PR #1395, opened 2024-10-09
Alternative take on fixing `Indexed` (and adding `Vector`) choice map.

femtomc:


sritchie:
@femtomc , mind if we close this one?

PR #1397, opened 2024-10-09
Tighten type bounds on Choice, see the consequences

sritchie:
This PR explores what happens if we lock down the type bounds on `Choice` to be `ArrayLike | Mask[ArrayLike]`.

TODO:
- [ ] it looks like inference breaks, and we have code that assumes that we can have a choice that is itself another choicemap. Maybe this is a case, @MathieuHuot , where we are already supporting an arbitrary pytree.
- [ ] make yet another ticket discussing this

### DIFF
--- a/src/genjax/_src/core/generative/choice_map.py
+++ b/src/genjax/_src/core/generative/choice_map.py
@@ -857,7 +857,7 @@ class ChoiceMap(Sample):
         return _empty
 
     @staticmethod
-    def choice(v: T) -> "Choice[T]":
+    def choice(v: ArrayLike | Mask[ArrayLike]) -> "ChoiceMap":
         """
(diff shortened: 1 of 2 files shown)


MathieuHuot:
@sritchie  sorry I'm confused. Per our written document on Notion, 
- do we agree that Static choicemaps are nested dictionaries?
- but the values are always, at least after simplify, ArrayLike?
- there may be some simplifications problems before we resolve Xor/Or/Indexed/Filtered, but what breaks exactly? One restriction for inference is that we said we'll need a more fine grained language to construct sparse constraints to be used in inference, while choicemaps are fully dense in our current models.

femtomc:
What breaks is that GenSP supports distributions whose choice maps are `Choice` with choice maps as values.

PR #1398, opened 2024-10-10
prefer key to PRNGKey

sritchie:
According to https://jax.readthedocs.io/en/latest/_autosummary/jax.random.PRNGKey.html#jax.random.PRNGKey we should prefer `jax.random.key` to `jax.random.PRNGKey`.


eightysteele:
@sritchie nice find, +1. 

one interesting caveat i found: https://jax.readthedocs.io/en/latest/jep/9263-typed-keys.html#what-does-this-mean-for-users

> If you call a JAX-based library which does not yet handle typed PRNG keys, you can use raw_key = jax.random.key_data(key) for now to recover the raw buffer, but please keep a TODO to remove this once the downstream library supports typed RNG keys.

i wonder if there's a potential breakage in user code that might call out to such libraries? probably a small chance, but we might consider explicitly flagging this for users in the next release.

eightysteele (approved):

PR #1401, opened 2024-10-14
Add `_ChoiceMapBuilder.update`, `Choice.build`

sritchie:
This PR:

- adds an `update` method to `ChoiceMapBuilder`, allowing the user to do `chm.at[0, "x"].update(lambda old_value: <produce new value>)`
- adds `Choice.build`, which guards against a 0-sized array or a concretely-False masked value and returns `ChoiceMap.empty()` instead

### DIFF
--- a/src/genjax/_src/core/generative/choice_map.py
+++ b/src/genjax/_src/core/generative/choice_map.py
@@ -84,7 +84,7 @@ SelectionBuilder = _SelectionBuilder()
 """Deprecated! please use `Selection.at`."""
 
 
-class Selection(Projection["ChoiceMap"]):
+class Selection(Projection["ChoiceMap"], Pytree):
     """
     A class representing a selection of addresses in a ChoiceMap.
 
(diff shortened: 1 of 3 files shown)


src/genjax/_src/core/generative/choice_map.py
```diff
@@ -737,6 +737,37 @@ def set(self, v) -> "ChoiceMap":
...
+            assert updated["y", "z"] == 20
+            # Updating a non-existent address
+            new_chm = chm.at["w"].update(lambda _: 42)
+            assert new_chm["w"] == 42
+            ```
+        """
+        if self.choice_map is None:
+            return self.set(f(None))
+        else:
+            submap = self.choice_map(tuple(self.addrs))
```
sritchie, line 765: @littleredcomputer , do you think that we should 

- keep the signature I have here for the `update` fn?
- pass `None` if there is no submap, vs the empty submap? (I think no)...
just wanted to gut-check that this feels good.

littleredcomputer (approved):

PR #1404, opened 2024-10-17
(1 / N) Remove `ChoiceMapEditRequest`, replace with `StaticRequest`

femtomc:
Eliminate source of strife in the form of deleting `ChoiceMapEditRequest` -- add back in hierarchical `EditRequest` via `StaticRequest` (lives off `StaticGenerativeFunction`). Allows creating hierarchical (compositional!) edit requests:

```python
request = StaticRequest({
    "y1": Regenerate(S.all()),
    "y2": Update(C.v(3.0)),
})
```

### DIFF
--- a/src/genjax/_src/generative_functions/static.py
+++ b/src/genjax/_src/generative_functions/static.py
@@ -24,9 +24,9 @@ from genjax._src.core.generative import (
     Argdiffs,
     ChoiceMap,
     ChoiceMapConstraint,
-    ChoiceMapEditRequest,
     Constraint,
     EditRequest,
+    EmptyRequest,
     GenerativeFunction,
     NotSupportedEditRequest,
     Projection,
@@ -120,6 +120,25 @@ class StaticTrace(Generic[R], Trace[R]):
         addresses = self.addresses.get_visited()
         idx = addresses.index(addr)
         return self.subtraces[idx]
+
+
+####################################
+# Static (trie-like) edit request  #
+####################################
+
+
+@Pytree.dataclass(match_args=True)
+class StaticEditRequest(EditRequest):
+    addressed: dict[StaticAddress, EditRequest]
+
+    def edit(
+        self,
+        key: PRNGKey,
+        tr: Trace[R],
+        argdiffs: Argdiffs,
+    ) -> tuple[Trace[R], Weight, Retdiff[R], "EditRequest"]:
+        gen_fn = tr.get_gen_fn()
+        return gen_fn.edit(key, tr, self, argdiffs)
 
 
 ##############################
@@ -544,10 +563,10 @@ def choice_map_change_transform(source_fn):
 
 
 @dataclass
-class ChoiceMapEditRequestHandler(StaticHandler):
+class StaticEditRequestHandler(StaticHandler):
     key: PRNGKey
     previous_trace: StaticTrace[Any]
-    requests_choice_map: ChoiceMap
+    addressed: dict[StaticAddress, EditRequest]
     address_visitor: AddressVisitor = Pytree.field(default_factory=AddressVisitor)
     score: FloatArray = Pytree.field(default_factory=lambda: jnp.zeros(()))
     weight: FloatArray = Pytree.field(default_factory=lambda: jnp.zeros(()))
@@ -573,8 +592,8 @@ class ChoiceMapEditRequestHandler(StaticHandler):
         self.address_visitor.visit(addr)
 
     def get_subrequest(self, addr: StaticAddress) -> EditRequest:
-        submap = self.requests_choice_map(addr)
-        return ChoiceMapEditRequest(submap)
+        subrequest = self.addressed.get(addr, EmptyRequest())
+        return subrequest
 
     def get_subtrace(
         self,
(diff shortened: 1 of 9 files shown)


src/genjax/_src/generative_functions/static.py
```diff
@@ -85,6 +85,10 @@ def get_visited(self):
         return self.visited
 
 
+def collapse_address(addr):
```
sritchie: wdyt of putting types on here, since we have those oh-so-nice types in `choice_map.py`? We could move them to `types.py` too to make it easier.

src/genjax/_src/generative_functions/static.py
```diff
@@ -122,6 +126,25 @@ def get_subtrace(self, addr: StaticAddress):
...
 
 
+####################################
+# Static (trie-like) edit request  #
+####################################
+
+
+@Pytree.dataclass(match_args=True)
+class StaticRequest(EditRequest):
+    addressed: dict[StaticAddressComponent | StaticAddress, EditRequest]
```
sritchie: do you want to support nesting? You might use

```python
from typing import TypeAlias

MyDict: TypeAlias = dict[StaticAddressComponent, "EditRequest | MyDict"]
```

src/genjax/_src/generative_functions/static.py
```diff
@@ -544,10 +567,10 @@ def wrapper(
 
 
 @dataclass
-class ChoiceMapEditRequestHandler(StaticHandler):
+class StaticEditRequestHandler(StaticHandler):
     key: PRNGKey
     previous_trace: StaticTrace[Any]
-    requests_choice_map: ChoiceMap
+    addressed: dict[StaticAddressComponent | StaticAddress, EditRequest]
     address_visitor: AddressVisitor = Pytree.field(default_factory=AddressVisitor)
```
sritchie, line 579: same note here for this type

src/genjax/_src/generative_functions/static.py
```diff
@@ -609,16 +634,18 @@ def handle_trace(
...
 
 
-def choice_map_edit_request_transform(source_fn):
+def static_edit_request_transform(source_fn):
     @functools.wraps(source_fn)
     def wrapper(
         key: PRNGKey,
         previous_trace: StaticTrace[R],
-        requests_choice_map: ChoiceMap,
+        addressed: dict[StaticAddressComponent | StaticAddress, EditRequest],
```
sritchie, line 647: whatever we do let's capture the type in an alias.

tests/generative_functions/test_static_gen_fn.py
```diff
@@ -748,6 +749,79 @@ def model(x, y, z: FloatArray | None = None):
...
+    def test_static_edit_request_composition(self):
+        @genjax.gen
+        def simple_normal():
+            y1 = genjax.normal(0.0, 1.0) @ "y1"
+            y2 = genjax.normal(0.0, 1.0) @ "y2"
+            return y1 + y2
+
+        key = jax.random.key(0)
+        tr = simple_normal.simulate(key, ())
+        request = StaticRequest({
```
sritchie, line 762: this is looking good!

tests/generative_functions/test_static_gen_fn.py
```diff
@@ -748,6 +749,79 @@ def model(x, y, z: FloatArray | None = None):
...
+    def test_static_edit_request_with_tuple_addr(self):
+        @genjax.gen
+        def simple_normal():
+            y1 = genjax.normal(0.0, 1.0) @ ("y1", "y3")
+            y2 = genjax.normal(0.0, 1.0) @ "y2"
+            return y1 + y2
+
+        key = jax.random.key(0)
+        tr = simple_normal.simulate(key, ())
+        request = StaticRequest({
```
sritchie, line 784: does this also work with

```python
        request = StaticRequest({
            "y1": {"y3": Regenerate(S.all())},
            "y2": Update(C.v(3.0)),
        })

sritchie (commented):
Couple notes about typing and nesting.

src/genjax/_src/generative_functions/static.py
femtomc: I'm inclined to be strict about this now, and support syntax which is convenient (like nested dictionaries), but not put nesting in the representation.

The point is -- a static generative function has a known set of addresses for callees (the addresses can be tuples, etc!) and the specification in our type should match the address set, and provide an EditRequest for each callee.

I think nesting should purely be a syntactical convenience, but it shouldn't affect this picture above.

tests/generative_functions/test_static_gen_fn.py
femtomc: No, you have to use the address which is directly referenced in the code of the generative function for now.

src/genjax/_src/generative_functions/static.py
femtomc: I slapped the static address types on this one, the return value type is a bit funky (it's the same as the input type 😆)

tests/generative_functions/test_static_gen_fn.py
MathieuHuot: @sritchie for now the syntax is clumsy for users but I think the concepts are good (McCoy and I discussed it this morning). After we solidify these a bit more it should be easy to write some convenience for users to express more compactly these complex structures. 
McCoy showed me an example where the trace structure was something like `("x", 4, "y", 5, "vertex")` which for now looks kinda horrible to write manually as a request, but it's fine.

sritchie (approved):

MathieuHuot (approved):

PR #1405, opened 2024-10-17
(7 / N) Add `IndexRequest` edit to `ScanCombinator`

femtomc:


tests/generative_functions/test_scan_combinator.py
```diff
@@ -454,3 +455,84 @@ def model(x, _):
...
+                return z, None
+
+            y1 = genjax.normal(0.0, 1.0) @ "y1"
+            _ = genjax.normal(0.0, 1.0) @ "y2"
+            return kernel.scan(n=10)(y1, None) @ "kernel"
+
+        key, sub_key = jax.random.split(key)
+        tr = scanned_normal.simulate(sub_key, ())
+        for idx in range(10):
+            # First, try indices and test for correctness.
```
MathieuHuot: This line is there 3 times.

tests/generative_functions/test_scan_combinator.py
```diff
@@ -454,3 +455,84 @@ def model(x, _):
...
+            old_z = tr.get_choices()["kernel", idx, "z"]
+            old_target_density = genjax.normal.logpdf(old_z, 0.0, 1.0)
+            request = StaticRequest({
+                "kernel": Checkerboard(Regenerate(S.at["z"])),
+            })
+            new_tr, fwd_w, _, _ = request.edit(key, tr, ())
+            new_z = new_tr.get_choices()["kernel", idx, "z"]
+            new_target_density = genjax.normal.logpdf(new_z, 0.0, 1.0)
+
+            # TODO: this needs to pass for Checkerboard to work soundly!
```
MathieuHuot: so if we now pass tihs test, we should remove the TODO and maybe explain what the test is checking?

src/genjax/_src/generative_functions/combinators/scan.py
```diff
@@ -311,7 +332,192 @@ def _project(
...
+            lambda v1, v2: jnp.where(idx_array == idx, v1, v2),
+            slice_scanned_out,
+            old_scanned_out,
+        )
+        new_scanned_retdiff = Diff.unknown_change(new_scanned_out)
+        max_length = self._static_scan_length(scanned_in, self.length)
+
+        # Mutate the trace in-place.
+        # Underneath JIT, should compile to an in-place update.
+        def mutator(v, idx, setter):
```
MathieuHuot, line 389: I think we discussed this in person: we should be consistent about when we decide to update in place or not.
Everything else in the code right now doesn't do in place changes so we should probably have a static arg in the editrequest asking to do the in-place change or not (with default to False?).

MathieuHuot: I think you mentioned that you had done a mixed of in-place and not-inplace in this editRequest, and I don't know if it was fixed or not. So just a ping on that?

MathieuHuot: After PLDI as it's not urgent we could ask this in-place argument to be propagated to the subrequests. Every request can in theory be chosen to be done in-place or not. But before that, I'd like to think a bit more about whether it's always correct to do it or not. If so, the default could be everything in-place and not in-place would mostly be useful for debugging.
And even if in-place could be done in theory, in practice it will break assumptions we make in certain parts of the codebase when we check the old_score from a trace and things like that.

MathieuHuot (approved):

PR #1406, opened 2024-10-17
(3 / N) Add an edit request which can be used to forcibly coerce Diff values according to user input.

femtomc:
The idea is that we can allow users to augment the behavior of edit programs by providing more information about the types of changes which are flowing through the generative function.

Implementing this is very simple in the compositional `edit` language (see the implementation).

This type (`DiffAnnotate`) can also be used for logging at trace time (to see what types of changes are available at a particular generative function).


src/genjax/_src/core/generative/requests.py
```diff
@@ -51,3 +53,24 @@ def edit(
...
+    def edit(
+        self,
+        key: PRNGKey,
+        tr: Trace[R],
+        argdiffs: Argdiffs,
+    ) -> tuple[Trace[R], Weight, Retdiff[R], "EditRequest"]:
+        new_argdiffs = self.argdiff_fn(argdiffs)
+        tr, w, retdiff, bwd_request = self.request.edit(key, tr, new_argdiffs)
+        new_retdiff = self.retdiff_fn(retdiff)
+        return tr, w, new_retdiff, bwd_request
```
femtomc, line 85: So simple -- by virtue of our compositional model for how `edit` is dispatched -- we can introspect on changes before and after application of `DiffDimap.request.edit`

This means, for instance -- we can decorate requests like `HMC` with assertions that have to hold true to support the correctness of the gradient computations which HMC relies upon (apply HMC to _subtraces_, as part of a compositional request like `StaticRequest`).

sritchie: Love it, this is a great, simple piece

src/genjax/_src/core/interpreters/forward.py
```diff
@@ -67,17 +67,6 @@ def _abstract(*flat_avals, **params):
...
-            primals_out, tangents_out = ad.jvp(
-                lu.wrap_init(self.impl, params)
-            ).call_wrapped(primals, tangents)
-            tangents_out = jax_util.safe_map(
-                ad.recast_to_float0, primals_out, tangents_out
-            )
-            return primals_out, tangents_out
-
-        ad.primitive_jvps[self] = _jvp
-
```
femtomc, line 80: Dead code which was tripping `pyright` because of the `recast_to_float0` issue for JAX>0.4.33

src/genjax/_src/core/generative/requests.py
sritchie: I would add the `map`, contramap` and `dimap` methods to `EditRequest`, like we did with the genfn combinators.

src/genjax/_src/core/generative/core.py
```diff
@@ -154,6 +155,29 @@ def edit(
...
+        post: Callable[[Retdiff[R]], Retdiff[R]],
+    ) -> "EditRequest":
+        return self.dimap(post=post)
+
+    def contramap(
+        self,
+        pre: Callable[[Argdiffs], Argdiffs],
+    ) -> "EditRequest":
+        return self.dimap(pre=pre)
+
```
femtomc, line 180: CC @sritchie here's the added methods. I don't yet think we need an "info" field (for clarity), but we could perhaps add one later -- one can't really change the behavior (e.g. ask for an entirely different edit move) of the underlying `EditRequest` with `DiffAnnotate`.

sritchie:
Boom, one last request - want to use the new methods in the test? that will test the methods AND the underlying wrapper that you're adding in the tests now. Nice work...

sritchie (approved):
Approving, just that one suggestion left.

femtomc:
I thought I did this.

Edit: here's link -- https://github.com/probcomp/genjax/blob/0e31ea9d6047326ee4475e5494aff3f2f67c9af9/tests/core/generative/test_requests.py#L110-L116

PR #1408, opened 2024-10-17
(5 / N) Add HMC as an EditRequest.

femtomc:
Add an implementation of HMC as an edit request. 

There's two types of `EditRequest` here -- the first is just `HMC`, copying the logic from Gen.jl's implementation of HMC. The second is `SafeHMC`, a `DiffAnnotate` wrapped version of `HMC` which can be used in _other hierarchical edit requests_ (like `StaticRequest`), but asserts a condition required to ensure that gradients for the "local" HMC update are computed correctly.

I've also added correctness checks to the test cases for `Regenerate` in this PR.


femtomc:
(CC @MathieuHuot)

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -322,12 +321,26 @@ def edit_regenerate(
...
+                new_trace = DistributionTrace(self, primals, chm.get_value(), new_score)
+                return (
+                    new_trace,
+                    new_score - trace.get_score(),
+                    Diff.no_change(trace.get_retval()),
+                    Update(
+                        ChoiceMap.empty(),
+                    ),
+                )
         else:
```
femtomc, line 344: Was missing this logic for `Regenerate`, which means when I added correctness checks, we had failures (e.g. it wasn't computing the correct weight).

tests/inference/test_requests.py
```diff
@@ -0,0 +1,273 @@
...
+        tr = simple_normal.simulate(sub_key, ())
+
+        # First, try y1 and test for correctness.
+        old_v = tr.get_choices()["y1"]
+        request = genjax.Regenerate(S["y1"])
+        new_tr, fwd_w, _, bwd_request = request.edit(key, tr, ())
+        old_density = genjax.normal.logpdf(old_v, 0.0, 1.0)
+        new_density = genjax.normal.logpdf(new_tr.get_choices()["y1"], 0.0, 1.0)
+        assert fwd_w != 0.0
+        assert fwd_w == new_density - old_density
```
femtomc, line 56: New correctness checks for `Regenerate`.

tests/inference/test_requests.py
```diff
@@ -0,0 +1,273 @@
...
+        new_target_density = genjax.normal.logpdf(
+            new_x, 0.0, 1.0
+        ) + genjax.normal.logpdf(new_y, new_x, 0.01)
+        assert fwd_w != 0.0
+        # The change in the target scores corresponds to the non-momenta terms in the HMC alpha computation.
+        assert (new_tr.get_score() - tr.get_score()) == pytest.approx(
+            new_target_density - old_target_density, 1e-6
+        )
+        # The weight factors in the target score change and the momenta, so removing the change in the target scores should leave us with a non-zero contribution from the momenta.
+        assert fwd_w - (new_tr.get_score() - tr.get_score()) != 0.0
```
femtomc, line 174: We don't know the exact momenta scores used in the alpha computation in HMC, but we know that the `fwd_w` _shouldn't_ be just the difference in scores between the targets.

tests/inference/test_requests.py
```diff
@@ -0,0 +1,273 @@
...
+
+        # Check for gradient convergence.
+        new_tr = tr
+        for _ in range(20):
+            key, sub_key = jrand.split(key)
+            new_tr, *_ = editor(sub_key, new_tr, ())
+        assert new_tr.get_choices()["x"] == pytest.approx(3.0, 5e-3)
```
femtomc, line 181: This is a gradient convergence check -- we're not doing accept/reject in the HMC edit request itself, so the traces should always move towards the optima via the gradients.

tests/inference/test_requests.py
```diff
@@ -0,0 +1,273 @@
...
+        # Compositional request _including_ HMC.
+        request = StaticRequest(
+            {
+                "x": SafeHMC(Selection.at["x"], jnp.array(1e-2)),
+                "y": StaticRequest({
+                    "x": Regenerate(Selection.all()),
+                    "y": Update(C.choice(3.0)),
+                }),
+            },
+        )
```
femtomc, line 360: This is a compositionality test case -- which shows how one can use `HMC` _safely_ in a nested edit request.

tests/inference/test_requests.py
```diff
@@ -0,0 +1,273 @@
+# Copyright 2024 MIT Probabilistic Computing Project
```
femtomc, line 1: This is the same as the old file `tests/core/generative/test_requests` -- just moved it over into `tests/inference`.

src/genjax/_src/inference/requests/hmc.py
```diff
@@ -0,0 +1,223 @@
...
+    Apply a Hamiltonian Monte Carlo (HMC) update that proposes new values for the selected addresses,
+    returning the new trace, and a weight which is equal to the alpha accept-reject ratio computation for HMC.
+
+    Hamilton's equations are numerically integrated using leapfrog integration with step size `eps` for `L` steps.
+    See equations (5.18)-(5.20) of Neal (2011).
+
+    # References
+    Neal, Radford M. (2011), "MCMC Using Hamiltonian Dynamics", Handbook of Markov Chain Monte Carlo,
+    pp. 113-162. URL: http://www.mcmchandbook.net/HandbookChapter5.pdf
+    """
```
femtomc, line 150: Here's the main logic to consider.

src/genjax/_src/core/generative/core.py
```diff
@@ -161,21 +162,21 @@ def dimap(
...
+    ) -> "genjax.DiffAnnotate[Self]":
         from genjax import DiffAnnotate
 
         return DiffAnnotate(self, argdiff_fn=pre, retdiff_fn=post)
 
     def map(
         self,
         post: Callable[[Retdiff[R]], Retdiff[R]],
-    ) -> "EditRequest":
+    ) -> "genjax.DiffAnnotate[Self]":
```
MathieuHuot, line 173: What's going on here?

src/genjax/_src/core/generative/requests.py
```diff
@@ -57,7 +59,7 @@ class Regenerate(PrimitiveEditRequest):
 
 # NOTE: can be used in an unsafe fashion!
 @Pytree.dataclass(match_args=True)
-class DiffAnnotate(EditRequest):
+class DiffAnnotate(Generic[ER], EditRequest):
```
MathieuHuot, line 62: what does this actually do/change?

src/genjax/_src/generative_functions/distributions/distribution.py
MathieuHuot: EDIT: the flag is not for masking but checking the selection. Ignore this comment.
Ah I see. We were implicitly relying on the `False` flag computing everything anyway, but here we weren't actually doing the computation. Can you add a comment basically saying that we need to compute the edit eagerly irrespective of the flag for the logic to be sound?

MathieuHuot: Also why does this live only in `Distribution`? Was it simpler than for generative functions that produce static choicemaps? Or is it automatically extended to them?

MathieuHuot (approved):
I'd like some clarity on the few questions I asked, but looks great overall!

src/genjax/_src/core/generative/requests.py
femtomc: this is just for `pyright` static typing, no semantic change

src/genjax/_src/core/generative/core.py
femtomc: Sam suggested I copy the dimap methods from generative function to do the same thing (but use `DiffAnnotate`) for `EditRequest`

src/genjax/_src/generative_functions/distributions/distribution.py
femtomc: Distributions holds a lot of the basic logic, then combinators (like the static generative function) implement their support compositionally -- relying on the implementation in e.g. distributions.

MathieuHuot: @femtomc ok thanks for clarifying re the first question. But does it mean we don't have HMC on static generative functions yet?

ahiser1117:
😎

PR #1411, opened 2024-10-19
(6 / N) Add `Rejuvenate` -- a proposal-driven regeneration `EditRequest`

femtomc:
This is an `EditRequest` whose logic follows custom proposal Metropolis-Hastings. The weight is the custom proposal Metropolis-Hastings accept reject ratio.


src/genjax/_src/core/generative/requests.py
```diff
@@ -49,14 +52,61 @@ def edit(
...
+    """
+    The `Wiggle` edit request is a compositional request which utilizes
+    a proposal generative function to propose a change to a trace.
+
+    Specifying a wiggle requires that a user provide a `proposal` generative function, and an `argument_mapping`, which is a callable that accepts the `ChoiceMap` from the previous trace and produces arguments to the invocation of the generative function.
+
+    `Wiggle` can be used for quick custom regeneration moves (e.g. a Gaussian at an address, can I propose a random walk around the previous value using another Gaussian?) as well as larger structured proposals.
+
+    It is isomorphic to the logic of Metropolis-Hastings with a custom proposal (which defines the MCMC kernel of the Metropolis-Hastings algorithm) _without the accept-reject step_. The ratio is returned as the SMCP3 weight of the move.
+    """
```
femtomc, line 91: The logic here is the same as: https://github.com/probcomp/Gen.jl/blob/a5fc8e39e056acc2edd583ca6bcaa9ba1a95bde4/src/inference/mh.jl#L39-L60

src/genjax/_src/generative_functions/static.py
```diff
@@ -442,9 +442,9 @@ def wrapper(
     return wrapper
 
 
-##########################
-# Choice map change edit #
-##########################
+###############
+# Update edit #
+###############
```
femtomc, line 448: Cleaning up some naming, no change to functionality.

tests/inference/test_requests.py
```diff
@@ -135,10 +136,69 @@ def linked_normal():
...
+        request = StaticRequest({
+            "y1": Wiggle(
+                genjax.normal,
+                lambda chm: (0.0, 1.0),
+            )
+        })
+        new_tr, w, _, _ = request.edit(sub_key, tr, ())
+        new_v = new_tr.get_choices()["y1"]
+        assert old_v != new_v
+        assert w == 0.0
```
femtomc, line 167: Here's a math correctness change (change in target should be exactly the same as the inverted proposal ratio -- therefore weight is 0).

tests/inference/test_requests.py
```diff
@@ -135,10 +136,69 @@ def linked_normal():
...
+
+        # Run Metropolis-Hastings for 100 steps.
+        for _ in range(100):
+            key, sub_key = jax.random.split(key)
+            new_tr, w, _, _ = request.edit(sub_key, tr, ())
+            key, sub_key = jax.random.split(key)
+            check = jnp.log(genjax.uniform.sample(sub_key, 0.0, 1.0)) < w
+            tr = jtu.tree_map(lambda v1, v2: jnp.where(check, v1, v2), new_tr, tr)
+
+        assert tr.get_choices()["y1"] == pytest.approx(3.0, 5e-3)
```
femtomc, line 194: Here's a convergence check for using `Wiggle` to drive a Metropolis-Hastings move.

src/genjax/_src/core/generative/requests.py
```diff
@@ -49,14 +52,61 @@ def edit(
         tr: Trace[R],
         argdiffs: Argdiffs,
     ) -> tuple[Trace[R], Weight, Retdiff[R], "EditRequest"]:
-        return tr, jnp.array(0.0), Diff.no_change(tr.get_retval()), EmptyRequest()
+        if Diff.static_check_no_change(argdiffs):
+            return tr, jnp.array(0.0), Diff.no_change(tr.get_retval()), EmptyRequest()
+        else:
```
MathieuHuot: why do we need the added logic?

src/genjax/_src/core/generative/requests.py
MathieuHuot: Can we make the docstring look more like the one in Gen.jl? 
Ideally we should define a norm for technical docstrings and be consistent. Something like:
- args
- one sentence explaining technically what it is. 
- conditions of applicability
- optionally, a running example using the thing

tests/inference/test_requests.py
```diff
@@ -135,10 +136,69 @@ def linked_normal():
...
+        tr, _ = linked_normal.importance(sub_key, C.kw(y2=3.0), ())
+
+        #######
+        # Do some data-driven behavior
+        #######
+
+        request = StaticRequest({
+            "y1": Wiggle(
+                genjax.normal,
+                lambda chm: (chm.get_value(), 0.3),
```
MathieuHuot, line 182: is `lambda chm: (chm.get_value(), 0.3)` the argument_mapping? what does it do?

MathieuHuot: Am I correct to understand this is using the proposal `normal(3.0, 0.3)` here?

MathieuHuot (changes requested):
Looks good overall. Apart from my comments above the main thing to change is the name `Wiggle`. What about calling it `Regenerate`? 
Can we unify so that Regenerate with no proposal uses the default internal one (called Regenerate in Gen.jl) and otherwise this one?

femtomc:
> Looks good overall. Apart from my comments above the main thing to change is the name Wiggle. What about calling it Regenerate?
Can we unify so that Regenerate with no proposal uses the default internal one (called Regenerate in Gen.jl) and otherwise this one?

How about just `Rejuvenate`?

src/genjax/_src/core/generative/requests.py
femtomc: If the arguments have changed, the fallback behavior of an `EmptyRequest()` should be that we need to re-evaluate the density (which we can access via `Update(...)`)

tests/inference/test_requests.py
femtomc: That's correct, it is the argument mapping -- it's designed to support a limited form of "introspect on local choices" functionality for the proposal (the user gets to provide a generative function as the proposal, and the argument mapping allows the generative function to introspect on the choice _at the subtrace_ which the proposal is proposing to change).

The proposal is not `normal(3.0, 0.3)` -- it's `normal(prev_value, 0.3)`.

If one wanted to do a data-driven proposal using this request, they would have to provide _a larger generative function_ (encapsulating a bigger part of the address hierarchy) to introspect on the data address.

src/genjax/_src/core/generative/requests.py
MathieuHuot: I don't understand. What is the spec for an `EmptyRequest` and when does it show up? Why do we need something else than `update`?

tests/inference/test_requests.py
MathieuHuot: ok got it! Could you just remove the "# Do some data-driven behavior" comment then as you're not doing anything data-driven here?

MathieuHuot:
> > Looks good overall. Apart from my comments above the main thing to change is the name Wiggle. What about calling it Regenerate?
> > Can we unify so that Regenerate with no proposal uses the default internal one (called Regenerate in Gen.jl) and otherwise this one?
> 
> How about just `Rejuvenate`?

sounds good!

src/genjax/_src/core/generative/requests.py
MathieuHuot: @femtomc can you add that this `Rejuvenate` uses the same proposal `q` for the K and for the L kernels when seen as an SMCP3 move?

src/genjax/_src/core/generative/requests.py
femtomc: Without thinking, the viewpoint on `EmptyRequest()` is that it shouldn't change anything -- but strictly speaking, that's only valid _if the arguments haven't changed_ (otherwise, you can't return a weight of 0.0). We can choose to either throw and error, or ... just do the right thing, by using `Update(ChoiceMap.empty())`

There's some parts of the implementation which use `EmptyRequest()` instead of `Update(ChoiceMap.empty())` -- I think these types should behave identically, but the latter is more verbose.

We have plenty of requests which use the logic from other requests, that's nothing new.

src/genjax/_src/core/generative/requests.py
femtomc: I agree with the docstring suggestion, but not sure I have time to make it really nice right now.

This one:
> 
    args
    one sentence explaining technically what it is.
    conditions of applicability
    optionally, a running example using the thing

MathieuHuot: ok good for this PR. We'll revisit docstring another time.

MathieuHuot (approved):

PR #1418, opened 2024-10-26
(StaticGenerativeFunction) Prevents allocation and accumulation of score, when its available in callee subtraces.

femtomc:


src/genjax/_src/generative_functions/static.py
```diff
@@ -319,7 +314,7 @@ def wrapper(key, args):
 @dataclass
 class AssessHandler(StaticHandler):
     choice_map_sample: ChoiceMap
-    score: FloatArray = Pytree.field(default_factory=lambda: jnp.zeros(()))
+    score: Score = Pytree.field(default_factory=lambda: jnp.zeros(()))
```
MathieuHuot, line 317: I don't know if I like this way of handling information with classes such as  Score or Weight. I think we should keep `FloatArray` and whenever something non obvious is going on, add comments/examples/tests and a written doc. These shallow types are not actually helping with anything.

MathieuHuot (approved):

MathieuHuot:
@littleredcomputer I think this is the first step going towards the "lazy computation of the score that doesn't store the same thing 30 times at different places in the trace" that you were hoping for. Could you give a look at this PR too and maybe start reflecting on how to expand on this simplification to other places? Or share some of your reflections on the matter if you think they're in shareable shape. I'd love to really nail the choicemap datastructure in terms of efficiency, simplicity and formal guarantees this quarter in the best way we can.

src/genjax/_src/generative_functions/static.py
femtomc: This is already standard throughout the codebase — this line was left over and missed by previous PRs.

femtomc: To discuss whether this standard is a good idea — can you raise your concern in an issue so we can talk about it?

I have a few arguments for why I think it is good — but I don’t want to put them in this PR (where they might be lost).

MathieuHuot: absolutely, sounds good.

PR #1421, opened 2024-10-28
Remove ellipsis support from queries, add `SwitchChm` (GEN-698, GEN-696, GEN-542)

sritchie:
This PR:

- renames `Mask.maybe` to `Mask.build` and `Mask.maybe_none` to `Mask.maybe_mask`, better docstrings throughout
- adds an `__or__` and `__xor__` implementation to `Mask`; this is now used by the `Or` and `Xor` choicemap classes instead of their `get_value` implementations
- adds a `Switch` subclass of `ChoiceMap`, which is now returned by `ChoiceMap.switch`
- adds a `flatten` method to `Mask` that drops the value to `None` or `.value` in the case of concrete flags
- removes `genjax._src.core.typing.Value`; this is just `Any`, let's just say it.
- replaces all `...` lookups with `:`, since that's what we are actually going for
- adds a constructor optimization to `Choice`; around a concrete `False` mask the choice simplifies to `None`, and around a concrete `True` mask the mask is stripped off.
- removes support for ellipses in `__getitem__`, `__call__` and `__contains__` for `Selection` and `ChoiceMap`.

(The reason for this is that `ChoiceMap.__getitem__` with a wildcard forces us to `xor` together the values for all possible keys. Users can now use `:` in place of `...` for `Indexed` layers, and we don't need the confusing wildcard lookup for a static level.)

## Next steps:

I am moving toward deleting `Or` and `Xor`, but `Filtered` is currently in the way. The next step is to see if we can push filters down to the bottom, and delete the `Filtered` class. See #1422


src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1617,15 +1605,12 @@ def merge_with(
     def get_value(self) -> Any:
         return None
 
-    def get_submap(self, addr: ExtendedAddressComponent) -> ChoiceMap:
-        def check(k: ExtendedAddressComponent) -> Flag:
-            return True if addr is Ellipsis else k == addr
-
-        acc = ChoiceMap.empty()
```
sritchie, line 1624: woohoo, get rid of this!!

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1682,34 +1707,20 @@ def build(
...
-        idx = pair_flag_to_idx(check1, check2)
-
-        if isinstance(idx, int):
-            # This branch means that both has_value() checks have returned concrete bools, so we can
-            # make the choice directly.
-            return [v1, v2][idx]
-        else:
-            return tree_choose(idx, [v1, v2])
+        match self.c1.get_value(), self.c2.get_value():
+            case None, b:
```
sritchie, line 1716: this removes the need for the `int` branch and makes the logic less confusing.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1756,28 +1767,21 @@ def build(
...
-            return first + 2 * FlagOp.and_(FlagOp.not_(first), second) - 1
-
-        idx = pair_flag_to_idx(check1, check2)
-        if isinstance(idx, int):
-            # This branch means that both has_value() checks have returned concrete bools, so we can
-            # make the choice directly.
-            return [v1, v2][idx]
-        else:
-            return tree_choose(idx, [v1, v2])
+        match self.c1.get_value(), self.c2.get_value():
```
sritchie, line 1776: this removes the need for the `int` branch and makes the logic less confusing.

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -120,17 +158,131 @@ def _check():
...
-            case _:
-                return self.flag
+            case flag:
+                return flag
+
+    ###############
+    # Combinators #
+    ###############
+
+    def _validate_mask_shapes(self, other: "Mask[R]"):
```
sritchie, line 193: wdyt of this validation, @MathieuHuot ? This is a first cut, and of course we need some tests here.

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -120,17 +158,131 @@ def _check():
...
+            1 -> select second value
+
+        Args:
+            first: The flag for the first value
+            second: The flag for the second value
+
+        Returns:
+            An index (-1, 0, or 1) indicating which value to select
+        """
+        return first + 2 * FlagOp.and_(FlagOp.not_(first), second) - 1
```
sritchie, line 234: I can re-use this for `or` and `xor`; in the `xor` case, we generate a new flag using `FlagOp.xor_`, so we don't care what value's chosen in the both-false or both-true case (as it'll be masked)

src/genjax/_src/core/generative/functional_types.py
MathieuHuot: is this raising an error for the case we looked at with the same traced value in 2 branches of a switch having different shapes and when we decided there was just no correct broadcasting in this situation?

tests/core/test_choice_maps.py
```diff
@@ -232,20 +229,6 @@ def test_selection_contains(self):
         assert not nested_sel("a")("b").check()
         assert nested_sel("a")("b")("c").check()
 
-    def test_selection_ellipsis(self):
```
MathieuHuot, line 235: why is all of this removed with no new tests?

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -120,17 +158,130 @@ def _check():
...
+                return Mask.build(self, False)
+            case True, False:
+                return other
+            case False, True:
+                return self
+            case self_flag, other_flag:
+                idx = self._pair_flag_to_idx(self_flag, other_flag)
+
+                chosen = tree_choose(idx, [self.value, other.value])
+                return Mask.build(chosen, FlagOp.xor_(self_flag, other_flag))
```
MathieuHuot, line 265: something's a bit strange: why do you return `Mask.build(chosen, FlagOp.xor_(self_flag, other_flag))` in one case and simply `return tree_choose(idx, [self, other])` in the other?
(one case is Or, the other Xor)

src/genjax/_src/core/generative/functional_types.py
sritchie: yes, that's the idea. (I need tests for this of course...)

tests/core/test_choice_maps.py
sritchie: because the functionality is gone. I can add a couple of these back and show that it raises a type error to try and pass an ellipsis in

src/genjax/_src/core/generative/functional_types.py
sritchie: in the `or` case,  `_pair_flag_to_idx` handles the `(True, True)`, `(True, False)`, `(False, True)` correctly by picking the proper side for each entry (which will have a `True` flag in the correct spot). For the `(False, False)` case, it doesn't matter which side we pick since the flag ends up `False` for that entry.

The XOR case is similar (duh) but in the `(True, True)` case neither side will give us the flag we want... so instead of doing anything special there I just rebuild the flag using `FlagOp.xor_`.

tests/core/test_choice_maps.py
MathieuHuot: yeah it would be good. Also perhaps to help people port their code when they have to change `...` into `:` when using a new release if they get a specific useful error message?

src/genjax/_src/core/generative/functional_types.py
MathieuHuot: ok could you add a comment about that last thing in the Xor case?

MathieuHuot (approved):

PR #1422, opened 2024-10-28
remove `Filtered`, fix `MaskTrace` vmappability (GEN-662, GEN-569, GEN-775, GEN-767)

sritchie:
This PR:

- makes `filter` an `@abstractmethod` of `ChoiceMap`, forcing all choicemap implementations to declare how they're filtered
- removes `Filtered` from `choice_map.py`, and distributes the `_pushdown_filters` clauses into the `filter` methods of each choicemap subclass
- removes the `eager` option from `filter`, as eager is now the only behavior
- deprecates `ChoiceMap.simplify`, as it's now a no-op
- modifies `MaskTrace` to compute its choicemap, retval and score in `MaskTrace.build`, rather than in `get_choices()` and `get_retval()`. This fixes the problems we were hitting with `inner.get_choices()` and `score` growing in dimension via vmap before being combined. By combining them in the constructor, any outer `vmap` call will grow the already-masked retval, score and choicemap appropriately
- removes the `Filtered` check from `Distribution.edit_update_with_constraint`, allowing this method to work with any choicemap
- adds a `__getitem__` implementation to `Mask` that stops selecting from `flag` once the selection hits a scalar endpoint, but continues down into `value`.


sritchie:
Oh, @MathieuHuot , the second bug is coming from the missing `.T` thing that we haven't merged yet. Looks like it was lurking, and appears if you try and push the filters down...

MathieuHuot:
> Oh, @MathieuHuot , the second bug is coming from the missing `.T` thing that we haven't merged yet. Looks like it was lurking, and appears if you try and push the filters down...

So after an unrelated discussion with McCoy I think the .T is not the answer here. 
The mask-value shape mismatch can come from 2 sources:
- no vmap lurking around: in that case we may want to 
    - do the transpose trick I was talking about 
    - or disallow this weird broadcast because it's not clear when that should show up
    - disallow any broadcasting that doesn't come from vmap, so we don't have to think hard about transpose vs non-transpose as we don't have good examples anyway
- some vmap has been applied: in this case the solution comes from how JAX handles vmap in the first place. In this failing example it's occuring in a masked-scan combinator, i.e. a place where we know we introduced a leading axis due to vmap. So we can take out the leading axis from (10) and (10,2), we obtain () and (2) which are broadcast-compatible. We can thus apply the usual logic of multiplying the flag by the value there, and wrap the thing with a vmap to recover a good match when adding back the leading 10 axis. Maybe the TLDR is that when we vmap a generative function, we vmap all of its methods. So if we do it manually add a vmap inside a method of a combinator, it's not so surprising we'd need to add other manual vmap to some other parts to preserve the JAX philosophy/typing system.  Anyway sorry for the speech and let me know if that clarifies it or not.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1898,14 +1818,8 @@ def loop(inner: ChoiceMap, selection: Selection) -> Selection:
             case Indexed(c, addr):
                 return loop(c, selection(_full_slice)).extend(...)
 
-            case Choice(v):
-                if isinstance(v, Mask) and FlagOp.concrete_false(v.primal_flag()):
```
sritchie, line 1902: this case is stripped out by `Choice.build`, so we'll never hit this code path.

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -177,10 +180,7 @@ def maybe_mask(v: "R | Mask[R] | None", f: Flag) -> "R | Mask[R] | None":
             - None if `f` is concretely False.
             - A new Mask instance with the given value and flag if `f` is not concrete.
         """
-        if v is None:
```
sritchie: I don't think `maybe_mask` should mess with the values at all. this should only respond to concrete flags being true or false; I want the user to be able to keep the correct type if they wanted to mask a none.

sritchie:
Some notes for myself on my confusion around masking / `Indexed`... cc @MathieuHuot 
- a `Mask` holds a scalar flag
- if I vmap over a `Mask` I get a vector flag
- if I `Mask` over a vmapped thing, I still have a scalar, and `jax.vmap` will complain at trying to vmap through this layer

More strangeness: 
- when I query an `Indexed` layer, I create a scalar-flagged `Mask` at the bottom
- if I query again with a slice... how do I interact with the `Mask`? do I slice into the value and ignore the scalar flag? what if the flag is not scalar? then is the slice targeted at the vmapped layer?

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -922,6 +922,41 @@ class ChoiceMap(Pytree):
     # Map-like interfaces #
     #######################
 
+    @abstractmethod
```
sritchie, line 925: I made this an abstractmethod so that each ChoiceMap type is forced to declare its own clause for how to filter itself.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1159,47 +1194,6 @@ def switch(idx: int | IntArray, chms: Iterable["ChoiceMap"]) -> "ChoiceMap":
...
-            filtered = chm.filter(selection)
-            assert "y" not in filtered
-
-            # Using eager filtering
-            eager_filtered = chm.filter(selection, eager=True)
-            assert "y" not in eager_filtered
-            ```
-        """
-        ret = Filtered.build(self, selection)
-        return ret if not eager else _pushdown_filters(ret)
```
sritchie, line 1201: we now push down filters automatically.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1464,6 +1452,14 @@ def build(v: T) -> ChoiceMap:
         else:
             return Choice(v)
 
+    def filter(self, selection: Selection) -> ChoiceMap:
```
sritchie, line 1455: each of these "filter" methods was a clause in `_pushdown_filters`.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1542,7 +1542,7 @@ def get_submap(self, addr: AddressComponent) -> ChoiceMap:
                 # If `check` contains a match (we know it will be a single match, since we constrain addr to be scalar), then `idx` is the index of the match in `self.addr`.
                 # Else, idx == 0 (selecting "junk data" of the right shape at the leaf) and check_array[idx] == False (masking the junk data).
                 idx = jnp.argwhere(check, size=1, fill_value=0)[0, 0]
-                return jtu.tree_map(lambda v: v[idx], self.c.mask(check))
+                return jtu.tree_map(lambda v: Mask.build(v[idx], check[idx]), self.c)
```
sritchie: this was broken before. Imagine that `check` is an array of shape (3,), but `v` has leading dimension `5`.

Our new stance is "a `Mask` has a scalar flag; if the flag is vectorized its leading dimensions better match the value / it better have been created with vmap"`. 

To match this we have to tree_map down and create a `Mask` using `check[idx]`, which is a scalar.

MathieuHuot (approved):

PR #1423, opened 2024-10-29
Enforce "Mask has scalar flag", allow vectorized `Mask` (GEN-768, GEN-765, GEN-769)

sritchie:
This PR:

- adds a `Mask.__init__` that checks that 
  - a `Mask` is not instantiated wrapping another `Mask`
  - the `Mask` has either a scalar flag, or if the flag has a shape, its shape is a prefix of all leaves of the value. This check ensures that we can treat the `Mask` as a vectorized structure and `vmap` down to the point where we have a scalar `Flag` again.
- modifies `maybe_mask` to preserve masked `None`s. This method should only be looking at the flag!
- adds an `__invert__` implementation to `Mask`, allowing `~mask`
- when calling `|` or `^` between two masks, validates that the shapes of the values and flags match. This ensures that we aren't combining `Mask`s at different vectorization levels
- when called with a `Mask` value, `Mask.build` has to either take a scalar flag OR a flag that matches the already-vectorized shape of the `Mask` value
-  checks that `or` and `xor` work between vectorized values
- modifies all of the `FlagOp` operations so that an `Array` on EITHER side coerces the result into an `Array`. We need this to maintain our vectorization levels, so that a scalar `bool` flag can only live in a non-vectorized `Mask`
- adds type annotations helping the compiler understand this latest check


tests/generative_functions/test_mask_combinator.py
```diff
@@ -161,19 +161,13 @@ def step(x):
         key = jax.random.key(0)
         mask_steps = jnp.arange(10) < 5
         model = masked_scan_combinator(step, n=len(mask_steps))
-        init_particle = model.simulate(key, ((0.0,), mask_steps))
```
sritchie, line 164: this was a lurking bug. the `(0.0,)` was getting passed into `genjax.normal` and causing singleton array draws.

notebooks/cookbook/expressivity/masking.ipynb
```diff
@@ -201,10 +206,10 @@
    "outputs": [],
    "source": [
     "stop_at_index = 5\n",
-    "mask = Mask(None, jnp.arange(state_size) < stop_at_index)\n",
+    "pairs = (None, jnp.arange(state_size) < stop_at_index)\n",
```
sritchie, line 209: I had to change this because my validation prevents you from such a `Mask`, where the flags have a higher dimension than the value. In this case I think it's more natural to use a tuple as the pytree for scanning, vs a `Mask`.

notebooks/cookbook/expressivity/masking.ipynb
```diff
@@ -163,25 +162,31 @@
...
-    "            Mask(masked_retval.value[0], masked_retval.flag),\n",
-    "            Mask(masked_retval.value[1], masked_retval.flag),\n",
-    "        ),\n",
-    "    )\n",
+    "            next_inval,\n",
+    "        )\n",
+    "\n",
+    "    def post(_, masked_retval):\n",
+    "        carry, acc_entry = masked_retval.value\n",
+    "        if acc_entry is not None:\n",
```
sritchie, line 176: I know we're not really using this... but if we DO, I think it's better to say that if the user's function returns nothing, then our scan should not accumulate anything.

notebooks/cookbook/expressivity/masking.ipynb
```diff
@@ -163,25 +162,31 @@
...
+    "        return (Mask(carry, masked_retval.flag), acc_entry)\n",
+    "\n",
+    "    mstep = step.mask().dimap(pre=pre, post=post)\n",
     "    scanned = mstep.scan(**scan_kwargs)\n",
     "\n",
     "    scanned_nice = scanned.dimap(\n",
     "        pre=lambda initial_state, masked_input_values: (\n",
     "            Mask(initial_state, True),\n",
-    "            Mask(masked_input_values.value, masked_input_values.flag),\n",
+    "            masked_input_values,\n",
```
sritchie, line 187: We don't need to mess with this pair at all.

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -68,7 +69,51 @@ class Mask(Generic[R], Pytree):
...
+        Mask._validate(v, f)
+        self.value, self.flag = v, f  # pyright: ignore[reportAttributeAccessIssue]
+
+    @staticmethod
+    def _validate(value: R, flag: Flag | Diff[Flag]) -> None:
+        """
+        Validates that the flag can be broadcast to match the shape of the value.
+
+        This method performs two key validations:
+        1. If the value is a pytree, ensures all leaves have the same shape
```
MathieuHuot: Surely I can create a choicemap with leaves oh shape (3,2) and (4,5), e.g. with 2 multivariate normal or something, and then I should be able to mask it.
So this seems unnecessarily restrictive

sritchie: you're right, removing this restriction based on our conversations.

src/genjax/_src/core/generative/functional_types.py
```diff
@@ -68,7 +67,80 @@ class Mask(Generic[R], Pytree):
...
+        Args:
+            this: First value to compare
+            other: Second value to compare
+
+        Raises:
+            ValueError: If any leaf shapes don't match exactly
+        """
+
+        # Check array shapes match exactly (no broadcasting)
+        def check_leaf_shapes(x, y):
```
MathieuHuot, line 125: @sritchie  Let's say I create a model using a `switch` combinator, where one branch has traced "x" and the other traced "y". 
By the time we reach this function, will the traced choices already be both "x" and "y" for both branches?

sritchie: In that case, if you access `"x"` you will be trying to combine a `Mask` and a `None`, and if you reach `"y"` you'll be trying to combine a `None` and a `Mask`.

You'll only hit the shape restriction for addresses that clash, like switching two models with an `"x"`... we should add tests for this case, good call.

MathieuHuot (approved):
Please add the test and then all good!

PR #1427, opened 2024-11-04
Add MALA as an edit request.

femtomc:


MathieuHuot:
@femtomc given that we got HMC in the codebase, do we want to cancel this one?

femtomc:
Closing this, as we have HMC -- and we can re-visit if required.

PR #1428, opened 2024-11-04
Add convenience regenerate method to traces and generative functions.

femtomc:
Also re-org the Update request class (put it in the requests file).


sritchie (approved):
Throw a quick test in to get coverage on that flagged line then let's goooo

femtomc:
Closed to be handled in a new PR.

PR #1429, opened 2024-11-04
dimap's post-xform should accept the original args (GEN-781)

sritchie:
This PR modifies `DimapCombinator` to pass args to `retval_mapping` without the `argument_mapping` applied.

### DIFF
--- a/src/genjax/_src/core/generative/generative_function.py
+++ b/src/genjax/_src/core/generative/generative_function.py
@@ -45,7 +45,6 @@ if TYPE_CHECKING:
     import genjax
 
 _C = TypeVar("_C", bound=Callable[..., Any])
-ArgTuple = TypeVar("ArgTuple", bound=tuple[Any, ...])
 
 # Generative Function type variables
 R = TypeVar("R")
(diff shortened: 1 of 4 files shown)


MathieuHuot:
@sritchie I know there's been a discussion with George somewhere around this, but as there's no example I find it hard to understand why we want that. It may make sense to allow the post-xform to accept the original args, but shouldn't it also accept the returned args? Wasn't clear to me whether it could still do that.

sritchie:
Yes, the change is whether it should accept a pair of (original args, return value) or (transformed args, retval). The current behavior is the latter and the PR changes it to the former

MathieuHuot (approved):

georgematheos:
Looks great!

Quick question -- is it better to have the post function accept (original args, return value), or to have it accept a triple (original args, transformed args, return value)?  (In the latter case users could always reconstruct the transformed args; the version with an arg triple would just enable the prevention of recomputation.)

sritchie:
Good idea, I think we should do it. See my PR...

PR #1433, opened 2024-11-15
Fix vmap over Pytree (GEN-853)

sritchie:
This PR fixes a bug in the vmap combinator that occurred if an integer specified somewhere in `in_axes` lined up with a pytree in the arguments.


georgematheos (approved):
Sweet!

MathieuHuot (approved):

PR #1434, opened 2024-11-15
Add a "static" version of `Const` (GEN-852)

sritchie:
This PR allows folks to pass non-consts into `Const.unwrap` (in addition to the existing `const_instance.unwrap()`


georgematheos (approved):
Love it!

PR #1436, opened 2024-11-18
Remove Xor, chip away at Or (GEN-774)

sritchie:
This PR:

- removes `Xor` in favor of only using `Or` (my next PR will push `Or` down into `Choice`)
- modifies `from_mapping` to use `|=` instead of `^=`
- bans `|` between a `Choice` and a non-`Choice`. This should never occur EXCEPT in `Switch` and we handle that separately with a `SwitchChm`
- `|` between two choices pushes the `|` call down into the values
- throws a `TypeError` if a marginal is supplied to a `Target`; this is not supported!
- removes the `ravi_stack.ipynb` notebook from the docs site, as this was hitting that error

### DIFF
--- a/src/genjax/_src/core/generative/choice_map.py
+++ b/src/genjax/_src/core/generative/choice_map.py
@@ -1101,13 +1101,13 @@ class ChoiceMap(Pytree):
             ```
 
         Note:
-            If multiple pairs have the same address, the resulting ChoiceMap will error on lookup, as duplicate addresses are not allowed due to the `^` call internally.
+            If multiple pairs have the same address, later pairs will overwrite earlier ones.
         """
         acc = ChoiceMap.empty()
 
         for addr, v in pairs:
             addr = addr if isinstance(addr, tuple) else (addr,)
-            acc ^= ChoiceMap.entry(v, *addr)
+            acc |= ChoiceMap.entry(v, *addr)
 
         return acc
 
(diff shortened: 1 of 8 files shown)


src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1454,11 +1457,8 @@ def build(v: T) -> ChoiceMap:
 
     def filter(self, selection: Selection) -> ChoiceMap:
         sel_check = selection.check()
-        masked = Mask.maybe_mask(self.v, sel_check)
-        if masked is None:
-            return ChoiceMap.empty()
-        else:
-            return ChoiceMap.choice(masked)
+        masked = Mask.build(self.v, sel_check)
```
sritchie, line 1460: the simplification cases are already handled by `Choice.build`

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1788,22 +1725,23 @@ def build(
...
-        match self.c1.get_value(), self.c2.get_value():
-            case None, b:
-                return b
-            case a, None:
-                return a
-            case a, b:
-                a = Mask.build(a)
-                b = Mask.build(b)
-                return (a | b).flatten()
+        return None
```
sritchie, line 1744: it's no longer possible to build a value-producing `Or` instance.

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1840,7 +1778,7 @@ def loop(inner: ChoiceMap, selection: Selection) -> Selection:
                 return acc
 
             case _:
-                return selection
+                raise ValueError(f"Unknown ChoiceMap type: {type(inner)}")
```
sritchie, line 1781: I want to fail here, we should never hit this case

MathieuHuot (approved):
Looks good!

PR #1437, opened 2024-11-18
Remove Indexed layers with None address

sritchie:
This PR:

- stops the `vmap` and `scan` combinators from wrapping their generated choicemaps with an `Indexed` layer with a `None` address
- modifies `get_submap` of `Choice` and `Static` to push any dynamic addresses that they are passed down to `__getitem__` at the leaves.

Before this PR, `C[:, :, "x"].set(2darray)` would look like

```python
Indexed(addr=None, Indexed(addr=None, Static({"x": Choice(2darray)})))
```

when queried like `chm[:, :, "x"]`, each `:` would drop an outer `Indexed`.

As of this PR, the data structure will look like

```python
Static({"x": Choice(2darray)})
```

and the `Static` layer will handle both `:` by dispatching the `__getitem__` call down to the leaves (to `2darray`). It will only strip itself off when it encounters a static address.

In the next PR I'll implement the transformation of `chm[:, :, "x"]` to `chm["x"][:, :]`. After that, `ChoiceMap` will never have to see any dynamic addresses in `get_submap` at all.

### DIFF
--- a/src/genjax/_src/core/generative/choice_map.py
+++ b/src/genjax/_src/core/generative/choice_map.py
@@ -1101,13 +1101,13 @@ class ChoiceMap(Pytree):
             ```
 
         Note:
-            If multiple pairs have the same address, the resulting ChoiceMap will error on lookup, as duplicate addresses are not allowed due to the `^` call internally.
+            If multiple pairs have the same address, later pairs will overwrite earlier ones.
         """
         acc = ChoiceMap.empty()
 
         for addr, v in pairs:
             addr = addr if isinstance(addr, tuple) else (addr,)
-            acc ^= ChoiceMap.entry(v, *addr)
+            acc |= ChoiceMap.entry(v, *addr)
 
         return acc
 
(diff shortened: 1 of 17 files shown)


src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1464,7 +1464,12 @@ def get_value(self) -> T:
         return self.v
 
     def get_submap(self, addr: AddressComponent) -> ChoiceMap:
-        return ChoiceMap.empty()
+        if isinstance(addr, StaticAddressComponent):
+            return ChoiceMap.empty()
+        else:
```
sritchie, line 1469: this is a placeholder, before the final push of removing `Indexed` completely. This branch pushes any dynamic address seen by a `Choice` down to the leaves (exactly what happened for an `Indexed(..., None)` anyway!)

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1623,7 +1628,9 @@ def get_submap(self, addr: AddressComponent) -> ChoiceMap:
             v = self.mapping.get(addr, {})
             return Static(v) if isinstance(v, dict) else v
         else:
-            return ChoiceMap.empty()
+            return jtu.tree_map(
```
sritchie, line 1631: same push-down-to-the-leaves logic here in `Static` as we have above in `Choice`.

MathieuHuot (approved):

PR #1447, opened 2024-11-26
update adev_example with latest code from adev_demo

mhuebert:


sritchie (approved):
approve, just one ruff error, looks like!

mhuebert:
@sritchie try as I might, I cannot get this ruff check to pass. I first formatted the imports using vscode, then ran ruff --fix locally and committed the changes. Are there special settings being applied somehow?

sritchie:
I bet the pre commit hook and the GitHub action are out of sync, I will try and fix and merge when it works

mhuebert:
@sritchie thanks!

PR #1450, opened 2024-11-29
masked_iterate, masked_iterate_final (GEN-327, GEN-859)

sritchie:
This PR:

- adds `genjax.masked_iterate` and `genjax.masked_iterate_final`
- adds `masked_iterate` and `masked_iterate_final` combinator methods to `GenerativeFunction`
- adds new docs
- modifies the masking notebook to use this combinator, instead of the previous inlined version
- also fixes a bug where `genjax.flip` returned an int instead of a boolean


MathieuHuot (approved):
Looks great, excited to see more combinators :).

PR #1461, opened 2024-12-15
Fix non-Diff return value issue (GEN-902)

sritchie:
@georgematheos , this should do it, but I need @femtomc to tell me if there is a better way + get your test case in. Feel free to cut a release from this branch to get unblocked!!


georgematheos:
Thank you @sritchie!

sritchie:
@femtomc noted that this is the right place to solve the problem, so I'm taking that as a +1...

PR #1468, opened 2024-12-19
Fix Switch choicemap `|` behavior with static chms (GEN-908)

sritchie:
This PR modifies `Or` to push `Static` and `Choice` into the choicemaps entry of a `Switch`. The guards we have against `Or` between `Static` and `Choice` still apply internally.


arijit-dasgupta (approved):
Works on my side! I can see that the non switch choicemaps merges nicely into the inner choicemaps

PR #1477, opened 2025-01-29
Remove EmptyConstraint and MaskedConstraint as unnecessary concepts

femtomc:
Spring cleaning.


femtomc:
@sritchie Can you look at the CI failures on this? I don't exactly know what's causing it or what they mean, seems to have to do with "exporting genstudio"

sritchie (approved):
lgtm, I moved the target to `sritchie/remove_auth`, that should fix the poetry issue and get the build passing.

PR #1481, opened 2025-01-29
(INF 1/N) Basic interface functionality to support enumeration (exposed to users via `edit`).

femtomc:


femtomc:
Closing this for now in favor of going deeper on delayed sampling, which will present as another PR.

PR #1482, opened 2025-01-30
(DOCS 1/N) Re-organize the documentation, and start filling in holes.

femtomc:


sritchie:
@femtomc wdyt of us using a mkdocs style where we just splat out all docs in the codebase, like this? https://chisym.github.io/genlm-backend/reference/genlm_backend/llm/__init__/

femtomc:
Can we tuple tomorrow so I can give a clear picture of some of my goals for organizing the documentation?

Then we can triage, figure out how to get mkdocs to work for us.

femtomc:
I want to organize a bit so it’s not so overwhelming, and users have a clear conceptual map of the different conceptual pieces of GenJAX, and how fit they into a workflow.

we won’t solve this immediately obviously, but I think there’s a few low hanging fruits.

previously we kind of took the “exhaustive document the API” approach, which can be useful for sure! But I think if we curate slightly it’ll be way less imposing

femtomc:
This will also involve _omitting_ certain technical details surrounding the core data types in favor of explanations based more on intuition (reserving math to be presented in a different place for advanced users who want to dig into the guts)

femtomc:
Closing, will start again in a new branch.

PR #1489, opened 2025-02-04
gen-946 deprecate get_sample

littleredcomputer:
`get_sample` changed to `get_choices` throughout. Deprecation warning left on `Trace.get_sample`, the last remaining use of this name. (Interesting note: `@deprecate` doesn't appear to work on functions with return type annotation. That was puzzling).


sritchie:
@littleredcomputer lots of conflicts, otherwise lgtm!

sritchie (approved):
Might be worth merging by checking out everything from main (all the moved notebooks) then just find-replacing again...

PR #1491, opened 2025-02-04
Allow kwarg construction of masks (GEN-944)

sritchie:
This PR allows `Mask`s to be built via kwargs.

### DIFF
--- a/src/genjax/_src/core/generative/choice_map.py
+++ b/src/genjax/_src/core/generative/choice_map.py
@@ -1478,6 +1478,12 @@ class Choice(Generic[T], ChoiceMap):
             )
 
 
+class DocstringExample:
+    def __init__(self, docstring: str):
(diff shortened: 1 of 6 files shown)


src/genjax/_src/core/generative/choice_map.py
```diff
@@ -1478,6 +1478,12 @@ def get_submap(self, addr: AddressComponent) -> ChoiceMap:
             )
 
 
+class DocstringExample:
```
littleredcomputer, line 1481: do you still want this?

littleredcomputer (approved):

src/genjax/_src/core/generative/choice_map.py
sritchie: haha whoops no, good catch!

PR #1500, opened 2025-02-08
GEN-957 internal `StaticTrace` improvements

littleredcomputer:
- AddressVisitor is eliminated. Since 3.7, Python guarantees iteration order == insertion order in dictionaries. This means having a list of visited addresses together with a vector of traces at those addresses is not necessary: a dictionary can serve this purpose. 
- Adds a VoidTrace object to stash in such dictionaries to defend against duplicate traces
- These new subtrace dictionaries store subtrace names as bare strings when that is the type of the trace address (i.e., almost always). This eliminates the annoying `get_subtrace((v,))` pattern. This usage is still supported, but now provokes a deprecation warning. 
- This means there is no further confusion about what a tuple in a trace list means: the address is the address, so if you trace to a tuple address, the tuple enters the dictionary. This allows the collapse of several instances of code that shifted between string and 1-tuple representation addresses.
- In consequence of all this, the type definition of StaticAddress is now `StaticAddress = StaticAddressComponent | tuple[StaticAddressComponent, ...]`, which is a big change, but I think reflective of how addresses are actually used in GenJAX.


sritchie:
More to come soon, but when I saw this in the description

> This means there is no further confusion about what a tuple in a trace list means: the address is the address, so if you trace to a tuple address, the tuple enters the dictionary. This allows the collapse of several instances of code that shifted between string and 1-tuple representation addresses.

I went digging for the thread where we discussed tracing-to-tuples with @alex-lew and found it here:  https://probcomp.slack.com/archives/C0DF9A8SV/p1725035641241159

In choicemap, I currently am careful to convert `{("a", "b"): 1}` => `{"a": {"b": 1}}` for the reasons described in that thread (which we should write down). I am VERY in favor of killing the one-tuple-vs-string distinction and making that simple. Just wanted to throw this design point in. Review coming soon!

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -50,7 +50,7 @@
 DynamicAddressComponent = int | IntArray | slice
 AddressComponent = StaticAddressComponent | DynamicAddressComponent
 Address = tuple[AddressComponent, ...]
-StaticAddress = tuple[StaticAddressComponent, ...]
+StaticAddress = StaticAddressComponent | tuple[StaticAddressComponent, ...]
```
sritchie, line 53: this is a good change

src/genjax/_src/generative_functions/static.py
```diff
@@ -116,20 +94,54 @@ def get_gen_fn(self) -> GenerativeFunction[R]:
         return self.gen_fn
 
     def get_choices(self) -> ChoiceMap:
-        addresses = self.addresses.get_visited()
-        sub_chms = (tr.get_choices() for tr in self.subtraces)
-        return ChoiceMap.from_mapping(zip(addresses, sub_chms))
+        return ChoiceMap.d({
```
sritchie, line 96: ah, nice, this will do the work of flattening out the tuple addresses.

sritchie (commented):
more coming...

femtomc:
> In choicemap, I currently am careful to convert {("a", "b"): 1} => {"a": {"b": 1}} for the reasons described in that thread (which we should write down). I am VERY in favor of killing the one-tuple-vs-string distinction and making that simple. Just wanted to throw this design point in. Review coming soon!

The reason why is because one wants a way to write proposals for models like:
```python
@gen
def submodel():
    x = normal(0.0, 1.0) @ "x"

@gen
def model():
    v = submodel() @ "y"
```

where proposals look like:

```python
@gen
def proposal():
    normal(0.0, 1.0) @ ("y", "x")
```

In other words, we'd like proposals to be able to arbitrarily re-order addresses, nest them, etc (essentially, do anything they want with the address structure -- and make it easy to do it!) -- so that one doesn't have to verbatim copy the structure of the code of the model to make a proposal targeting certain variables.

Edit: that's exactly what the Slack thread says 😄

src/genjax/_src/generative_functions/static.py
```diff
@@ -255,34 +267,29 @@ def dispatch(self, primitive, *tracers, **_params):
...
     def handle_trace(
         self,
         addr: StaticAddress,
         gen_fn: GenerativeFunction[Any],
         args: tuple[Any, ...],
     ):
-        self.visit(addr)
+        if addr in self.traces:
+            raise AddressReuse(addr)
+        self.traces[addr] = VOID_TRACE
```
femtomc: I don't yet understand why this is inserted if the logic below is just going to create a trace and insert it? No one reflects on the contents of `self.traces` after you check for address reuse.

femtomc:
We'd also like the following to throw. I'm not sure if `main` actually prevents against this.

```python
@gen
def submodel():
    x = normal(0.0, 1.0) @ "y"

@gen
def model():
    q = normal(0.0, 1.0) @ ("x", "y")
    v = submodel() @ "x"
```

But this should be okay:
```python
@gen
def model():
    q = normal(0.0, 1.0) @ ("x", "y")
    v = normal(0.0, 1.0) @ ("x", "z")
```

We could defer these questions to another PR!

src/genjax/_src/generative_functions/static.py
littleredcomputer: I remember the conversation with Alex and worried about it, but if we don't have a test for the behavior I was going to say let's write one :)

src/genjax/_src/generative_functions/static.py
littleredcomputer: inserting the void trace establishes a "lease" on the dictionary key, which only "this code" is allowed to fill, which does the job of address_visitor.visit formerly

littleredcomputer:
> 
> We could defer these questions to another PR!

I will try to add them. The more the merrier. (note: main allows this)

Edit: Hrm. this is thornier than I thought: either trace could go first, and what you have to do to figure out that the traces are mutually disjoint is different.

However, I feel this PR does no harm: the trace created is addressed the same way before and after.

src/genjax/_src/generative_functions/static.py
```diff
@@ -116,20 +94,54 @@ def get_gen_fn(self) -> GenerativeFunction[R]:
...
 
+        return self.subtraces[addr]
+
+
+@Pytree.dataclass
+class VoidTrace(Trace[None]):
+    """Under normal circumstances, you should not use or receive an object
+    of this type, which represents a Trace whose content is not yet known."""
+
+    def get_retval(self):
```
sritchie, line 127: would you mind adding a quick set of test cases checking these methods?

src/genjax/_src/generative_functions/static.py
```diff
@@ -173,7 +185,7 @@ class AddressReuse(Exception):
 # stage, any traced values stored in `gen_fn`
 # get lifted to by `get_shaped_aval`.
 def _abstract_gen_fn_call(
-    _: tuple[Const[StaticAddress], ...],
+    _: Const[StaticAddressComponent] | tuple[Const[StaticAddress], ...],
```
sritchie, line 187: could this be `Const[StaticAddress]`?

src/genjax/_src/generative_functions/static.py
```diff
@@ -412,20 +403,11 @@ def wrapper(
...
         retval = forward(source_fn)(stateful_handler, *args)
-        (
-            weight,
-            address_visitor,
-            address_traces,
-        ) = stateful_handler.yield_state()
+        (weight, traces) = stateful_handler.yield_state()
         return (
             weight,
             # Trace.
```
sritchie, line 412: what's this comment saying?

src/genjax/_src/generative_functions/static.py
```diff
@@ -482,7 +459,9 @@ def handle_trace(
         args: tuple[Any, ...],
     ):
         argdiffs: Argdiffs = args
-        self.visit(addr)
+        if addr in self.traces:
```
sritchie: can we test for this error? if anything it'll help with documentation

sritchie (approved):
Lovely, small note on testing otherwise 🚢

src/genjax/_src/generative_functions/static.py
littleredcomputer: we do test for this

src/genjax/_src/generative_functions/static.py
littleredcomputer: the comment is suggestive that: the tuple that follows is of the shape that the Trace constructor is expecting as arguments

src/genjax/_src/generative_functions/static.py
littleredcomputer: I wish, but the constness appears to live inside the tuple. (maybe that could be fixed but I'm not sure where it comes from at this point)

src/genjax/_src/generative_functions/static.py
femtomc: The concept of a void trace seems unnecessary, when you’re already checking if the address is in the dictionary?

That state isn’t being written into simultaneously elsewhere — we check the keys, and throw. Otherwise, we’re good to go (I think)

Tell me if I’m missing something

littleredcomputer: You're correct. I thought the split between the update of the address visitor and the installation of the trace result was meant to guard against updating the trace *while* invoking the handler. But I've changed it to just do the check at the end and it works fine.

femtomc: Thanks! I was just having nightmares of `EmptyTrace` again 😄

femtomc (approved):

PR #1507, opened 2025-02-10
Restrict Selection to static addresses only (GEN-964)

sritchie:
With the demise of Indexed, let's kill dynamic indices in `Selection` (as these weren't handled well anyway). This fixes a bug where projecting against a selection like `S["x", "y"]` in a model with a scan or vmap would send `Selection.none()` down to the subtraces.


femtomc (approved):
Incredible 1 rep PR.

PR #1508, opened 2025-02-10
GEN-963 move get_subtrace to Trace

littleredcomputer:
- tweak address class hierarchy
- remove obsolete VoidTrace

### DIFF
--- a/src/genjax/_src/core/generative/generative_function.py
+++ b/src/genjax/_src/core/generative/generative_function.py
@@ -193,18 +193,26 @@ class Trace(Generic[R], Pytree):
             selection,
         )
 
-    def get_subtrace(self, *addresses: ExtendedAddress):
+    def get_subtrace(self, *addresses: ExtendedAddress) -> "Trace[R]":
         """
         Return the subtrace having the supplied address. Specifying multiple addresses
-        will apply the operation recursively. Beware of using the value of `get_score` on
-        internal traces obtained this way in further inference: the score of the parent
-        trace may depend on the score of the subtrace in a non-obvious way."""
+        will apply the operation recursively.
+
+        GenJAX does not guarantee the validity of any inference computations performed
+        using information from the returned subtrace. In other words, it is safe to
+        inspect the data of subtraces -- but it not safe to use that data to make decisions
+        about inference. This is true of all the methods on the subtrace, including
+        `Trace.get_args`, `Trace.get_score`, `Trace.get_retval`, etc. It is safe to look,
+        but don't use the data for non-trivial things!"""
 
         return functools.reduce(
             lambda tr, addr: tr.get_inner_trace(addr), addresses, self
         )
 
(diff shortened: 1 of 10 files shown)


sritchie (approved):

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -188,6 +193,22 @@ def project(
             projection,
         )
 
+    def get_subtrace(self, *addresses: ExtendedAddress):
+        """
+        Return the subtrace having the supplied address. Specifying multiple addresses
+        will apply the operation recursively. Beware of using the value of `get_score` on
+        internal traces obtained this way in further inference: the score of the parent
+        trace may depend on the score of the subtrace in a non-obvious way."""
```
femtomc: I'd rather a message like:

"GenJAX does not guarantee the validity of any inference computations performed using information from the returned subtrace. In other words, it is safe to inspect the data of subtraces -- but it not safe to use that data to make decisions about inference. This is true of all the methods on the subtrace, including `Trace.get_args`, `Trace.get_score`, `Trace.get_retval`, etc. It is safe to look, but don't use the data for non-trivial things!"

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -188,6 +193,22 @@ def project(
...
+
+        return functools.reduce(
+            lambda tr, addr: tr.get_inner_trace(addr), addresses, self
+        )
+
+    def get_inner_trace(self, address: ExtendedAddress):
+        raise NotImplementedError(
+            "This type of Trace object does not possess subtraces."
+        )
+
```
femtomc, line 219: Add documentation noting that this is the method that one should overload for `Trace.get_subtrace` to work correctly. Add a type annotation for the return value (presumably, a `Trace`).

tests/core/generative/test_core.py
```diff
@@ -76,6 +77,100 @@ def f():
...
+        def h():
+            flip = genjax.flip(0.5) @ "flip"
+            return f.or_else(g)(flip, (), ()) @ "z"
+
+        tr = h.simulate(jax.random.key(0), ())
+        flip_tr = tr.get_subtrace("flip")
+        flip = flip_tr.get_retval()
+        print(tr)
+        print(flip_tr)
+        print(flip)
```
femtomc: Do we want to expunge `print` out of tests?

femtomc: I'm aware they only print to the console if the test fails -- but wonder if we should just keep them out entirely in PRs.

femtomc (approved):
Few comments on documentation messages, otherwise looks good!

In the actual documentation website -- I'd like a way to tag some of these methods as "unsafe!" or "be careful!"

tests/core/generative/test_core.py
littleredcomputer: Mistake on my part will remove

PR #1510, opened 2025-02-11
remove ExactDensityFromCallables

littleredcomputer:
- instead we let `exact_density` gin up a runtime type to hold the methods
- removed ExactDensity.handle_kwargs (nobody seems to be using it, and I added kwarg capability in the new type. Maybe that's good enough?

### DIFF
--- a/src/genjax/_src/generative_functions/distributions/distribution.py
+++ b/src/genjax/_src/generative_functions/distributions/distribution.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """This module contains the `Distribution` abstract base class."""
 
+import warnings
 from abc import abstractmethod
 
 import jax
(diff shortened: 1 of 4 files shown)


src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -368,20 +369,6 @@ def logpdf(self, v: R, *args) -> Score:
     def __abstract_call__(self, *args):
         return to_shape_fn(self.sample, jnp.zeros)(_fake_key, *args)
 
-    def handle_kwargs(self) -> GenerativeFunction[R]:
```
sritchie, line 371: @femtomc wdyt? I think @littleredcomputer is right that we don't need this here since sample and logpdf can just... take the kwargs... but I haven't fixed everything up in #1467 so I need your eyes too.

femtomc: The whole “handle kwargs” business is a way for generic generative functions (not just distributions) to register how they should behave when  invoked with kwargs — this is how GenFnClosure knows what to do with kwargs for its wrapped generative function.

I’m happy with this change here!

I feel like we have some cruft with GenFnClosure … can we think clearly if we could support the `gen_fn(*args, **kwargs).simulate(key)` or `gen_fn(*args, **kwargs).assess(key, chm)` syntax?

femtomc: Of course, one idea is that we could change the interface methods to accept varargs and kwargs, and keep GenFnClosure — then you’d get syntax exactly like above.

femtomc: Fixing signatures would be a global breaking change — but possibly worth it (getting rid of argument tuples everywhere, not having a weird mechanism for handling kwargs)

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -433,26 +420,27 @@ def _check():
...
+    this new type, and should be treated as a singleton."""
+    if name is None:
+        warnings.warn("You should supply a name argument to exact_density")
+        name = "unknown"
+
+    T = type(
+        f"genjax.{name}",
+        (ExactDensity,),
+        {
+            "sample": lambda self, key, *args, **kwargs: sample(key, *args, **kwargs),
```
sritchie, line 441: can you add a test that this new kwarg is going through? You could do it with a dtype argument to `Bernoulli`, or something like that, check that you get either an int or a bool back.

sritchie (approved):
One request for a test, otherwise lgtm

PR #1511, opened 2025-02-11
remove info parameter from DimapCombinator

sritchie:
We never use this and I don't think that it buys us anything. If we want to provide good info, we should have treescope point to the function definition, yeah?

Let's reduce a touch of visual noise in the treescope output.


src/genjax/_src/core/generative/generative_function.py
```diff
@@ -1473,7 +1463,7 @@ def model(x):
         """
         import genjax
```
littleredcomputer, line 1465: Lgtm

littleredcomputer, line 1465: Lgtm

littleredcomputer (approved):

PR #1515, opened 2025-02-11
fix printed repr of genjax distributions created by lambda

littleredcomputer:
- we just add an optional name parameter for this case

### DIFF
--- a/tests/core/generative/test_core.py
+++ b/tests/core/generative/test_core.py
@@ -237,3 +237,19 @@ class TestCombinators:
 
         else_tr = jit_fn(key, (False,))
         assert "else_value" in else_tr.get_choices()("tossed")
+
+class TestRepr:
+    def test_distribution_repr(self):
+        @genjax.gen
+        def model():
+            x = genjax.normal(0.0, 1.0) @ 'x'
+            y = genjax.bernoulli(0.0) @ 'y'
+            z = genjax.flip(0.5) @ 'z'
+            t = genjax.categorical([0., 0.]) @ 't'
+            return x, y, z, t
+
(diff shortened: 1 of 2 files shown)


tests/core/generative/test_core.py
```diff
@@ -237,3 +237,20 @@ def switch_model(toss: bool):
...
+        def model():
+            x = genjax.normal(0.0, 1.0) @ "x"
+            y = genjax.bernoulli(0.0) @ "y"
+            z = genjax.flip(0.5) @ "z"
+            t = genjax.categorical([0.0, 0.0]) @ "t"
+            return x, y, z, t
+
+        tr = model.simulate(jax.random.key(0), ())
+        assert str(tr.get_subtrace("x").get_gen_fn()) == "genjax.Normal()"
+        assert str(tr.get_subtrace("y").get_gen_fn()) == "genjax.Bernoulli()"
```
sritchie, line 254: should all of these be lowercased? and without parens, since they are not called with ()?

sritchie (approved):

tests/core/generative/test_core.py
littleredcomputer: - I don't think it's me putting the parens in !? Because I agree... but don't see where they're coming from 
- I'm on the fence. For the ones that came from TF, I want them to have names people can compare, and they're all named LikeThis, so if we lowercase them we would have to add _s too, IMO. I left flip lowercase because it's not in TF. I agree that this doesn't feel perfect

sritchie: I love what we have now, I'm fine with your call here!

PR #1516, opened 2025-02-12
Take multiple address components in get_submap

sritchie:
This PR modifies get_submap to be able to take in either a single entry or multiple entries.

@littleredcomputer , unlike what you did for ChoiceMap I don't think I can provide a single tuple. wdyt? Is that okay?


littleredcomputer:
> This PR modifies get_submap to be able to take in either a single entry or multiple entries.
> 
> @littleredcomputer , unlike what you did for ChoiceMap I don't think I can provide a single tuple. wdyt? Is that okay?

I think so -- if you mean that get_submap(('a',)) won't work, I think that's fine, as I think that is deprecated in any event unless I'm missing something

littleredcomputer (approved):

PR #1518, opened 2025-02-13
Adding `IndexRequest` for `Vmap` and cookbook entry on `IndexRequest`

MathieuHuot:


src/genjax/_src/generative_functions/combinators/vmap.py
```diff
@@ -275,6 +278,66 @@ def _edit(key, idx, subtrace, argdiffs):
...
+        argdiffs_slice = Diff.tree_diff(primal_slice, Diff.tree_tangent(argdiffs))
+
+        new_trace_slice, w, retdiff, bwd_request = self.gen_fn.edit(
+            key,
+            trace_slice,
+            request,
+            argdiffs_slice,
+        )
+
+        def mutator(v, idx, setter):
```
sritchie: I would recommend inlining this down below

src/genjax/_src/generative_functions/combinators/vmap.py
```diff
@@ -275,6 +278,66 @@ def _edit(key, idx, subtrace, argdiffs):
...
+        # First get the primal. The shape of this is going to match the in_axes shape.
+        primal_slice = jax.tree_util.tree_map(
+            slice_argdiffs,
+            self.in_axes,
+            primals,
+            is_leaf=lambda x: x is None,
+        )
+        argdiffs_slice = Diff.tree_diff(primal_slice, Diff.tree_tangent(argdiffs))
+
+        new_trace_slice, w, retdiff, bwd_request = self.gen_fn.edit(
```
sritchie: if we're ignoring this retdiff let's use `_` for it instead of the name.

src/genjax/_src/generative_functions/combinators/vmap.py
```diff
@@ -275,6 +278,66 @@ def _edit(key, idx, subtrace, argdiffs):
...
+            return v.at[idx].set(setter)
+
+        new_inner_trace = jtu.tree_map(
+            lambda v, v_: mutator(v, idx, v_), trace.inner, new_trace_slice
+        )
+
+        map_tr = VmapTrace.build(self, new_inner_trace, primals, dim_length)
+
+        # We always set the carried out value to be an unknown change, conservatively.
+        retdiff = Diff.unknown_change(map_tr.get_retval())
```
sritchie, line 334: what about the retdiff that came back from the update? do we need to use it somehow, or splice it in to the original retval?

sritchie (changes requested):
Looking good! Can you add a test to the test suite using the new type of edit request? I know it's tested in the notebook but let's put something in test_vmap.

src/genjax/_src/generative_functions/combinators/vmap.py
MathieuHuot: fixed.

src/genjax/_src/generative_functions/combinators/vmap.py
MathieuHuot: good point, fixed.

src/genjax/_src/generative_functions/combinators/vmap.py
MathieuHuot: We do this actually. This line
```
new_inner_trace = jtu.tree_map(
            lambda v, v_: v.at[idx].set(v_), trace.inner, new_trace_slice
        )
```
will update the return value at the right slice.
The diff on that new return value obtained from the updated trace is then `nochange` or `unknown_change`, and we take the conservative `unknown_change` here.

MathieuHuot:
> Looking good! Can you add a test to the test suite using the new type of edit request? I know it's tested in the notebook but let's put something in test_vmap.

ok I added some tests. wdyt?

sritchie (approved):

PR #1523, opened 2025-02-17
Make "mix" tracing addresses customizable (GEN-892)

sritchie:
Still needs a docs update and tests.


MathieuHuot:
@sritchie  great idea! I think it should be customisable as you propose in this PR, and maybe the default should be to add no extra outer address (as if we were using `.inline` or as Fabian found it we can currently write `@ ()` and get the same effect).
And maybe we could have a little checker at trace time that checks whether addresses are the same in both branches and return the difference as a warning if not!

PR #1524, opened 2025-02-17
Remove docs on deprecated ^ for choicemaps (GEN-961)

sritchie:
This PR removes the docs claiming that ^ throws if the addresses clash — this is not true, and we have deprecated ^ (it now behaves like |)

### DIFF
--- a/docs/cookbook/active/choice_maps.ipynb
+++ b/docs/cookbook/active/choice_maps.ipynb
@@ -154,27 +154,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


MathieuHuot:
@sritchie I think we need some follow up PRs (happy to help review/pair on it) to write a few convenience functions that users could use to check things about addresses. E.g. to check whether you use addresses twice as Fabian mentioned: https://probcomp.slack.com/archives/C089VJQMRQB/p1739186374908339

MathieuHuot (approved):

PR #1528, opened 2025-02-19
warn on implicit use of log scale in `categorical`, `bernoulli`

littleredcomputer:
- add a wrapper which can be used to issue a deprecation warning for the use of bare parameters in those cases where we implicitly selected log scale for probability.

- Fixed docs, unit tests and examples

- fixed string repr of distributions to map camel to snake case, corresponding to GenJAX usage instead of TF


src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -420,10 +421,22 @@ def _check():
                 return w, v
 
 
+def canonicalize_distribution_name(s) -> str:
+    """Converts underlyinig distribution name from CamelCase to snake_case
```
sritchie: typo on underlyinig

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -420,10 +421,22 @@ def _check():
...
+    and prepents `genjax.`"""
```
sritchie: "prepends"

src/genjax/_src/generative_functions/distributions/distribution.py
```diff
@@ -420,10 +421,22 @@ def _check():
                 return w, v
 
 
+def canonicalize_distribution_name(s) -> str:
```
sritchie: add `s: str`?

docs/cookbook/inactive/expressivity/ravi_stack.ipynb
```diff
@@ -44,7 +44,7 @@
    "source": [
     "@gen\n",
     "def model():\n",
-    "    idx = genjax.categorical([0.5, 0.25, 0.25]) @ \"idx\"\n",
```
sritchie, line 47: so was this a case of someone thinking they were passing probs when in fact they were passing logits?

sritchie: Looks like it!!

tests/generative_functions/test_static_gen_fn.py
```diff
@@ -147,7 +147,7 @@ def test_get_zero_trace(self):
         @genjax.gen
         def model(x):
             y = genjax.normal(x, 1.0) @ "y"
-            z = genjax.bernoulli(0.7) @ "z"
```
sritchie, line 150: wow, basically in all cases this was probably a mistake.

sritchie (approved):

PR #1529, opened 2025-02-19
Remove MaskSel, Flag => bool in a number of spots (GEN-979)

sritchie:
This PR:

- removes `MaskSel` (as selections should only be built to match against str addresses)
- changes the signature of Selection.check to return `bool` (no need to return a flag if we don't check against dynamic)
- fixes some `get_subselection` signatures that weren't taking `StaticAddressComponent`
- changes `ChoiceMap.has_value` to return a `bool`, `True` if the node is a non-`None` `Choice`, `False` otherwise. (This is only "hairy" for `SwitchChm`, and I've implemented it so this returns true if any of the branches have a value. Thoughts?)
- `ChoiceMap.filter` now takes a `Selection | Flag`, and `ChoiceMap.mask` passes through to `ChoiceMap.filter`.

This last one is a little weird, and maybe the two ideas need to stay separate. `mask` is an operation for leaves, `filter` is an operation for the `ChoiceMap`. we could have two implementations and use `tree_map` to make it really clear that we are trying to get down to the leaves. This is probably the way?

### DIFF
--- a/src/genjax/_src/core/generative/choice_map.py
+++ b/src/genjax/_src/core/generative/choice_map.py
@@ -24,7 +24,6 @@ from deprecated import deprecated
 
 from genjax._src.core.generative.core import Constraint
 from genjax._src.core.generative.functional_types import Mask
-from genjax._src.core.interpreters.staging import FlagOp
 from genjax._src.core.pytree import Pytree
 from genjax._src.core.typing import (
     Any,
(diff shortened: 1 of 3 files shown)


littleredcomputer:
LGTM. I suppose there used to be a way to use this kind of thing to get a sub-score of a vector trace, but there are other ways of doing that and interestingly this change has very little test footprint

littleredcomputer (approved):

PR #1530, opened 2025-02-20
Tighten up types on get_submap, get_subtrace (GEN-980)

sritchie:
This PR:
- adds hypothesis as a dev dependency and our first hypothesis tests (thanks to @DRMacIver !)
- cleans up our `Address` types. `Address` and `ExtendedStaticAddress` were pointing to a tuple only, when they should have followed the `component | tuple[component, ...]` pattern. This rippled some clean-up through the codebase
- changes various return types to `bool` that I'd left as `Flag`
- `ChoiceMap.get_submap` is now variadic, and can take multiple `Address` inputs (so, `component | tuple[component, ...]`). These will all be flattened out on query, just like `ChoiceMap` creation flattens out tupled addresses.
- `ChoiceMap.__call__` now simply calls `ChoiceMap.get_submap`
- `Trace.get_subtrace` and `Trace.get_inner_trace` no longer allowed ellipses (accomplished by changing `ExtendedAddress` to `Address` in these methods)

### DIFF
--- a/src/genjax/_src/core/generative/choice_map.py
+++ b/src/genjax/_src/core/generative/choice_map.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import functools
 from abc import abstractmethod
 from dataclasses import dataclass
 from operator import or_
@@ -938,13 +939,12 @@ class ChoiceMap(Pytree):
     ) -> "ChoiceMap":
         pass
 
-    def get_submap(self, *addr: AddressComponent) -> "ChoiceMap":
-        addr = _validate_addr(addr, allow_partial_slice=True)
-
-        submap = self
-        for comp in addr:
-            submap = submap.get_inner_map(comp)
-        return submap
+    def get_submap(self, *addresses: Address) -> "ChoiceMap":
+        addr = sum((a if isinstance(a, tuple) else (a,) for a in addresses), ())
+        addr: tuple[AddressComponent, ...] = _validate_addr(
+            addr, allow_partial_slice=True
+        )
+        return functools.reduce(lambda chm, addr: chm.get_inner_map(addr), addr, self)
 
     def has_value(self) -> bool:
         return self.get_value() is not None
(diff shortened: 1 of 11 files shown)


src/genjax/_src/core/generative/choice_map.py
```diff
@@ -45,18 +46,23 @@
...
 ExtendedStaticAddressComponent = StaticAddressComponent | EllipsisType
-ExtendedStaticAddress = tuple[ExtendedStaticAddressComponent, ...]
 ExtendedAddressComponent = ExtendedStaticAddressComponent | DynamicAddressComponent
+
+# Addresses
+Address = AddressComponent | tuple[AddressComponent, ...]
+StaticAddress = StaticAddressComponent | tuple[StaticAddressComponent, ...]
+ExtendedStaticAddress = (
+    ExtendedStaticAddressComponent | tuple[ExtendedStaticAddressComponent, ...]
+)
```
littleredcomputer, line 61: 👍

src/genjax/_src/core/generative/choice_map.py
```diff
@@ -933,13 +938,14 @@ def get_inner_map(
...
 
-    def get_submap(self, *addr: AddressComponent) -> "ChoiceMap":
-        addr = _validate_addr(addr, allow_partial_slice=True)
-
-        submap = self
-        for comp in addr:
-            submap = submap.get_inner_map(comp)
-        return submap
+    def get_submap(self, *addresses: Address) -> "ChoiceMap":
+        addr = tuple(
```
littleredcomputer, line 942: is this like a deep flatten? I had to stare at it. you could try `tuple(itertools.chain(*address))`

littleredcomputer (approved):

src/genjax/_src/core/generative/choice_map.py
sritchie: it's just a one level flatten — that doesn't quite work since the "address" can be a bare thing or a tuple, so we'd need to tuple them all first...

I don't like reading the nested python for comprehensions!! Clojure ones feel way nicer...

```clj
(for [a     addresses
      label (if (seq? a) a [a])]
  label)
```

PR #1532, opened 2025-02-23
Expunge the concept of Constraint; just rely on choice map everywhere.

femtomc:
This removes `Constraint` from the core set of concepts, simplifying several of the types in interface signatures (this is similar to the PR which expunged `Sample`).

In the future, when special assumptions about constraints need to be communicated, I expect these to live directly in `EditRequest` or (to be introduced), `GenerateRequest` (corresponding to the "advanced" interfaces `edit` ("change an existing trace") and `generate` ("sample a new properly weighted trace")).


sritchie:
image

sritchie (approved):
Couple issues in the docs, otherwise let's do it!

littleredcomputer (approved):

PR #1536, opened 2025-02-27
use existing index to descend to subtrace in SwitchTrace

littleredcomputer:


sritchie:
Nice!

sritchie (approved):

femtomc:
@littleredcomputer Can you add documentation indicating that this will not work under `jax.jit`?

(In other words, fine if people use this outside of JAX transformation boundaries -- but will break if used within code that JAX transforms)

PR #1537, opened 2025-02-27
Fix mix side-effecting issue (GEN-986)

sritchie:
This PR moves the `@gen` decorator down to a `gen(...)` call. Before doing this, beartype 0.18.0 would somehow capture the `mixture_model` function internally in such a way that calls to `simulate` (with nested models only, not single layer models) of a totally different model would try and call `mixture_model`.

This change gets around that issue.

### DIFF
--- a/tests/generative_functions/test_mix_combinator.py
+++ b/tests/generative_functions/test_mix_combinator.py
@@ -40,11 +40,12 @@ class TestMixture:
         trace = mixture.simulate(key, (logits, (0.0,), (0.0,)))
 
         # Check structure
-        assert "mixture_component" in trace.get_choices()
-        assert ("component_sample", "y") in trace.get_choices()
+        chm = trace.get_choices()
+        assert "mixture_component" in chm
+        assert ("component_sample", "y") in chm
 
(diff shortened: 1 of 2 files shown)


tests/generative_functions/test_mix_combinator.py
```diff
@@ -0,0 +1,67 @@
...
+        @genjax.gen
+        def comp2(x):
+            return genjax.normal(x + 2.0, 0.5) @ "y"
+
+        # Create mixture model
+        mixture = genjax.mix(comp1, comp2)
+
+        # Test simulation
+        key = jax.random.key(0)
+        logits = jnp.array([0.3, 0.7])
```
littleredcomputer: those look like probabilities since they add to 1. If you use (0.0, 0.0) that would be the logit scale version of (0.5, 0.5) in prob space. this is such a problem in GenJAX I think we should use small negatives in tests like this to reinforce that fact

sritchie: I blame Claude plus my skimming mind...

PR #1538, opened 2025-02-27
Upgrade to beartype 0.20.0

sritchie:
This PR upgrades our beartype package to 0.20.0. Beartype seems to have problems with deprecation, so we had to exclude those from beartyping.


littleredcomputer (approved):
lgtm

PR #1541, opened 2025-02-28
Gen1D cookbook: intro to modelling and block-gibbs

MathieuHuot:


docs/cookbook/inactive/update/7_application_dirichlet_mixture_model.ipynb
```diff
@@ -0,0 +1,550 @@
...
+   "metadata": {},
+   "source": [
+    "# Block-Gibbs on Dirichlet Mixture Model"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We will now see some of the key ingredients in action in a simple but more realistic setting and write a dirichlet mixture model in GenJAX."
```
littleredcomputer: cap D

docs/cookbook/inactive/update/7_application_dirichlet_mixture_model.ipynb
```diff
@@ -0,0 +1,550 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "datapoints = C[\"datapoints\", \"obs\"].set(\n",
+    "    jnp.concatenate([\n",
```
littleredcomputer: I don't like the for-loop here: jax.random.uniform can take a shape argument with multiple axes. try removing N_CLUSTERS from the quotient of the shape and dropping the list comprehension

littleredcomputer: I see also that you are using the cluster index to perturb the variance of the different clusters. This could be done with something like `variances = jnp.arange(N_CLUSTERS) / N_CLUSTERS * 8 = 4`

littleredcomputer: Q: why are we not inferring N_CLUSTERS?

docs/cookbook/inactive/update/7_application_dirichlet_mixture_model.ipynb
```diff
@@ -0,0 +1,550 @@
...
+    "\n",
+    "        return all_posterior_means, all_posterior_weights, all_cluster_assignment, tr\n",
+    "\n",
+    "    else:\n",
+    "        # One Gibbs sweep consist of updating each latent variable\n",
+    "        def update(carry, _):\n",
+    "            key, tr = carry\n",
+    "            # Gibbs update on `(\"clusters\", i, \"mean\")` for each i, in parallel\n",
+    "            key, subkey = jax.random.split(key)\n",
+    "            tr = jax.jit(update_cluster_means)(subkey, tr)\n",
```
littleredcomputer: I think the jitting should be requested _outside_ the scanned function

docs/cookbook/inactive/update/7_application_dirichlet_mixture_model.ipynb
```diff
@@ -0,0 +1,550 @@
...
+    "            key, subkey = jax.random.split(key)\n",
+    "            tr = jax.jit(update_datapoint_assignment)(subkey, tr)\n",
+    "\n",
+    "            # Gibbs update on `probs`\n",
+    "            key, subkey = jax.random.split(key)\n",
+    "            tr = jax.jit(update_cluster_weights)(subkey, tr)\n",
+    "            return (key, tr), None\n",
+    "\n",
+    "        # Overall inference performs a fixed number of Gibbs sweeps\n",
+    "        (key, tr), _ = jax.lax.scan(update, (key, tr), None, length=N_ITER)\n",
```
littleredcomputer: in the event that there is no data to "scan over" (evidence: the 2nd parameter to update is unused), you might be able to use @sritchie 's `iterate` combinator. Otherwise, it is unfortunate to introduce the "naked" use of `jax.lax.scan`

docs/cookbook/inactive/update/7_application_dirichlet_mixture_model.ipynb
```diff
@@ -0,0 +1,550 @@
...
+    "    # P(x | y1, y2) = P(x | y1)\n",
+    "    # Sampling from Categorical(P(x = 1 | y ), P(x = 2 | y), ...) is the same as\n",
+    "    # sampling from Categorical(P(x = 1, y), P(x = 2, y))\n",
+    "    # as the weights need not be normalized\n",
+    "    # In addition, if the model factorizes as P(x, y1, y2) = P(x, y1)P(y1 | y2),\n",
+    "    # we can further simplify P(y1 | y2) from the categorical as it does not depend on x. More generally We only need to look at the children and parents of x (\"idx\" in our situation, which are conveniently wrapped in the generate_datapoint generative function).\n",
+    "    def compute_local_density(x, i):\n",
+    "        datapoint_mean = trace.get_choices()[\"datapoints\", \"obs\", x]\n",
+    "        chm = C[\"obs\"].set(datapoint_mean).at[\"idx\"].set(i)\n",
+    "        clusters = Cluster(trace.get_choices()[\"clusters\", \"mean\"])\n",
```
littleredcomputer: Consider dropping the `Cluster` class and just call the variable `cluster_means`, a plain array. Unless you are planning to extend the `Cluster` class in a later lesson, it is just adding boilerplate

docs/cookbook/inactive/update/7_application_dirichlet_mixture_model.ipynb
```diff
@@ -0,0 +1,550 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Function to create plot for a given index\n",
+    "def create_plot(idx):\n",
```
littleredcomputer: any chance we could use genstudio?

docs/cookbook/inactive/update/7_application_dirichlet_mixture_model.ipynb
MathieuHuot: We could infer `N_CLUSTERS`, it would be an extension to the model, and it makes inference a bit harder.
 
The usual way to do that is using doing it with a Dirichlet Process. It's a lazy infinite Dirichlet distribution. 
In JAX we can implement a truncated version with a max number of clusters.

The simplest change for us here would be to put a non-symmetric Dirichlet prior (right now it's a "symmetric" vector `alpha / n_clusters.unwrap() * jnp.ones(n_clusters.unwrap()`). 
This makes the clusters ordered, and we typically put more mass on the earlier numbers, thereby enshrining into the model a version of Occam's razor.

docs/cookbook/inactive/update/7_application_dirichlet_mixture_model.ipynb
MathieuHuot: Yes, it would be great. I did this quickly when my GenStudio install was broken (Sam fixed it).

docs/cookbook/inactive/update/7_application_dirichlet_mixture_model.ipynb
MathieuHuot: It's a bit unfortunate but this line really isn't a generative function so `iterate` would add another layer of boiler plate stuff.

littleredcomputer (approved):

PR #1548, opened 2025-03-04
Relax tensorflow-probability dependency

sritchie:
This PR:
- fixes 3.13 installation issues by pinning to sub-2.0 poetry (EDIT: moved to #1549 )
- relaxes our tensorflow-probability dependency, so that colab installation doesn't require a downgrade of tfp
- updates the dependencies in our lockfile

This upgrade exposed two bugs... one's linked in the comment below, another's filed here: https://github.com/jax-ml/jax/issues/26908


sritchie:
@femtomc , `new_main` (which we use in forward.py https://github.com/ChiSym/genjax/blob/main/src/genjax/_src/core/interpreters/forward.py#L41-L47) disappeared from JAX here:

https://github.com/jax-ml/jax/commit/c36e1f7c1ad4782060cbc8e8c596d85dfb83986f#diff-713d667c14b1dd075ec025dbd1cd8ea222e42c8b4c094ae05373d4d05b276dc5

can you help me figure out how to fix it?

femtomc:
We can actually expunge a lot of that code, and I’ve done so on my delayed sampling branch if you want to copy.

PR #1551, opened 2025-03-06
Remove EditRequest restriction for DiMap

fzaiser:
As far as I can tell, this assertion can be removed as `DiMap` only modifies the input and output, not the trace itself. But someone should double-check this.


sritchie:
@fzaiser would you mind adding a quick test? Thank you @fzaiser !

sritchie (approved):

PR #1552, opened 2025-03-07
Add mapping tutorial to the cookbook (inactive)

fzaiser:
As discussed in the team meeting last week, I'm adding this to the `inactive` folder of the cookbook for now.

The tutorial inverts the setting of the localization tutorial: the robot knows its position, but not the map (i.e. position of the walls). For simplicity, we assume that the walls are arranged on a grid.

The main inference method used in the tutorial is Gibbs sampling. So this can also be viewed as a Gibbs tutorial.

[Link to Linear](https://linear.app/chisym/project/mapping-tutorial-18891513e95a/overview)


sritchie (approved):

sritchie:
Nice, this will auto-merge when tests pass!

PR #1556, opened 2025-03-10
Do a little cleanup around InitialStylePrimitive.

femtomc:
Removed a bunch of extra "low level" JAX registration stuff which we don't even use (because our program transformations eliminate primitives for regular old JAX primitives).

This should also clean up the "pretty prints" of JAXPRs of generative functions.


src/genjax/_src/adev/core.py
```diff
@@ -159,31 +157,7 @@ def _adev_prim_call(adev_prim, *args):
 
 # TODO: this is gnarly as fuck.
 def batch_primitive(args, dims, **params):
```
femtomc, line 159: This was from an old experiment in extending ADEV -- it's not fully tested, and I'm not even sure that it works -- so removing.

sritchie (approved):

PR #1557, opened 2025-03-10
Show mapping tutorial in docs

fzaiser:
This was forgotten in #1552.


MathieuHuot (approved):
Great work on the tutorial!
I could check you took the time to explain the math, several different extensions, you also tried out different things that don't necessarily improve runtime (RIP EditRequest on that example). And the visuals are great :).

fzaiser:
Thanks for catching those typos and leftover text @MathieuHuot!

PR #1559, opened 2025-03-12
Cookbook `IndexRequest` for `ScanCombinator` models

MathieuHuot:
This adds a simple cookbook in the list of cookbooks that are built toward explaining  `edit`.
This one explains a particular case: `IndexRequest` for a `scan` model. 
This is a nice optimization feature that we leverage in the paper to do checkerboard Gibbs, here the example is more toyish.


compute
femtomc, line 1: What's going on with this submodule? Can we kill it?

MathieuHuot: woopsie. yes

PR #1562, opened 2025-03-13
[POPL] Start to do the `edit` makeover.

femtomc:
Our (myself, Mathieu -- at least, and maybe more) claim is that the system should be about creating and editing traces, with a coherent batch semantics, and concise idioms for writing SMC programs using `edit`. This harkens back to several of Colin's intuitions about "promoting array semantics" into the generative world.

What better place to start than with the README example -- what did I do?
* I added `importance_k` and `edit_k` ("derived") interfaces to `GenerativeFunction` and `Trace` -- these methods just invoke the underlying `importance` and `edit` methods using batch semantics. The `_k` postscript means _either_ (for `importance_k`: give me a `n : Nat` for batch dimension) _or_ (for `edit_k`: apply an `EditRequest` vectorized over a batch of traces).
* I added a `resample_k` method to `Trace` -- `resample_k` takes a batch trace & SMC weights and returns a batch trace, as well as an estimate of the log marginal likelihood.

This enables the concise posterior program in the new README ("HMC-within-SIR").

View the README on this branch for a walkthrough.


src/genjax/_src/core/generative/requests.py
```diff
@@ -65,6 +64,11 @@ class Regenerate(PrimitiveEditRequest):
     selection: Selection
 
 
+@Pytree.dataclass(match_args=True)
+class Update(PrimitiveEditRequest):
+    constraint: ChoiceMap
```
femtomc, line 67: I just moved this here from `generative_function.py`

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -674,6 +698,22 @@ def model():
...
+        ) -> tuple[Any, tuple[Trace[R], Weight]]:
+            nk, sk = jrand.split(key)
+            sks = jrand.split(sk, n)
+            tr, ws = vmap(
+                lambda g, k, c, a: g.importance(k, c, a),
+                in_axes=(None, 0, None, None),
+            )(self, sks, constraint, args)
+            return nk, (tr, ws)
+
+        return _importance_k
```
femtomc, line 695: This requires the user provide a `n : int` -- because we don't have a trace (with a batch shape) yet.

So `gen_fn.importance_k(n)` will return a callable (the logic of doing the vmap'd `gen_fn.importance`).

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -193,41 +245,31 @@ def project(
...
-            "This type of Trace object does not possess subtraces."
+        ks, sk = jrand.split(key)
+        batch_shape = self.batch_shape()
+        idx = tfd.Categorical(logits=ws).sample(
+            seed=sk,
+            sample_shape=batch_shape,
         )
+        tr_k = jtu.tree_map(lambda x: x[idx], self)
+        Z = logsumexp(ws)
+        return ks, (tr_k, Z)
```
femtomc: Traces with proper batch semantics are just particle collections without the weights ... (later maybe, including the weights -- depending on where Jay gets / we understand the root of `score`, etc).

In line with that view, we provide `resample_k` (for all traces).

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -150,6 +154,38 @@ def get_gen_fn(self) -> "GenerativeFunction[R]":
...
+        """Override this method to provide `Trace.get_subtrace` support
+        for those trace types that have substructure that can be addressed
+        in this way.
+
+        NOTE: `get_inner_trace` takes a full `Address` because, unlike `ChoiceMap`, if a user traces to a tupled address like ("a", "b"), then the resulting `StaticTrace` will store a sub-trace at this address, vs flattening it out.
+
+        As a result, `tr.get_inner_trace(("a", "b"))` does not equal `tr.get_inner_trace("a").get_inner_trace("b")`."""
+        raise NotImplementedError(
+            "This type of Trace object does not possess subtraces."
+        )
```
femtomc, line 181: Moved the methods around a bit, just so I could put all the "derived" (methods which are _derived_ from some compositional) methods together.

src/genjax/__init__.py
```diff
@@ -34,6 +34,7 @@
 
 from .checkify import *
 from .core import *
+from .edits import *
```
femtomc, line 37: Instead of `genjax.inference.requests` -- shortened to just `genjax.edits` (easier to remember)

src/genjax/_src/core/generative/generative_function.py
```diff
@@ -193,41 +245,31 @@ def project(
...
+        return ks, (tr_k, Z)
 
     ###################
     # Batch semantics #
     ###################
 
-    @property
     def batch_shape(self):
-        return len(self.get_score())
+        return jnp.array(self.get_score(), copy=False).shape
```
femtomc: This is the proper JAX "zero-copy" way to get the batch shape.

`len` will fail on "scalar" arrays.

src/genjax/_src/generative_functions/static.py
```diff
@@ -100,6 +100,7 @@ def get_choices(self) -> ChoiceMap:
     def get_score(self) -> Score:
         return jnp.sum(
             jnp.array([tr.get_score() for tr in self.subtraces.values()], copy=False),
+            axis=0,
         )
```
femtomc, line 104: This was an important bug -- if we're allowing `get_score` to be a computation (not just returning a stored value), it needs to be compatible with batching.

If the subtraces have been batched, they are batched along the `0` axis -- so we need to sum along that axis. Otherwise, if they haven't been batched -- again, the `0` axis is all there is (it doesn't matter).

So this works for both cases.

tests/generative_functions/test_vmap_combinator.py
```diff
@@ -194,7 +194,9 @@ def foo(loc: float, scale: float):
         with pytest.raises(
             IndexError,
         ):
-            jax.jit(foo.vmap(in_axes=0).simulate)(key, (jnp.arange(2), jnp.arange(3)))
+            jax.jit(foo.vmap(in_axes=(0, 0)).simulate)(
+                key, (jnp.arange(2), jnp.arange(3))
+            )
```
femtomc: This is broken on latest JAX due to a regression which Sam filed at the main JAX repo.

docs/assets/img/logo.png
femtomc, line 1: Absolute banger of a PPTX logo created by yours truly.

femtomc: image

MathieuHuot:
@femtomc this is great and definitely going in the direction we want!
I don't get why the `blah_k` methods return a key. For instance you never use it for `resample_k`.
I get the general idea of just composing the things without adding an extra `split` line everywhere, but we're not doing that elsewhere so idk if we should change that.

MathieuHuot:
Another point is that I don't know if we should have an explicit `n` in `importance_k`. For example the batched shape is not given as argument for `resample_k` or `edit_k`. 
What happens is the batching is done for a rank > 1 tensor?
Maybe an idea would be that the shape for `importance_k` is given by the shape of the keys given as input.
Then, one could store a general shape in the trace and use that thing to know "how much batched" the trace has been so that `resample_k` and `edit_k` know what to do.

femtomc:
@MathieuHuot Will answer both of these with a bit of context around my philosophy of design here: there's two types of methods in GenJAX for the fundamental objects (gen fns and traces) -- "fundamental compositional methods" and "derived methods".

- Fundamental compositional methods _are explicit about everything_ -- they are "the lowest level", razor sharp automation. Meaning things like: explicit key splitting, if you want batching you gotta `vmap`, etc -- `edit` for instance, users have to make and provide a specific type of `EditRequest` ... these are things that users have to deal with if they use the "lowest level" automation.

- Contrast fundamental methods with "derived methods": these methods use the fundamental compositional methods for their implementation, and add _implicit automation_ to make things concise, and remove cognitive overhead from the user. For instance, `tr.update` -- that's secretly `tr.edit`, but we make an `Update(...)` request and do `tr.edit` with that under the hood of `tr.update`. A small savings, but useful -- if you want to more concisely express your program.

Okay, with that in mind -- the questions:

>  I get the general idea of just composing the things without adding an extra split line everywhere, but we're not doing that elsewhere so idk if we should change that.

This is exactly true -- we do not do this for "fundamental compositional methods" (we enforce a strict key-split-then-pass discipline). 

For the derived methods (`importance_k`, `edit_k`, `resample_k`, etc) the point of passing the key is that we're removing a bit more cognitive overhead from the user: 
- We're explicitly tracking the leading batch shape for traces already! 
- Therefore, we already know ahead of time how many keys the user needs to split into -- so we do that automatically.

It's just about conciseness -- the user is going to have to do the key splitting stuff anyways, and ... oh, they're going to split over the batch dimension ... but we already know what that is? So why don't we take it off their plate? Instead, they just get to pipe a key through and forget about the `jax.random.split` 

(Infamously, I had an early version of GenJAX where _all the interfaces did this_, which was criticized at the internal JAX chat at Google for being "not idiomatic JAX" ... which I sort of find laughable now. Reducing LoCs for things users will write _a lot_ is not idiomatic? Concerning...)

> Another point is that I don't know if we should have an explicit n in importance_k. For example the batched shape is not given as argument for resample_k or edit_k.

It's not given as an argument because _we know what it is_ for `resample_k` and `edit_k` -- it's known by the trace (`Trace.batch_shape`). However, we don't know what the shape is for `importance_k` -- and in any importance sampling program written in GenJAX, the user has to provide this anyways. So the real question is "where does that information live" -- I address your proposal for (it lives "on the shape of the keys") further below

> What happens is the batching is done for a rank > 1 tensor?

This is a great question. I strongly believe we need a consistency rule concerning user-directed batching: _if you introduce a `vmap` to create a (batched) instance of a type, expect to use `vmap` for all interfaces on that (batched) instance of a type_. This ensures that "things are lined up" correctly for _our batch semantics_. This consistency rule allows our batch semantics _and user-directed batch semantics_ to "play nice".

Said another way: how exactly would one create a rank > 1 tensor inside of a `Trace` (like a rank > 1 tensor for the score)? Well, users can use our batch semantics interfaces (`importance_k`, `edit_k`, `resample_k`, etc) -- if they're working with these interfaces, they don't need to care about using `vmap`, we're making a decision about how it is to be used for them (see above: _implicit automation_).

If they manually use `vmap`, they're taking things into their own hands! "Sure, you can still use the `vmap(edit_k)(...)`, etc -- but now you're on the hook for ensuring that your batching properly. In other words, you made an (N, K) trace via `vmap(importance_k ...)` -- you're on the hook to ensure that you `vmap(edit_k)(...)` mapping over that trace.

> Maybe an idea would be that the shape for importance_k is given by the shape of the keys given as input. Then, one could store a general shape in the trace and use that thing to know "how much batched" the trace has been so that resample_k and edit_k know what to do.

This is not a bad idea -- but it conflicts with the "implicit splitting" automation above in a particular "code smell" way: now, `importance_k` needs a batched key array ... but `edit_k` and `resample_k` don't. So a user needs to _remember that distinction_. I dislike this on the grounds of _consistency of syntax_ -- if users are using `importance_k` and `edit_k` and `resample_k`, they should roughly expect each to have some sort of overlapping automation, and they should expect to use them in "the same sort of way".

This might feel arbitrary, but I'd argue strongly that little decisions like these are critical to "the feel" of a piece of software.

@MathieuHuot I'd be interested to think if there's a "general batch semantics" which could account for both "implicit" batching and user-directed batching _together_ ("no matter what the user does, they'll get something coherent"). A priori, it seems hard to me -- but I could have the wrong intuition!

femtomc:
Rebased this on #1567, will wait to merge this one until we make progress on that one.

MathieuHuot:
@femtomc  presumably this one would come after the PR 2/N, and therefore should be refactored to remove all key-handling from the user? 
All of these seem like high-level user-facing methods like simulate and assess, and these are being changed not to need key handling.
In that case, it would make sense for `resample_k` to accept the `n` argument for the number of copies. But maybe it should accept an arbitrary shape, not just an int.

femtomc:
@MathieuHuot 
> @femtomc  presumably this one would come after the PR 2/N, and therefore should be refactored to remove all key-handling from the user? 

Yes, all handled!

> All of these seem like high-level user-facing methods like simulate and assess, and these are being changed not to need key handling. In that case, it would make sense for `resample_k` to accept the `n` argument for the number of copies. But maybe it should accept an arbitrary shape, not just an int.

Can you describe a bit more what you're thinking for arbitrary shape (and how traces with these shapes might be produced by batching traces, using methods like `importance_k`)?

PR #1563, opened 2025-03-14
Update usage of JAX partial_eval / wrap_init to accept debug info

femtomc:
JAX claims this will improve the error messages coming from inside interpreters / metaprogramming. Excited to see it!

Edit: I had to bump the JAX versioning, and update the lockfile -- hence the large LoC change.

Edit2: I also went through and corrected a bunch of deprecation warnings from JAX (and our own `MixtureCombinator`).


femtomc:
@sritchie Do you know what this means on our workflow:
```
Dependency walk failed at mkdocs-material (*) @ git+ssh://git@github.com/probcomp/mkdocs-material-insiders.git
nox > Command poetry export --format=requirements.txt --with=dev --with=docs --extras=genstudio --extras=all --without-hashes failed with exit code 1:
nox > Session tests failed.
```

mkdocs.yml
```diff
@@ -120,15 +118,14 @@ plugins:
   - markdown-exec
   - mkdocs-jupyter:
         execute: true
-        allow_errors: false
+        allow_errors: true
```
sritchie: woah!! what's the story here? relaxing our standards?

femtomc: Several notebooks have snippets that try catch the ValueError that JAX regressed on (I want to go through the notebooks and make them "_legible_" as well, and I think that will involve expunging such snippets).

sritchie (approved):
Can we make failures not okay in the docs? what breaks in that notebook?

mkdocs.yml
femtomc: In general, "testing the notebooks" is very painful locally -- it takes forever to build and execute them through `mkdocs`. I don't find it ideal yet at all.

PR #1565, opened 2025-03-15
Add a simple contract system to the time travel debugger.

femtomc:


src/genjax/_src/core/interpreters/time_travel.py
```diff
@@ -38,36 +39,92 @@
...
+    contracts: list[ClosedContract[Any, Any]]
+    f: Callable[..., R] = Pytree.static()
+    cont: Callable[..., S] = Pytree.static()
+
+    def check(self):
+        return all(contract.check() for contract in self.contracts)
+
+
+@Pytree.dataclass
+class Breakpoint(Generic[R, S], Pytree):
```
fzaiser, line 106: Maybe add a docstring?

src/genjax/_src/core/interpreters/time_travel.py
```diff
@@ -77,21 +134,15 @@ def _cont_prim_call(brk_pt, *args):
         return initial_style_bind(record_p)(_cont_prim_call)(self, *args)
 
 
-def rec(
+def brk(
     callable: Callable[..., R],
     debug_tag: str | None = None,
 ):
```
fzaiser, line 140: Could you add a docstring? For instance, it was unclear to me what callable is used for. First I thought it was the continuation for after the breakpoint, but `Breakpoint.handle` is already given a continuation.

fzaiser (commented):
This is a cool addition! The code looks reasonable to me, just some more docstrings would be nice. Also adding a test could be good as it would also serve as documentation of the interface.

femtomc:
@fzaiser thanks for taking a look! I’ll add a bunch of docstrings and tests and ask for a quick re-review sometime this week / next

PR #1566, opened 2025-03-15
[Cold Harbor, 1 / N] Organize the lowest levels of the GenJAX compiler.

femtomc:
**No new functionality**

Performing a little re-organization in `_src.core` to make the structure of the GenJAX stack significantly more clear.

A new lower level module `genjax._src.core.compiler` -- contains infrastructure for JAX primitives, interpreters, staging, etc. Interpreters share some core functionality (`Environment`) in `genjax._src.core.compiler.interpreters.environment` and `genjax._src.core.compiler.initial_style_primitive`.


src/genjax/_src/core/compiler/interpreters/common.py
```diff
@@ -0,0 +1,71 @@
...
+VarOrLiteral = Var | Literal
+
+
+@Pytree.dataclass
+class Environment(Pytree):
+    """Keeps track of variables and their values during propagation."""
+
+    env: dict[int, Any] = Pytree.field(default_factory=dict)
+
+    def read(self, var: VarOrLiteral) -> Any:
```
sritchie: can you call `self.get(var)` inside here, then do the `None` check?

src/genjax/_src/core/compiler/interpreters/common.py
```diff
@@ -0,0 +1,71 @@
...
+            "Environments do not support __setitem__. Please use the "
+            "`write` method instead."
+        )
+
+    def __contains__(self, var: VarOrLiteral):
+        if isinstance(var, Literal):
+            return True
+        return var.count in self.env
+
+    def copy(self):
```
sritchie: let's definitely get some docstrings on these! I read it all and I get it, but I think redoing these gives us a great chance to lock down a "what the heck is this" docstring.

sritchie: get that sweet sweet cursor magic

src/genjax/_src/core/compiler/interpreters/common.py
femtomc: Yes, good point!

MathieuHuot (approved):

sritchie:
No blocker on my comments I had just forgtotten to submit

src/genjax/_src/core/compiler/interpreters/common.py
femtomc: Will fix these docstrings before merging.

PR #1605, opened 2025-05-24
Dev Container Feature

mwbrulhardt:
## Contribution
- Two devcontainer setups
    1. Base setup to be used if user does not have GPU or if the devcontainer is being used in codespaces.
    2. GPU setup that enables visibility of GPU hardware within the devcontainer 

## Other Updates
- Updated poetry from 1.8.5 to 2.1.3 (added plugin for `poetry self add poetry-plugin-export` so nox will run)
- Added pre-commit and pyright (made sure matched with pre-commit hook) to dev dependencies
- Fixed issue with mismatched version of ruff in the `.pre-commit-config.yaml` file and the `pyproject.toml` file. This was causing issues between `formatOnSave` and pre-commit hook. With this I upgraded ruff for the project such that they match.
- Made pre-commit hooks explicitly point at the pyproject file.
- Updated contributors documentation for how to use the devcontainers.


sritchie:
Great! does dynamic versioning still work? that was the reason we had stayed back on the older version of Poetry.

mwbrulhardt:
I just checked and it seems like they fixed that issue in [v1.5.0](https://github.com/mtkennerly/poetry-dynamic-versioning/releases/tag/v1.5.0).

mwbrulhardt:
@sritchie hey! Just wanted to check in and see if there was any update on this.

sritchie:
thank you!

mwbrulhardt:
No problem! However, I think I forgot to update the GitHub workflows. Do you want me to put in a quick PR to update that?

sritchie:
that would be great.

PR #1612, opened 2025-05-27
Updated GitHub workflow poetry version.

mwbrulhardt:


mwbrulhardt:
@sritchie seems like the only workflow that is having trouble is the `docs` which might be out of my control for what I can fix. There are a couple that haven't run but we might only find that out after the PR gets merged unless you're able to trigger them on this branch. I've updated the poetry version and added the export-plugin installation where I thought it might be necessary. Let me know if you want me to change anything!

sritchie:
Let's merge and see what happens. Thanks for fixing this...

PR #1613, opened 2025-05-29
Fixed vulnerability issue and updated packages.

mwbrulhardt:
Fixed the safety vulnerability issue by updating packages.

- Transitive dependencies `setuptools` and `h11` needed to be updated
- Ran an update for all packages and upgraded `safety`
- Kept `jax` version the same because when updated a new error was thrown in the `test_vmap_combinator` file and I wanted to keep that expectation unchanged.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,26 +2,38 @@
 
 [[package]]
 name = "absl-py"
-version = "2.1.0"
+version = "2.3.0"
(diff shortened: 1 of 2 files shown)


mwbrulhardt:
@sritchie So it seems like the `docs` passed yesterday when merged in. However, the safety check failed which was weird because it was passing on this PR. I reproduced it locally though, and fixed it so hopefully this is the last PR for this. lol

